PR #6, opened 2019-08-30
External Decision Agent (EDA) builtin policy

marquiz:
Implements a very simple (builtin) policy that simply looks at cpu.*
container annotations (ofthe CRI container object) in the
cri-resource-manager namespace (i.e.
cri-resource-manager.intel.com/cpu.*). If any annotations are found, the
cpusets within are combined and and the cpuset of the container is set
accordingly.

### DIFF
--- a/pkg/cri/resource-manager/cache/container.go
+++ b/pkg/cri/resource-manager/cache/container.go
@@ -19,6 +19,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/kubernetes"
 	"github.com/intel/cri-resource-manager/pkg/sysfs"
 
 	v1 "k8s.io/api/core/v1"
@@ -254,6 +255,21 @@ func (c *container) GetArgs() []string {
 	return args
 }
 
+func keysInNamespace(m *map[string]string, namespace string) []string {
+	keys := make([]string, 0, len(*m))
+
+	for key := range *m {
+		split := strings.SplitN(key, "/", 2)
+		if len(split) == 2 && split[0] == namespace {
+			keys = append(keys, split[1])
+		} else if len(split) == 1 && len(namespace) == 0 {
+			keys = append(keys, split[0])
+		}
+	}
+
+	return keys
+}
+
 func (c *container) GetLabelKeys() []string {
 	keys := make([]string, len(c.Labels))
 
@@ -268,6 +284,15 @@ func (c *container) GetLabelKeys() []string {
 
 func (c *container) GetLabel(key string) (string, bool) {
 	value, ok := c.Labels[key]
+	return value, ok
+}
+
+func (c *container) GetResmgrLabelKeys() []string {
+	return keysInNamespace(&c.Labels, kubernetes.ResmgrKeyNamespace)
+}
+
+func (c *container) GetResmgrLabel(key string) (string, bool) {
+	value, ok := c.Labels[kubernetes.ResmgrKey(key)]
 	return value, ok
 }
 
@@ -298,6 +323,14 @@ func (c *container) GetAnnotation(key string, objPtr interface{}) (string, bool)
 	}
 
 	return jsonStr, true
+}
+
+func (c *container) GetResmgrAnnotationKeys() []string {
+	return keysInNamespace(&c.Annotations, kubernetes.ResmgrKeyNamespace)
+}
+
+func (c *container) GetResmgrAnnotation(key string, objPtr interface{}) (string, bool) {
+	return c.GetAnnotation(kubernetes.ResmgrKey(key), objPtr)
 }
 
 func (c *container) GetEnvKeys() []string {
--- a/pkg/cri/resource-manager/policy/builtin/eda/eda.go
+++ b/pkg/cri/resource-manager/policy/builtin/eda/eda.go
@@ -1,0 +1,160 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


pkg/cri/resource-manager/cache/container.go
```diff
@@ -25,6 +25,11 @@ import (
 	cri "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
+const (
+	resmgrLabelNs      = "cri-resource-manager.intel.com/"
+	resmgrAnnotationNs = "cri-resource-manager.intel.com/"
```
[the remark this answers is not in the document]
klihub: I think these should be upper in our internal pkg-hierarchy. Although directly at cri/resource-manager would probably be the best, since we put the default socket paths to cri/resource-manager/sockets, maybe we should put these to cri/resource-manager/kubernetes. Something like pkg/cri/resource-manager/kubernetes/metadata.go: (sans my typos)
```

import (
...
    v1 "k8s.io/api/core/v1"
...
)

const (
...
    KeyNamespace = "cri-resource-manager.intel.com"
...
)

func ResourceManagerKey(name string) string {
    return KeyNamespace + "/" + name
}

func InResourceManagerNamespace(key string) bool {
    return strings.HasPrefix(key, KeyNamespace + "/")
}

func IsResourceManagerLabel(key string) bool {
    return InResourceManagerNamespace(key)
}

func IsResourceManagerAnnotation(key string) bool {
    return InResourceManagerNamespace(key)
}

func IsResourceManagerResource(name v1.ResourceName) bool {
    return InResourceManagerNamespace(string(name))
}

...
```

pkg/cri/resource-manager/policy/builtin/eda/eda.go
```diff
@@ -0,0 +1,172 @@
...
+func (i Implementation) CreateFn() policy.CreateFn {
+	return policy.CreateFn(i)
+}
+
+var _ policy.Implementation = Implementation(nil)
+
+func init() {
+	policy.Register(Implementation(CreateEdaPolicy))
+
+	rand.Seed(time.Now().Unix())
```
klihub: Why is this done ? Is it some golang best practice ?

pkg/cri/resource-manager/cache/container.go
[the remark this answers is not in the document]
marquiz: > `com.intel.cri-resource-manager/` ?
> not sure nowadays, which will be more readable/best practice.

I guess it's a matter of taste/personal preference. In my experience what I used is more common in k8s, and, I personally find it more readable nowadays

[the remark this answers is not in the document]
marquiz: > I think these should be upper in our internal pkg-hierarchy. Although directly at cri/resource-manager would probably be the best, since we put the default socket paths to cri/resource-manager/sockets, maybe we should put these to cri/resource-manager/kubernetes. Something like this: (sans my typos)
> 
> ```
> // pkg/cri/resource-manager/kubernetes/metadata.go:
> 
> import (
> ...
>     v1 "k8s.io/api/core/v1"
> ...
> )
> 
> const (
> ...
>     KeyNamespace = "cri-resource-manager.intel.com"
> ...
> )
> 
> func ResourceManagerKey(name string) string {
>     return KeyNamespace + "/" + name
> }
> 
> func InResourceManagerNamespace(key string) bool {
>     return strings.HasPrefix(key, KeyNamespace + "/")
> }
> 
> func IsResourceManagerLabel(key string) bool {
>     return InResourceManagerNamespace(key)
> }
> 
> func IsResourceManagerAnnotation(key string) bool {
>     return InResourceManagerNamespace(key)
> }
> 
> func IsResourceManagerResource(name v1.ResourceName) bool {
>     return InResourceManagerNamespace(string(name))
> }
> 
> ...
> ```

Makes sense. My thinking for originally keeping this functionality inside cache/container (it should've been under `cache/cache.go`, but anyway :smile:) was to keep it private until some other uses  would appear (and then refactor it out). I changed this now, adding `pkg/cri/resource-manager/kubernetes/kubernetes.go`. I only added the functionality that is needed as I didn't want to add any dead code.

I used **Resmgr** instead of **ResourceManager** in the code as it is shorter, needed less renaming and resmgr is already used in the name of the executable, for example. But, I don't really have strong feelings about the naming :wink:

pkg/cri/resource-manager/cache/container.go
```diff
@@ -253,35 +254,56 @@ func (c *container) GetArgs() []string {
 	return args
 }
 
-func (c *container) GetLabelKeys() []string {
+func (c *container) getLabelKeys(prefix string) []string {
 	keys := make([]string, len(c.Labels))
```
klihub: Sorry if this is getting borderline insane nitpicking, but in principle this code would consider
cri-resource-manager.intel.computer/label as belonging to the cri-resource-manager.intel.com namespace, which I'm guessing is not intentional. This could be easily fixed by adding something like this to the beginning of getLabelKeys():

```
func getLabelKeys(prefix string) []string {
...
    if prefix != "" && !strings.HasSuffix(prefix, "/") {
        prefix += "/"
    }
...
```
Ditto for annotations/getAnnotationKeys().

marquiz: Ha, good that somebody is reading the code :smile: 

`getLabelKeys()` (and `getAnnotationKeys()`) was actually intentionally written this way so that it could be used to filter by actually prefix (instead of namespace), e.g. list all `example.com/cpu*` labels. Makes sense?

Nevertheless, there's a problem with the slashes in this version of the patch: `GetResmgrLabelKeys()` and `GetResmgrAnnotationKeys()` did not append slash to the namespace when composing the prefix. This is now fixed in the latest version of the patchset

klihub: Ok. Being able to filter based on prefix instead of just namespace does make sense.

marquiz: I just pushed yet another version. After some "proofing time" I started to think something like what you originally suggested was good :smile: At the time being, we want to get keys in a namespace, stripping the namespace part from the results. If we need flexible prefix filtering, that can be easily added later.

Also, the newest version adds some debug printing in the eda policy.

marquiz:
New version. Added one unit test. Plus fixed a bug (:grimacing:) by this test

lmdaly:
I haven't looked into the code for this policy, but have tested it out with an EDA and it is correctly picking up the container annotations and applying the cgoup updates

So +1 for this functionality

marquiz:
Fixed unit tests

PR #12, opened 2019-09-11
topology-aware-policy: add initial/prototype version.

klihub:
Add a new policy that splits the system up into a number of
    hierarchically organised, partially overlapping pools (a tree),
    based on hardware topology: full system / individual sockets /
    NUMA nodes, CPUs/cores.

    Containers are allocated resources from the best fitting pool.
    The 'best fitting' pool is selected by calculating a score for
    each available pool, taking into account container resource
    requests, devices (including ones deduced from mounts), etc.
    The scoring algorithm is not quite what it should be, it only
    takes CPU into account ATM, is probably incomplete at best and
    outright broken at worst. Work in progress...

    The other preparing patches in this series add functionality required
    by the topology-aware policy to sysfs and the cache, and fix a few
    minor related bugs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ policy configuration adhere the following scheme:
 The data format of the policy configuration is policy-specific and may be
 different between policies (the `static` and `stp` policies use YAML).
 There is a
-[sammple ConfigMap spec](sample-configs/cri-resmgr-configmap.example.yaml)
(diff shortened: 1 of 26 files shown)


klihub:
> Generally ok for first iteration. What is really missing is updated part of readme that explains annotations/config parameters

I added a brief initial section about configuration.

PR #14, opened 2019-09-14
Implement RDT support

marquiz:
This patch implements RDT support. The resource manager now has a common
confugration of a set of RDT classes (or resctrl CTRL groups) which the
policies are able to assign containers to. This RDT configuration is
dynamically updatable via the resource manager node agent
(cri-resmgr-agent). Resource manager has a built-in default
configuration containing three classes corresponding to the Pod QOS
classes of Kubernetes. The set of classes can be freely specified in the
configuration, but, one must ensure that classes required by the active
policies have been defined, and, that the maxmimum number of classes
(CLOSes) supported by the hardware is not exceeded.

This patch integrates RDT support to the static policy. The policy
simply assigns each container to an RDT class corresponding its Pod QOS
class (i.e.  'Burstable', 'BestEffort' or 'Guaranteed').

RDT support can be disabled by using the '-no-rdt' command line option.

TODO:
- [ ] better unit test coverage
- [ ] support CDP (codee and data priorization) schemas (i.e. separate
  L3data and L3code)
- [ ] policy-level tri-state (on/off/auto) knob for rdt
- [ ] change rdt class configuration to allow soft (preferred) requirements for rdt features (i.e. e.g. enforce MB if supported by system)
- [ ] per-node overrides for rdt config

### DIFF
--- a/pkg/cri/resource-manager/config/config.go
+++ b/pkg/cri/resource-manager/config/config.go
@@ -50,3 +50,13 @@ func (c *RawConfig) HasIdenticalData(data map[string]string) bool {
 
 	return true
 }
+
+// Get returns one config section
+func (c *RawConfig) Get(key string) string {
+	if c.Data != nil {
+		if value, ok := c.Data[key]; ok {
+			return value
+		}
+	}
+	return defaultConfig(key)
+}
--- a/pkg/cri/resource-manager/config/defaults.go
+++ b/pkg/cri/resource-manager/config/defaults.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2019 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+var defaultRdtRawConf = `{
+  "resctrlGroups": {
+    "Guaranteed": {
+	  "l3schema": {
+        "all": "100%"
+      }
+    },
+    "Burstable": {
+	  "l3schema": {
+        "all": "100%"
+      }
+    },
+    "BestEffort": {
+	  "l3schema": {
+        "all": "100%"
+      }
+    }
+  }
+}`
+
+func defaultConfig(key string) string {
+	switch {
+	case key == "rdt":
+		return defaultRdtRawConf
+	}
+	return ""
+}
--- a/pkg/cri/resource-manager/policy/builtin/eda/eda.go
+++ b/pkg/cri/resource-manager/policy/builtin/eda/eda.go
@@ -48,7 +48,7 @@ var _ policy.Backend = &eda{}
 //
 
 // CreateEdaPolicy creates a new policy instance.
-func CreateEdaPolicy(opts *policy.Options) policy.Backend {
+func CreateEdaPolicy(opts *policy.BackendOptions) policy.Backend {
 	eda := &eda{Logger: logger.NewLogger(PolicyName)}
 	eda.Info("creating policy...")
 	// TODO: policy configuration (if any)
@@ -149,7 +149,7 @@ func edaError(format string, args ...interface{}) error {
 //
 
 // Implementation is the implementation we register with the policy module.
-type Implementation func(*policy.Options) policy.Backend
+type Implementation func(*policy.BackendOptions) policy.Backend
 
 // Name returns the name of this policy implementation.
 func (i Implementation) Name() string {
--- a/pkg/cri/resource-manager/resource-control/control.go
+++ b/pkg/cri/resource-manager/resource-control/control.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2019 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resourcecontrol
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	"github.com/intel/cri-resource-manager/pkg/rdt"
+	"github.com/intel/cri-resource-manager/pkg/utils"
+)
+
+// CriRdt is the RDT control interface for resource-manager
+type CriRdt interface {
+	rdt.Control
+
+	SetContainerClass(cache.Container, string) error
+}
+
+type criRdt struct {
+	rdt.Control
+}
+
+// NewCriRdt creates a new CriRdt instance
+func NewCriRdt(resctrlpath string, config string) (CriRdt, error) {
+	var err error
+	c := &criRdt{}
+	c.Control, err = rdt.NewControl(resctrlpath, config)
+	return c, err
+
+}
+
+// SetContainerClass assigns all processes of a container into am RDT class
+func (c *criRdt) SetContainerClass(container cache.Container, class string) error {
+	cID := container.GetID()
+	pod, ok := container.GetPod()
+	if !ok {
+		return controlError("Pod of container %q not found", cID)
+	}
+	cgroupParent := pod.GetCgroupParentDir()
+
+	pids, err := utils.GetProcessInContainer(cgroupParent, cID)
+	if err != nil {
+		return controlError("failed to get PIDs of container %s: %v", cID, err)
+	}
+
+	for _, pid := range pids {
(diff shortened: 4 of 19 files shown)


pkg/rdt/rdt.go
```diff
@@ -0,0 +1,248 @@
...
+	i := 0
+	for k, _ := range r.conf.ResctrlGroups {
+		ret[i] = k
+		i++
+	}
+	sort.Strings(ret)
+	return ret
+}
+
+func (r *rdtControl) SetClassContainer(class string, container cache.Container) error {
```
klihub: Again a bit nitpicking, but my gut feeling is that it would be better to keep the pkg/rdt module generic/resmgr-agnostic. AFAICT, this is the only function which pulls in a dependency on `cache` which would be nice to avoid. It can be easily done by moving this function out of here altogether only keeping the single-pid version (which I would rename the other way around to the IMO more intuitive/natural `SetProcessClass`) , or IMO preferably by leaving here a `SetCgroupClass` which reads all pids from the given cgroups task file an applies the class to them.

pkg/rdt/rdt.go
```diff
@@ -0,0 +1,248 @@
...
+
+	for _, pid := range pids {
+		if err = r.SetClassPID(class, strconv.Itoa(pid)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *rdtControl) SetClassPID(class, pid string) error {
```
klihub: nitpick: IMO it is more natural to say that the `RDT class` is a property of the `process` and not the other way around. To back this up/or to look at it in another way: setting the class of a process does not/shpuld not change the class of other processes, but `SetClassPID`, at least to me, seems to suggest a one-to-one relationship between the two, thus soft-implying that setting the class pid would undo a previously set one.

Therefore, I'd rename this to `SetPIDClass` or preferably to the more descriptive `SetProcessClass`.

pkg/rdt/config.go
```diff
@@ -0,0 +1,78 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rdt
+
+import (
+	"fmt"
+)
+
```
klihub: I think a brief/concise comment section about the cache and memory bandwidth control in RDT vs. the configuration data here, and perhaps a pointer to more detailed documentation about RDT or resctrl would be warranted here.

pkg/rdt/rdt.go
kad: I agree, better to keep rdt generic. not sure even if cgroup name will be good idea.
IMO, rdt should be able to set/get list of pids per class. And code outside should verify that container list of pids correctly intersects with rdt list.

kad:
two comments:
1. with bitmask operations, I think some bits from `math/bits` can be used, instead of re-implementing from scratch.
2. we need to be sure that different RDT parts of schemata can be configured separately. not all CPUs have all the features, so something might be not present or not enabled by sysadmin.

pkg/rdt/rdt.go
klihub: Then the easiest would be tp just change SetClassPID to
```
func (r *rdtControl) SetProcessClass(class string, pids ...int) error {
   for _, pid := range pids {
...
```

kad: I wonder, can we utilize some methods from libcontainer? this should be more appropriate than trying to do it all on our side.

marquiz:
Pushed a new version
- rebased on top of the latest master
- one new commit addressing the review comments of the first version

> 1. with bitmask operations, I think some bits from `math/bits` can be used, instead of re-implementing from scratch.

Thanks for pointing this out! Now changed/fixed

> 2. we need to be sure that different RDT parts of schemata can be configured separately. not all CPUs have all the features, so something might be not present or not enabled by sysadmin.

This is already possible. Added TODO item for soft requirements

pkg/rdt/rdt.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (r *rdtControl) resctrlGroupPath(name string) string {
+	return filepath.Join(r.info.resctrlPath, resctrlGroupPrefix+name)
+}
+
+func (r *rdtControl) getResctrlGroups() ([]string, error) {
+
+	files, err := ioutil.ReadDir(r.info.resctrlPath)
+	groups := make([]string, 0, len(files))
+	if err != nil {
```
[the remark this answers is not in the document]
marquiz: Hmm, no reason. Either a accidental leftover from some refactoring, or, I was just ignorant. Fixed in the new version of the PR

pkg/cri/resource-manager/config/config.go
```diff
@@ -50,3 +50,12 @@ func (c *RawConfig) HasIdenticalData(data map[string]string) bool {
 
 	return true
 }
+
+func (c *RawConfig) Get(key string) string {
+	if c.Data != nil {
+		if _, ok := c.Data[key]; ok {
+			return c.Data[key]
```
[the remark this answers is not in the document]
marquiz: Yup, looks really stupid. Now fixed...

pkg/rdt/rdt.go
marquiz: Dropped dependency on `cache` in the new version of the PR. Renamed `SetClassPID()` to `SetProcessClass` as @klihub suggested (and added support for multiple PIDs).

Haven't taken a look at libcontainer yet

pkg/rdt/rdt.go
marquiz: Agree, the naming was bad. Changed as you suggested

pkg/rdt/config.go
marquiz: Agree. This isn't yet addressed. Belongs under `documentation` in my todo list :smile:

pkg/cri/resource-manager/config/defaults.go
```diff
@@ -0,0 +1,45 @@
...
+var defaultRdtRawConf = `{
+  "resctrlGroups": {
+    "Guaranteed": {
+	  "l3schema": {
+        "default": "100%"
+      }
+    },
+    "Burstable": {
+	  "l3schema": {
+        "default": "66%"
```
klihub: All default classes should default to 100%

marquiz:
Removed RFC/WIP status. I would be ready to merge this

pkg/cri/resource-manager/resource-control/control.go
```diff
@@ -0,0 +1,72 @@
...
+func NewCriRdt(resctrlpath string, config string) (CriRdt, error) {
+	var err error
+	c := &criRdt{}
+	c.Control, err = rdt.NewControl(resctrlpath, config)
+	return c, err
+
+}
+
+// SetContainerClass assigns all processes of a container into am RDT class
+func (c *criRdt) SetContainerClass(class string, container cache.Container) error {
```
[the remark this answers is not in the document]
marquiz: Makes sense, changed

pkg/cri/resource-manager/resource-control/control.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	cgroupParent := pod.GetCgroupParentDir()
+
+	pids, err := utils.GetProcessInContainer(cgroupParent, cID)
+	if err != nil {
+		return controlError("failed to get PIDs of container %s: %v", cID, err)
+	}
+
+	for _, pid := range pids {
+		if err = c.SetProcessClass(class, strconv.Itoa(pid)); err != nil {
```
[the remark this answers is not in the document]
marquiz: This doesn't work as (based on previous review comments) I just changed `SetProcessClass()` to accept multiple PIDs

pkg/cri/resource-manager/policy/builtin/eda/eda.go
```diff
@@ -48,7 +48,7 @@ var _ policy.Backend = &eda{}
 //
 
 // CreateEdaPolicy creates a new policy instance.
-func CreateEdaPolicy(opts *policy.Options) policy.Backend {
+func CreateEdaPolicy(opts *policy.BackendOpts) policy.Backend {
```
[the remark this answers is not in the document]
marquiz: Dunno, lint was happy with this naming :smile: 

Changed this, anyway

marquiz:
> 1. Related to integration with policies... Currently the RDT class assigned to a container is not stored anywhere, neither in the container nor among the private data of the static policy. That's fine for now. However, since we do intend to offer RDT-control as a stock policy mechanism available to all policies wishing to exercise RDT control over containers, I think it would be only natural to also extend cache.Container with functions for getting/setting the RDT class. Storing it would then happen automatically.

Agree that this should probably be reworked. We don't necessarily need to do it now as it's sort of internal implementation detail. I opened #34 that should cover this.

> 2. I'm not sure if it is now possible for the user to split up the cache to classes without gaps or overlap using the percentage-notation. It should be possible for the user to easily specify such configurations. For instance to say something like
> 
> * split up the cache to 3 classes in 2 groups
>   
>   1. `BestEffort` should be `50%` and `Burstable` `100%` of the lower `50%` of cache lines
>   2. Guaranteed should be `100%` of the cache lines.
>      This should end up with no gaps, and the only overlap between Guaranteed and the rest.

No, that's not currently possible. I think this can be a further enhancement as well. Created #35 for tracking that

kad:
We need at least some minimal and reasonable configuration examples on how to configure RDT.

kad:
And please squash the commits once PR will be in the shape for merging.

marquiz:
> And please squash the commits once PR will be in the shape for merging.

Of course. I'll force push a squashed patch when we're happy with the content

marquiz:
> We need at least some minimal and reasonable configuration examples on how to configure RDT.

Added minimal documentation plus a sample config

marquiz:
Take a look at the latest version:
- altered configuration file format
- do not leave mb or l3 in undefined state: configure to default value (100%) if they are enabled in the system but no configuration is specified by the user
- made rdtInfo a global variable in pkg/rdt scope
- some typo fixes

marquiz:
Squashed and rebased

klihub:
LGTM

PR #17, opened 2019-09-16
topology-aware: expose allocated shared, exclusive and isolated CPUs to contianers

klihub:
Add code to expose the allocated CPUs to containers similarly to the `static-plus` policy.

Container resource allocation (currently CPU only) is exposed using a container-specific directory wich is bind-mounted to `/.cri-resmgr` inside the container. It is populated with a single entry, `resource.sh`, which is a shell-source-able file containing information about the various kinds of CPUs allocated to the container. The contents of this file can be used by `isolation-aware` and `multi-CPU exclusive` containers to further pin their corresponding `threads` to the exclusively allocated (isolated or other) CPUs using `sched_setaffinity`. This file can also be monitored with `inotify` to get notifications about possible changes in CPU assignments.

The other patch in this PR fixes a small glitch in the handling of fake hints.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/topology-aware-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/topology-aware-policy.go
@@ -164,7 +164,27 @@ func (p *policy) UpdateResources(c cache.Container) error {
 
 // ExportResourceData provides resource data to export for the container.
 func (p *policy) ExportResourceData(c cache.Container, syntax policyapi.DataSyntax) []byte {
-	return nil
+	grant, ok := p.allocations.CPU[c.GetCacheID()]
+	if !ok {
+		return []byte{}
+	}
+
+	shared := grant.SharedCPUs().String()
+	isolated := grant.ExclusiveCPUs().Intersection(grant.GetNode().GetCPU().IsolatedCPUs())
+	exclusive := grant.ExclusiveCPUs().Difference(isolated).String()
+
+	data := ""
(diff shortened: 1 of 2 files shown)


pkg/cri/resource-manager/policy/builtin/topology-aware/topology-aware-policy.go
```diff
@@ -164,7 +164,27 @@ func (p *policy) UpdateResources(c cache.Container) error {
...
+	shared := grant.SharedCPUs().String()
+	isolated := grant.ExclusiveCPUs().Intersection(grant.GetNode().GetCPU().IsolatedCPUs())
+	exclusive := grant.ExclusiveCPUs().Difference(isolated).String()
+
+	data := ""
+	if shared != "" {
+		data += "SHARED_CPUS=\"" + shared + "\"\n"
+	}
+	if isolated.String() != "" {
+		data += "ISOLATED_CPUS=\"" + isolated.String() + "\"\n"
```
kad: why do string conversion here, and not above like with others?

klihub: We need isolated in CPUSet form so we can .Intersect() it to calculate 'exclusive'. The other fully regular way would have been to not .String() any of them, then check for emptiness using (cpuset.CPUSet).IsEmtpy() and finally .String()ify them during string concatenation. Dunno which is better... I can change it to be like that if it is considered preferable/better.

Issue #34, opened 2019-10-09
RDT: "standardized" mechanism for out-of-band resource control enforcement


marquiz:
Inspired by https://github.com/intel/cri-resource-manager/pull/14#pullrequestreview-299308092

There would need to be a standard way of adding out-of-band (i.e. outside of CRI) resource control enforcement points, such as RDT or blockio (or other) cgroup controls. Some design goals:
- adding new resource control types should be easy
- using the resource controls from policies should be simple, avoiding duplicate/boilerplate code

One possible scheme would be utilizing `cache.Container` (and/or `Pod`) for storing the desired state, and, `resource-manager` level callbacks on enforcing the desired state(s). E.g. in the case of RDT there would be `Container.SetRdtClass()` for setting/storing the desired RDT class and a separate callback in `resource-manager` (registered by the RDT handler)  would do the enforcement.

klihub:
The recently merged PR #82 attempts to generalize resource control, among other things by
 - more explicitly separating decision making from enforcement, and
 - introducing resource controllers, entities responsible for enforcing decisions in a particular domain

The PR implements 3 controllers:
 - cri: control of resources covered by the CRI protocol/API
 - rdt: control over cache and memory BW usage, using the resctrl fs
 - blockio: (a stub for eventually) policying block I/O using the blkio cgroupfs controller

Policies use decision functions provided by cache.Container to alter/set the desired allocation/state for any of these controlled resources. The associated controller takes care of enforcing any pending decisions.

PR #40, opened 2019-10-16
static-plus policy: support RDT

marquiz:
Integrate RDT support to the static-plus policy. Implementation
replicates static policy, that is, the container is assigned to an RDT
class that corresponds its Pod QOS class ('Burstable', 'BestEffort' or
'Guaranteed').


marquiz:
> LGTM. But eventually I'd prefer the policies to do something like this _as part of the normal decision making process_:
> 
> ```
> func (p *policy) AllocateResources(c cache.Container) error {
> ...
>     rdtClass := p.QOSToRDTClass(pod.GetQOSClass()
>     if err := container.SetRDTClass(rdtClass); err != nil {
>         if p.strictRdt {
>             return policyError("failed set RDT class to %s: %v", rdtClass, err)
>         } else {
>             p.Warning("failed set RDT class to %s: %v", rdtClass, err)
>         }
>     }
> ...
>     return nil
> }
> ```
> 
> And then hook in the actual control part from either pkg/rdt itself or from pkg/cri/resource-manager.

Yes, this is tracked in a separate issue #34

PR #44, opened 2019-10-17
pkg/log: support for multi-line logging, don't calculate message prefix all the time

klihub:
This patch
- adds support for multi-line logging
- precalculates and caches message prefix string, recalculates it only when necessary

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -45,6 +45,11 @@ type Logger interface {
 
 	DebugEnabled() bool
 	Debug(format string, args ...interface{})
+	Block(fn func(string, ...interface{}), prefix string, format string, args ...interface{})
+	DebugBlock(prefix string, format string, args ...interface{})
+	InfoBlock(prefix string, format string, args ...interface{})
+	WarnBlock(prefix string, format string, args ...interface{})
+	ErrorBlock(prefix string, format string, args ...interface{})
 
 	Stop()
 }
@@ -67,6 +72,7 @@ type logger struct {
 	enabled bool   // logger source module
 	level   Level  // first non-suppressed severity level
 	debug   bool   // debugging for this instance
+	prefix  string // message prefix
 }
 
 // Get an existing logger or create a new one.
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -294,6 +370,11 @@ func (o *options) updateLoggers() {
 	}
 }
 
+// loggerError returns a formatted logger-specific error.
+func loggerError(format string, args ...interface{}) error {
+	return fmt.Errorf("log: "+format, args...)
+}
```
marquiz: Is this used somewhere?

pkg/log/log.go
```diff
@@ -236,6 +275,37 @@ func (l *logger) Debug(format string, args ...interface{}) {
 	opt.active.Debug(l.formatMessage(format, args...))
 }
 
+// Block emits a block of messages with using the given emitting function.
+func (l *logger) Block(fn func(string, ...interface{}), prefix string, format string, args ...interface{}) {
+	for _, line := range strings.Split(fmt.Sprintf(format, args...), "\n") {
+		fn("%s%s", prefix, line)
```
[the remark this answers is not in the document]
klihub: Yes, it's faster and since we generate prefix, we can trust it to not have formatting directives.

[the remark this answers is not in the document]
klihub: No sorry, actually not... we can't trust prefix, since it comes from the user.

pkg/log/log.go
klihub: No. Removed it.

pkg/log/log.go
```diff
@@ -285,6 +355,12 @@ func registeredBackendNames() string {
 	return names
 }
 
+// ListBackends lists the registered logger backends.
+func ListBackends(w io.Writer) {
+	fmt.Fprintf(w, "available logger backends: %s\n",
+		registeredBackendNames())
+}
```
marquiz: Uh, missed this one in the previous review. Is there (planned) use for this function?

klihub: Yes. It could be used form main() to list available loggers if, for instance, an appropriate command line option is specified. We can't register an option for it and do the listing here, as it must be done after main has entered, to make sure all init()-time registrations are done.

We just don't bother doing any of this now as we only have a single fmt logger.

marquiz: Really a minor thing, but I'd prolly leave this dead (unused) code out for now, until it's used. wdyt?

klihub: Sure, I can remove it.

PR #45, opened 2019-10-17
RFC: Implement support for more generic, unified configuration.

klihub:
pkg/config: support for more generic configurability.
    
    This patch set 
        1) adds infrastructure for more generic, unified configurability
        2) switches most of our configuration over to using it
        3) removes the old, now obsolete policy-only SetConfig() interface
    
    More specifically:
    
    Add a configuration package which provides a unified, more
    generic mechanism for receiving configuration data from a
    number of sources. Like our previous configuration handling
    code this is also built on top of the standard flag package.
    Therefore the switch to using it should be straightforward,
    at least in principle.
    
    The configuration abstraction splits runtime configuration
    data into a set of modules. Each module has its own set of
    variables, namespaced by the name of the module. A module
    could be any self-contained logical unit of functinality,
    typically a package. Modules declare variables (configuration
    data) in an identical manner to the flag package.
    
    In addition to the basic (flag package) functionality the
    config package supports
    
    - taking configuration from multiple sources, currently
      o command line
      o YAML or JSON file
      o an(y) external source
    - multiple configurations
    - configuration change check/notifications
      o both per configuration and per module
      o with support for
        - rejecting configuration changes
        - automatic rollback of rejected configuration


klihub:
Hmm... to be honest, if I take now an objective look at this, I can't help the feeling that this is done the wrong way around: we essentially decompose our configuration data to command-line options/basic data types, then reconstruct the data from those. Right now this feels pretty strange and maybe was simply caused by some tunnel-vision effect triggered by the existing bits of code we had in place.

I think this should be done the other way around. Our overall configuration should be composed of the registered module-specific data/types. Especially considering that our primary source of configuration is not the command line but YAML from either a configuration file or from an external source (network/ConfigMap). So that'd feel a much more natural abstraction. Command-line configurability, if that's really needed at all with such level of detail/generality, should then be built on top of it. For instance using reflect to decompose the registered configuration data/types and generate flag.Flag's as needed...

What do you guys think ?

PR #49, opened 2019-10-21
resource-manager/cache: fix Pod QoS class detection

marquiz:
The naming of the cgroup directory varies depending on the cgroup driver
used by kubelet. Thus, the QoS class detection based on prefix matching
was not reliable. Change the detection to be more flexible, based on
substring matching.

### DIFF
--- a/pkg/cri/resource-manager/cache/utils.go
+++ b/pkg/cri/resource-manager/cache/utils.go
@@ -37,10 +37,6 @@ const (
 	// memory limit to OOM score adjustement
 	guaranteedOOMScoreAdj int = -998
 	besteffortOOMScoreAdj int = 1000
-
-	// cgroup parent to Pod QoS class mapping prefixes
-	cgroupBestEffortPrefix = "/kubepods.slice/kubepods-besteffort.slice/"
-	cgroupBurstablePrefix  = "/kubepods.slice/kubepods-burstable.slice/"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/cache/utils.go
```diff
@@ -198,15 +194,21 @@ func getMemoryCapacity() int64 {
...
+	case len(split) < 2:
 		qos = corev1.PodQOSClass("")
-	case strings.HasPrefix(dir, cgroupBurstablePrefix):
+	case strings.Index(split[1], "burstable") != -1:
 		qos = corev1.PodQOSBurstable
-	case strings.HasPrefix(dir, cgroupBestEffortPrefix):
+	case strings.Index(split[1], "besteffort") != -1:
 		qos = corev1.PodQOSBestEffort
-	default:
+	case strings.Index(split[1], "guaranteed") != -1:
```
kad: this wouldn't work properly. subdirectories are only created to burstable/besteffort. Guaranteed are in the root of kubepods hierarchy.
Also, constants are here: k8s.io/api/core/v1/types.go: "PodQOSBurstable" and so on.

marquiz: Ach, it was my last minute modifications that broke this :/ Thanks for spotting! Now fixed

PR #54, opened 2019-10-24
resource-manager/config: make sure we don't process configuration update requests in parallel.

klihub:
This PR is stacked on (includes) PR #53 

Since gRPC servers tend to spawn a separate goroutine for each request, make as sure as humanly possible (without adding sequence numbers to the requests) that we process configuration updates sequentially by locking the server for the (initial) processing of every request.

### DIFF
--- a/pkg/agent/kubernetes.go
+++ b/pkg/agent/kubernetes.go
@@ -20,17 +20,20 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"time"
 
 	core_v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/watch"
+	k8swatch "k8s.io/apimachinery/pkg/watch"
 	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
 	agent_v1 "github.com/intel/cri-resource-manager/pkg/agent/api/v1"
 )
+
+type namespace string
 
 // nodeName contains the name of the k8s we're running on
 var nodeName string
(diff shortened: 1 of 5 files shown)


pkg/agent/watcher.go
```diff
@@ -104,86 +95,155 @@ func (w *watcher) ConfigChan() <-chan resmgrConfig {
...
+
+	if node, err := nodew.Query(); err != nil {
+		w.Warn("failed to query node %s: %v", nodeName, err)
+	} else {
+		group = node.(*core_v1.Node).Labels[opts.labelName]
+		w.Info("configuration group is set to '%s'", group)
+	}
+
+	cfgw := newConfigMapWatch(w, opts.configMapName+".node."+nodeName, namespace(opts.configNs))
+	grpw := newConfigMapWatch(w, groupMapName(group), namespace(opts.configNs))
```
marquiz: Hmm, shouldn't we add a watch for the default map as well? If I read this correctly the implemented logic is not quite as the commit message suggests (which I think makes sense).

I think that in the case that the node has group label set but the group map does not exists, we should still return the default map(?) Or what do you think?

pkg/agent/kubernetes.go
```diff
@@ -94,17 +97,166 @@ func patchNodeStatus(cli *k8sclient.Clientset, fields map[string]string) error {
 	return err
 }
 
-// watchConfigMap watches changes in a ConfigMap object
-func watchConfigMap(cli *k8sclient.Clientset, ns string, name string) (watch.Interface, error) {
-	listOpts := meta_v1.ListOptions{
-		FieldSelector: "metadata.name=" + name,
+// watch is a wrapper around the k8s watch.Interface
+type watch struct {
+	parent  *watcher
```
marquiz: nit/question: do we need the whole parent instance here? Would just the k8sclient be enough? Would make the architecture prolly slightly cleaner.

pkg/agent/kubernetes.go
```diff
@@ -94,17 +97,166 @@ func patchNodeStatus(cli *k8sclient.Clientset, fields map[string]string) error {
...
+		events:  make(chan k8swatch.Event),
+		openfn:  open,
+		queryfn: query,
 	}
-	watch, err := cli.CoreV1().ConfigMaps(ns).Watch(listOpts)
-	if err != nil {
-		return nil, agentError("failed to create a watch for configmaps in ns %q: %v", ns, err)
+}
+
+// configNodeWatch creates a watch for k8s Node
```
marquiz: `newNodeWatch` instead of `configNodeWatch`

pkg/agent/kubernetes.go
```diff
@@ -94,17 +97,166 @@ func patchNodeStatus(cli *k8sclient.Clientset, fields map[string]string) error {
...
+			if err != nil {
+				return nil, err
+			}
+			return node, nil
+		})
+	w.Start(nodeName)
+	return w
+}
+
+// configMapWatch creates a watch for k8s ConfigMap
```
marquiz: newConfigMapWatch

pkg/agent/kubernetes.go
```diff
@@ -94,17 +97,166 @@ func patchNodeStatus(cli *k8sclient.Clientset, fields map[string]string) error {
...
+
+// Query queries the object being watched.
+func (w *watch) Query() (interface{}, error) {
+	if w.name == "" {
+		return nil, nil
+	}
+	return w.queryfn(w.ns, w.name)
+}
+
+// Start restart watching a potentially new object.
```
marquiz: "Start restart"?

pkg/agent/kubernetes.go
```diff
@@ -94,17 +97,166 @@ func patchNodeStatus(cli *k8sclient.Clientset, fields map[string]string) error {
...
+	ns      namespace
+	name    string
+	openfn  func(namespace, string) (k8swatch.Interface, error)
+	queryfn func(namespace, string) (interface{}, error)
+	stop    chan struct{}
+	events  chan k8swatch.Event
+}
+
+type openFn func(ns namespace, name string) (k8swatch.Interface, error)
+type queryFn func(ns namespace, name string) (interface{}, error)
```
marquiz: nit: add a linefeed between? With short comments(?)

pkg/agent/watcher.go
```diff
@@ -21,16 +21,18 @@ import (
 	"time"
 
 	core_v1 "k8s.io/api/core/v1"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
+	//meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
marquiz: nit: remove this commented out line

pkg/agent/watcher.go
```diff
@@ -21,16 +21,18 @@ import (
...
+	k8swatch "k8s.io/apimachinery/pkg/watch"
 	k8sclient "k8s.io/client-go/kubernetes"
 
 	"github.com/intel/cri-resource-manager/pkg/log"
 )
 
 type cachedConfig struct {
 	sync.RWMutex
-	config resmgrConfig
```
marquiz: Related to the longer comment below, should we have `defaultCfg` here, too(?)

pkg/cri/resource-manager/config/server.go
```diff
@@ -122,6 +124,9 @@ func (s *server) Stop() {
 
 // GetNode gets K8s node object.
 func (s *server) SetConfig(ctx context.Context, req *v1.SetConfigRequest) (*v1.SetConfigReply, error) {
+	s.Lock()
+	defer s.Unlock()
```
marquiz: Hmm, doesn't the channel (below) ensure thread safety?

pkg/agent/watcher.go
klihub: Well, we track the 2 necessary of the 3 ConfigMaps. We either track the group or the default one, but never both simultaneously.

Initially we used to track all 3 ConfigMap simultaneously and the logic for missing group configuration was what you suggest now. I deliberately changed it to be what it is now. I don't have a strong opinion about this, but it felt more natural to have it this way.

pkg/agent/watcher.go
klihub: I was thinking about that for cases where we want a different node-local default than the built-in one. However, in other PR's we've been taking out bells and whistles that we were not using at the moment, I decided to omit it and add it only later if it becomes 100 % sure we'll both want and need it.

pkg/cri/resource-manager/config/server.go
```diff
@@ -90,7 +123,8 @@ func (s *server) Stop() {
 // GetNode gets K8s node object.
 func (s *server) SetConfig(ctx context.Context, req *v1.SetConfigRequest) (*v1.SetConfigReply, error) {
 	s.Debug("REQUEST: %s", req)
-	return &v1.SetConfigReply{}, s.setConfigCb(&RawConfig{NodeName: req.NodeName, Data: req.Config})
+	s.requests <- req
+	return &v1.SetConfigReply{}, nil
```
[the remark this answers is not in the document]
klihub: https://github.com/intel/cri-resource-manager/pull/53#issue-332274584https://github.com/intel/cri-resource-manager/pull/53#issue-332274584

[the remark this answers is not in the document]
klihub: I can only quote myself from #53 , which is the PR introducing that change and explains the reasoning and potential remedy.
```
Also always reply with success for configuration updates. The old behavior of replying with an error and then indefinitely trying to resubmit the exact same already rejected data does not make sense. We can start reporting configuration failures asynchronously once we figure out how and where to propagate the error back. In the stock agent case, it could be propagated back to the offending ConfigMap.
```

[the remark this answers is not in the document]
klihub: About getting stuck in the old configuration... IMO there is exactly zero point in trying to push a rejected configuration over and over to cri-resmgr, hoping that it suddenly would start magically working.

And that is what was happening with the old implementation. It did not differentiate between a failed configuration push and a successful configuration push that failed to get activated. It just kept retrying forever for both. It is the right thing to do for the former case and the wrong thing to do for the latter.

pkg/agent/kubernetes.go
klihub: It is mostly there so we can log using the watcher's logger from the 'bottom part' of the watcher, the watch. I could have passed over it just as a log.Logger but I don't see the point.

pkg/cri/resource-manager/config/server.go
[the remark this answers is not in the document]
klihub: That said, let me rework this so that I move the ratelimiting also to the agent side. That way we won't waste the reply error.

pkg/agent/kubernetes.go
klihub: Thanks, fixed.

pkg/agent/kubernetes.go
klihub: Thanks fixed.

pkg/agent/kubernetes.go
klihub: Fixed comment. It used both to start and restart a watch.

pkg/agent/kubernetes.go
klihub: Fixed.

pkg/agent/watcher.go
klihub: Thx, fixed.

pkg/cri/resource-manager/config/server.go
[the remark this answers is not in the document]
klihub: See PR #53 (rate-limiting moved to the agent side) and #56 (separate errors for update delivery vs. application errors).

klihub:
Yes, I agree that I think it is next to impossible to ever run into this in practice. It's just that since I spotted it, it bothered me because of my previous frequent encounters with the legacy of a certain Mr. Murphy. Plus it was literally two lines of code, one import and one declaration.

PR #56, opened 2019-10-25
config: return errors related to applying updates in the reply message

klihub:
This PR is stacked on (includes) PR #54 .

    Return configuration update errors as a field in the reply message.
    This allows the agent to differentiate between errors in delivery
    vs. errors in applying an update.
    
    For delivery errors the agent will keep retrying indefinitely, sending
    an update request at periodic intervals. For errors in applying an
    update the agent will give up immediately and only try sending a new
    update when the configration is updated the next time.

### DIFF
--- a/pkg/agent/config-updater.go
+++ b/pkg/agent/config-updater.go
@@ -18,6 +18,7 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"time"
 
@@ -32,6 +33,8 @@ const (
 	rateLimitTimeout = 2 * time.Second
 	// setConfigTimeout is the duration we wait at most for a SetConfig reply
 	setConfigTimeout = 5 * time.Second
+	// retryTimeout is the timeout after we retry sending configuration updates upon failure
+	retryTimeout = 5 * time.Second
 )
 
 // configUpdater handles sending configuration to cri-resmgr
@@ -77,9 +80,14 @@ func (u *updater) Start() error {
 				ratelimit = time.After(rateLimitTimeout)
 
 			case _ = <-ratelimit:
-				if _, err := u.setConfig(pending); err != nil {
+				mgrErr, err := u.setConfig(pending)
+				if err != nil {
 					u.Error("failed to send configuration update: %v", err)
+					ratelimit = time.After(retryTimeout)
 				} else {
+					if mgrErr != nil {
+						u.Error("cri-resmgr error: %v", mgrErr)
+					}
 					pending = nil
 					ratelimit = nil
 				}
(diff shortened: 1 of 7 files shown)


klihub:
> This is good.
> 
> However, I'm still yearning for the "re-try logic" we had. I agree that it was bas design to
> have it in the agent side, trying to indefinitely push a failed configuration. However, I contend
> that there is point trying to apply a failed configuration.

I don't think it was bad design per se. It was more of being a bit overly simplistic not to differentiate between transport errors and application errors. Given that limitation (hence not being able to tell if delivery was successful at all), retrying indefinitely was probably the most sensible choice. However, with this patch in place the agent always knows if delivery was successful.

How the agent should react to application-level errors is not that obvious, at least not to me. I can see cases where retrying would make sense, for instance if the returned application error is indicated to be a transient one. Then again, if the application knows it to be a transient error, why would it not answer with success in the first place and keep retrying itself. 

The agent is in a pretty bad spot for deciding what to do, since it

1) does not have semantic knowledge about the application or its configuration,
2) does not know about, control, or event get notified about changes in the system configuration,

So, one sensible thing I can imagine on the agent side is to provide the means for the application to indicate whether reapplying the configuration should be tried later. This can be done in many ways, for instance by adding a field for it to the reply message.

> The system configuration underneath might change and the configuration might become
> functional. I've experienced this many times when hacking on CRI-RM.

Do you have a practical example about this ? Was it about RDT ?

> A similar thing
> would be re-configuring/bootstrapping a cluster where the admin notices that on a certain
> node the configuration does not apply because of reason X in the system configuration.
> When he fixes X it would be nice to see the configuration automatically applied, without
> restarting cri-resmgr or re-editing the configmap just in order to get an update.

Sure. Still, my gut feeling is that unconditionally retrying forever is not the right thing to do.

> Could you e.g. open an issue tracking this, WDYT?

I'll open an issue for this. It clearly needs more head scratching.

marquiz:
> > This is good.
> > However, I'm still yearning for the "re-try logic" we had. I agree that it was bas design to
> > have it in the agent side, trying to indefinitely push a failed configuration. However, I contend
> > that there is point trying to apply a failed configuration.
> 
> I don't think it was bad design per se. It was more of being a bit overly simplistic not to differentiate between transport errors and application errors. Given that limitation (hence not being able to tell if delivery was successful at all), retrying indefinitely was probably the most sensible choice. However, with this patch in place the agent always knows if delivery was successful.

Yes, this is definitely an improvement in the communication. I like the agent side, now.

> How the agent should react to application-level errors is not that obvious, at least not to me. I can see cases where retrying would make sense, for instance if the returned application error is indicated to be a transient one. Then again, if the application knows it to be a transient error, why would it not answer with success in the first place and keep retrying itself.

As I stated above, I think that the agent side prolly is doing what it should, now. I would implement what ever re-try logic or similar on the resource manager side. It would also be its (`cri-resmgr`´s) responsibility to update the node status or whatever to inform about the condition. WDYT?

> The agent is in a pretty bad spot for deciding what to do, since it 
> 1. does not have semantic knowledge about the application or its configuration,
> 2. does not know about, control, or event get notified about changes in the system configuration,

Yup, totally agree. That's why I'd keep the agent as "stupid" as possible. I think it's quite stupid, now 😄 


> So, one sensible thing I can imagine on the agent side is to provide the means for the application to indicate whether reapplying the configuration should be tried later. This can be done in many ways, for instance by adding a field for it to the reply message.

I would keep it as it is (at least that's what I'm thinking today 😉)

> > The system configuration underneath might change and the configuration might become
> > functional. I've experienced this many times when hacking on CRI-RM.
> 
> Do you have a practical example about this ? Was it about RDT ?

Yes, with RDT (e.g. too many CLOSes i.e. resctrl groups reserved by system admin).


> > A similar thing
> > would be re-configuring/bootstrapping a cluster where the admin notices that on a certain
> > node the configuration does not apply because of reason X in the system configuration.
> > When he fixes X it would be nice to see the configuration automatically applied, without
> > restarting cri-resmgr or re-editing the configmap just in order to get an update.
> 
> Sure. Still, my gut feeling is that unconditionally retrying forever is not the right thing to do.

Not on the agent side, at least. On the resource-manager side we (may) have better understanding about the situation.

> > Could you e.g. open an issue tracking this, WDYT?
> 
> I'll open an issue for this. It clearly needs more head scratching.

Cool, thanks!

PR #68, opened 2019-11-07
Switch to generic, YAML-based runtime configuration handling.

klihub:
The newly added pkg/config package attempts to provide a more
    generic, unified mechanism for handling runtime configuration.
    It models configuration data as a tree of modules. Each module
    has associated with it a description, a configuration fragment
    (IOW, the configuration data structure for the module), and a
    function to get the default configuration for the module. The
    configuration hierarchy is stitched together as the modules re-
    gister themselves to their desired path in the tree. Any parent-
    child relationship of modules has no other significance than to
    provide means for arbitrary logical grouping of configuration
    data.
    
    A module can also have one or more update notification callack
    functions associated with it, both during registration or later.
    Every time configuration is updated modules are notified using
    these notification callbacks. A callback can reject an update by
    returning an error. Rejected configuration will be automatically
    rolled back to its previous state.
    
    In the current implementation configuration can be updated from
    a YAML/JSON configuration file or from an external ConfigMap-
    like source, IOW  maps[string]string data.

    The other patches switch packages from flag-based command
    line configuration to runtime configurability and add dual command
    line/runtime configurability for logging and message dumping.


pkg/config/help.go
```diff
@@ -0,0 +1,149 @@
...
+				ftype = field.Type.String()
+			}
+			if fkind == ftype {
+				fkind = ""
+			} else {
+				fkind = " (" + fkind + ")"
+			}
+			fmt.Printf("      - '%s' of type %s%s\n", fname, ftype, fkind)
+		}
+	*/
```
marquiz: remove this commented out block?

pkg/config/data.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// Data is our internal representation of configuration data.
+type Data map[string]interface{}
+
+// DataFromObject remarshals the given object into configuration data.
+func DataFromObject(obj interface{}) (Data, error) {
+	raw, err := yaml.Marshal(obj)
+	if err != nil {
+		return nil, configError("failed to unmarshal object %T to data: %v", obj, err)
```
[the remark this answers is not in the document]
klihub: Fixed.

pkg/config/data.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	data := make(Data)
+	if err = yaml.Unmarshal(raw, &data); err != nil {
+		return nil, configError("failed to marshal object %T to data: %v", obj, err)
```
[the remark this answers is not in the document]
klihub: Fixed.

pkg/config/help.go
klihub: Fixed.

pkg/config/config.go
```diff
@@ -0,0 +1,418 @@
...
+			field := ptr.Type().Field(i)
+			name := fieldName(field)
+			fields[name] = struct{}{}
+		}
+	}
+
+	for name := range data {
+		if _, ok := fields[name]; !ok {
+			if _, ok := m.children[name]; !ok {
+				return configError("module %s has neither data nor child module '%s'",
```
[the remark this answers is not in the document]
klihub: Replaced those as suggested.

pkg/log/flags.go
```diff
@@ -54,283 +47,319 @@ var LevelNames = map[Level]string{
...
+
+	for key, value := range cfg {
+		switch key {
+		case "Level":
+			if err := o.level.Set(value); err != nil {
+				return err
+			}
+		case "Logger":
+			o.logger = backendName(value)
+			activateBackend(value)
```
marquiz: It's a bit  peculiar that the options are acivated applied in the json unmarshal phase when config parsing is still underway. But I guess logger is a special case and this is, if not inevitable, easier this way(?)

README.md
```diff
@@ -101,17 +101,29 @@ Node via an environment variable, as well as a valid kubeconfig.  For example:
 
 ## Running the relay with policies enabled
 
-You can enable active policying of containers by passing the
-`--policy <policy-name>` commandline option to the relay. For instance,
+You can enable active policying of containers by using an appropriate ConfigMap
+or a configuration file and setting the `Active` field of the `policy` section
```
marquiz: AFAIU you cannot dynamically change the active policy(?) If so, this could be mentioned

README.md
```diff
@@ -141,76 +153,41 @@ for more details.
...
+the Kubernetes using the the `-kubeconfig` command line option. The agent
+will monitor ConfigMaps, pick node-specific configuration and send it to
+cri-resmgr running on the same node.
+
+Configuration for the node is taken from one of these ConfigMaps in
+decreasing order of preference:
+
+1. cri-resmgr-config.node.$NODE_NAME: node-specific configuration
+2. cri-resmgr-config.group.$GROUP_NAME: group-specific node configuration
+3. cri-resmgr-config.default: default node configuration
```
marquiz: I still contend that this wording/description is wrong. We only track two configurations - we do not track the .default ConfigMap if the node has a group label. IOW the "group map" for nodes that do not have the group label is `cri-resmgr-config.default`

pkg/log/flags.go
```diff
@@ -54,283 +47,319 @@ var LevelNames = map[Level]string{
...
-			}
-		case "none":
-			if state {
-				flags = make(map[string]bool)
-			} else {
-				flags = nil
-			}
+	prev := "on"
+	state := true
+	for _, req := range strings.Split(strings.Trim(value, " \t"), ",") {
```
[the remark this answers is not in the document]
klihub: Fixed.

pkg/log/flags.go
klihub: In principle we could delay reconfiguring the logger until the notification callback comes in. But I agree that logging is a bit special and that's the reason we try to make a successful switch as early as possible.

README.md
klihub: Added  a note about this to README.md.

README.md
klihub: Rewrote the description to explicitly spell out the fact that we only monitor 2 of the potential ConfigMaps.

klihub:
> typo in commit message: * callack* should be _callback_

Well spotted, thanks ! Fixed.

pkg/log/flags.go
klihub: And I guess if this turns out to be confusing/disturbing in case of rollbacks after failure, we can always switch to reconfiguring from the callback.

sample-configs/cri-resmgr-configmap.example.yaml
```diff
@@ -1,94 +1,277 @@
...
+    Active: topology-aware
+      topology-aware:
+        PinCPU: true
+        PinMemory: true
+        PreferIsolatedCPUs: true
+        PreferSharedCPUs: false
+      static:
+        RelaxedIsolation: true
+      static-pools:
+        # This is an example configuration file for the builtin cmk policy
```
kad: we probably need to cleanup this "cmk"

klihub: Update config accordingly.

klihub:
I've updated this with the last changes I wanted to get in (separate validation phase from configuration). If there are no objections, I think we could merge this and address any further issues as new PRs.

PR #71, opened 2019-11-15
Internal metrics using Prometheus collectors

mythi:


mythi:
@klihub I've updated the registration based on our conversation. Is this approach OK?

pkg/metrics/builtin/register.go
```diff
@@ -0,0 +1,15 @@
+package builtinmetrics
```
rojkov: The go coding style recommends to avoid tautological package names. `builtin` would be enough here. Also this way it would be easier to understand where `RegisterMetric()` comes from in the file `pkg/avx/collector.go` (currently it's not clear whether `builtinmetrics` is a package name or a global object defined in another file).

mythi: Agree. I was struggling with the naming and this version was not very good. I've now re-worked the registration. PTAL.

pkg/avx/collector.go
```diff
@@ -253,3 +254,13 @@ func (c collector) collectLastCPUStats(ch chan<- prometheus.Metric) {
 			fmt.Sprintf("CPU%d", binary.LittleEndian.Uint32(lastCPU)))
 	}
 }
+
+func init() {
+	// TODO: collector config
+	c, err := newCollector("/sys/fs/cgroup/unified", "./libexec/avx512.o")
```
rojkov: I guess adding configureability to `init()` may lead to problems in unit tests later. Can we have this collector registration more explicit in the consumer code (in `cmd/gatherer/gatherer.go`)?

go.mod
```diff
@@ -9,8 +9,12 @@ require (
 	github.com/golang/protobuf v1.3.1
 	github.com/google/go-cmp v0.3.0
 	github.com/iovisor/gobpf v0.0.0-20191024162143-7c8f8e040b4b
+	github.com/json-iterator/go v1.1.6 // indirect
```
klihub: Hmm... BTW are you still compiling with golang < 1.13, or is this just one of those cases where go mod for some reason still inserts indirect dependencies (also) to go.mod ?

mythi: I'm using Go 1.13 (on Clear Linux). I'll see if I can clean this up.

pkg/metrics/register/register_avx.go
```diff
@@ -0,0 +1,5 @@
+package register
+
+import (
+	_ "github.com/intel/cri-resource-manager/pkg/avx" // To call init()
```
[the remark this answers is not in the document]
mythi: the idea is there would be a 1:1 mapping of `register_xyz.go` (with one import) and xyz collector implementation.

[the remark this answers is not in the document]
klihub: Is that really a good idea ?

go.mod
klihub: Don't worry, probably no need for it. If you're already on 1.13 then my guess is it's just go mod being buggy. We were wondering about a very similar case with @marquiz  the other day. Eventually we couldn't figure out how to convince go mod not to add a few indirect dependencies to go.mod.

pkg/metrics/register/register_avx.go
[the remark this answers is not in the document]
mythi: I ran into this approach in container storage repo where all graphdrivers are registered like this. This way it's easy to see going-to-be registered collectors by checking the files in the folder. TBH, I'm flexible and register_metrics.go is also fine.

pkg/avx/collector.go
```diff
@@ -38,6 +40,8 @@ var (
 			"cgroup",
 		}, nil,
 	)
+
+	BpfBinaryName = "avx512.o"
```
rojkov: for now this can be `const`.

pkg/cgroups/cgroupid.go
```diff
@@ -33,7 +33,7 @@ func getID(path string) uint64 {
 func (cgid *CgroupID) List() error {
 	return filepath.Walk(cgid.root, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			fmt.Printf("WalkFunc called with an error (path %q: %v\n)", path, err)
+			fmt.Printf("WalkFunc called with an error (path %q: %v)", path, err)
```
rojkov: no new line?

pkg/avx/collector.go
mythi: OK with the change I'm proposed?

pkg/avx/collector.go
mythi: AFAIK `const`s cannot be overridden using `-ldflags -X` which was my idea to keep available..

pkg/cgroups/cgroupid.go
mythi: my bad, will fix

pkg/avx/collector.go
rojkov: Yes, looks good now

pkg/avx/collector.go
rojkov: I see. Let's have it var then

klihub: This (exported var) should have a docment to make go lint shut up.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+		time.Sleep(m.period)
+
+		if m.isClosing() {
+			close(m.events)
+			return
+		}
+
+		g, err := m.gatherer.Gather()
+		if err != nil {
+			fmt.Println(err)
```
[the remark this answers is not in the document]
mythi: Your patch added to this PR.

pkg/avx/collector.go
mythi: fixed.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	if _, fn := builtInCollectors[name]; fn {
+		return metricsError("Collector %s already registered", name)
+	}
+
+	builtInCollectors[name] = init
+
+	return nil
+}
+
+var registeredCollectors = []prometheus.Collector{}
```
rojkov: nit: I'd rather have one `var` section per module to observe what constitutes module's state in one place.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+func init() {
+	builtInCollectors = make(map[string]InitCollector)
+}
+
+type InitCollector func() (prometheus.Collector, error)
+
+func RegisterCollector(name string, init InitCollector) error {
+	log.Info("registering collector %s...", name)
+
+	if _, fn := builtInCollectors[name]; fn {
```
rojkov: `fn` is often meant to represent `function`. Perhaps `found` would be clearer here.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	log               = logger.NewLogger("metrics")
+)
+
+func init() {
+	builtInCollectors = make(map[string]InitCollector)
+}
+
+type InitCollector func() (prometheus.Collector, error)
+
+func RegisterCollector(name string, init InitCollector) error {
```
rojkov: nit: `init` overshadows the `init()` function here. Probably makes sense to choose a different name for the arg to avoid confusions.

pkg/avx/collector.go
```diff
@@ -253,3 +276,18 @@ func (c collector) collectLastCPUStats(ch chan<- prometheus.Metric) {
...
+
+func init() {
+	err := metrics.RegisterCollector("avx", NewCollector)
+	if err != nil {
+		log.Error("Failed to register AVX collector: %v", err)
+		return
+	}
+}
+
+func init() {
```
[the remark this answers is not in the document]
mythi: a package can have multiple `init()` functions.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	builtInCollectors map[string]InitCollector
+	log               = logger.NewLogger("metrics")
+)
+
+func init() {
+	builtInCollectors = make(map[string]InitCollector)
```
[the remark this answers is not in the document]
mythi: yes

pkg/metrics/metrics.go
mythi: +1

pkg/metrics/metrics.go
mythi: will change

pkg/metrics/metrics.go
mythi: true, I'll rename

mythi: per our chat, let's keep it

PR #83, opened 2019-11-26
CI: build avx512 eBPF

mythi:

### DIFF
--- a/elf/avx512.c
+++ b/elf/avx512.c
@@ -1,6 +1,4 @@
-#include <linux/version.h>
 #include <uapi/linux/bpf.h>
-
 #include <asm/page_types.h>
 
 /* asm/fpu/types.h assumes __packed is defined */
@@ -8,6 +6,10 @@
 #include <asm/fpu/types.h>
 
 #define SEC(NAME) __attribute__((section(NAME), used))
+
+#ifndef KERNEL_VERSION
+    #define KERNEL_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + (c))
+#endif
 
 #define BUF_SIZE_MAP_NS 256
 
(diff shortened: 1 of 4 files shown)


elf/avx512.c
```diff
@@ -9,6 +8,9 @@
 
 #define SEC(NAME) __attribute__((section(NAME), used))
 
+/* Host kernel needs to run >= 5.2.0 (the version is checked upon eBPF loading). */
+#define LINUX_VERSION_CODE 328192
```
mythi: @rojkov PTAL

rojkov: With this change we assume that the layouts of all used structs (like `struct sched_switch_args` or `struct x86_fpu_args`) are never changed and the ELF built on a host with kernel v3.16 will use correct offsets to structs' fields on a host running v5.2. I don't know if structs in kernel are that stable, but if they are I'm good with change.

mythi: AFAICS, the build host version can be anything but you need new enough headers.  Currently we don't check the headers are suitable (but compiler errors, e.g., if fpu/types.h is too old). The tracepoint ABI overall is an open conversation in the kernel community based on what I've seen. struct layouts are not stable (as we can see from `avx512_timestamp` itself)

We could also keep the exact major.minor version match but it cannot be based on `<linux/version.h>` but something the targeted kernel version headers provide.

rojkov: Fair enough. Looks good to me.

klihub: Hmm... wouldn't something like this be much easier to understand, especially for those that are somewhat familiar with the kernel:

```
+
+#ifndef KERNEL_VERSION
+#  define KERNEL_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + (c))
+#endif
+
+#ifndef LINUX_VERSION_CODE
+#  define LINUX_VERSION_CODE KERNEL_VERSION(5, 2, 0)
+#endif
+
```

mythi: yup, making it more readable makes sense. I'm also thinking we should probably add a build time  `version.h` check and `#error` too if the version is not right.

mythi:
@klihub travis is green with my latest changes.

PR #84, opened 2019-11-26
devel: apply initial configuration before startup.

klihub:
This patch set changes the startup sequence to fetch the latest known configuration and apply it before starting any policy/request processing. This allows, in principle, the active policy to boot
directly into its latest up-to-date configuration.

Moreover, loading this initial configuration is attempted, in decreasing order of preference, from
      - forced configuration (for development, disables updates),
      - the agent,
      - last up-to-date configuration stored in the cache,
      - fallback configuration (also if cached one fails to apply)

The original active/saved policy check recently removed from the cache is
now attempted by the resource manager once the initial configuration is
loaded.

***Notes***: The extra squashme-commits take care of modifying the unit-tests and mocked interfaces according to the changes pending in this patch set. They are split out to help (avoid) re-reviewing and will be squashed before this PR is merged.

### DIFF
--- a/cmd/cri-resmgr/main.go
+++ b/cmd/cri-resmgr/main.go
@@ -29,12 +29,10 @@ import (
 )
 
 func main() {
-	var configFile string
 	var printConfig bool
 
 	log := logger.Default()
 
-	flag.StringVar(&configFile, "config", "", "Read initial configuration from given file.")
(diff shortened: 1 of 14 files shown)


klihub:
> Looks good!! Just one commnent: shouldn't you drop `PostStart()` from stp-policy, too?

Yes, definitely. I'll update.

> This is dropping RDT support so I wouldn't merge until we're happy with #85

I agree.

PR #85, opened 2019-11-27
devel: generalize resource control, add support for RDT and Block I/O policying.

klihub:
*Note: this patch series depends on PR #84 .*

This patch series is an attempt to generalize the existing RDT-only resource control in resource manager. It adds a new `control(ler)` and related interfaces as a counterpart to the existing `policy` and its interfaces. Much like policies are responsible for making *abstract* `decisions` for containers, controllers are now responsible for the *concrete* actions to `enforce` those decisions in practice.

The controller interface defines a set of *hooks* which correspond to various points in the life cycle of a container. The controllers use these to inject themselves into the CRI request processing pipeline and enforce policy decisions in their jurisdiction. The current set of hooks is `PreCreateHook`, `PreStartHook`, `PostStartHook`, `PostUpdateHook`, `PostStopHook`. This list is neither a definitive set of all hooks imaginable nor all hooks are used at the moment. Unnecessary ones can be removed and missing ones added if there is a need.

The patch series introduces 3 initial controllers: `cri`, `rdt`, and `blockio`. The `cri` controller takes care of enforcing anything that can be enforced by either modifying a relayed request or sending a new one to the CRI runtime. The `rdt` controller uses the *resctrl pseudofilesystem* to care of enforcing decisions related to last-level cache or memory bandwidth allocations. The `blockio` controller should take care of enforcing block I/O-related policies (bandwidth allocation, etc) using the *blkio cgroup controller* , but it is currently just a stub.

Functions have been added to the cache container abstraction for both `rdt` and `blockio` policy decision making. Policy implementations can use these to set the name of the `RDT` and `Block I/O` classes containers should be assigned to. Currently both the `rdt` and the `blockio` controllers internally default to using the name of the `QOS class` if nothing has been explicitly set by the active policy.

Request processing has been reworked moving much of the processing/modification of CRI requests to controllers, more precisely to the `cri` controller. Related to this, the cache has been updated to track, per container, the `controller domains` with *pending, unenforced* decisions. As a result, the policy interface has been slightly simplified removing the need and interfaces for cache- or policy-level transactions. Similarly, supporting functions have been added for administering pending changes for containers, and for updating CRI requests from outside of the cache.

One choice to be made for the CRI processing overhaul was whether actual *communication* with the `CRI runtime` should be pushed to the `cri` controller or kept at the upper layer in the `resource manager proper`. I ended up keeping communication fully in the `upper layer`. The `controller` is  responsible for both *modifying* relayed (CRI container creation) requests and *creating* unsolicited (CRI container resource update) requests.  These are then *relayed/sent* from the `upper layer` (and communicated between the two using the container and the related functions).

The other reasonable compromise would have been to *push all communication down* to the `controller` with the `upper layer` taking care of orchestrating overall request processing, including policy *decision making*, controller *invocation*, and any necessary *error handling*. I'm not sure if either of these choices provides any significant benefits compared to the other. My gut feeling was that with the current split readability (and following error paths) is probably easier. But as I said, I'm not sure...

Related to *error handling*, whether failure in a controller in various stages is considered a fatal error is currently determined based on the configuration. Controllers can be marked as
  - `Disabled`: controller is not started (or stopped if running), no hooks triggered
  - `Required`: controller must be able to start, hooks must succeed
  - `Optional`: controller is `Disabled` if it can't start, otherwise it is `Required`
  - `Relaxed`: controller is `Disabled` if it can't start, hook failures are logged but not propagated up

With this setup, a *policy* cannot indicate for an `rdt` or `blockio` decision if it must succeed or is allowed to fail as this info comes from the *configuration*. If we want more dynamic choices for a policy, for instance to let them require decisions to be enforced for `Guaranteed` Pods but allow them to fail for others, then we need to make that preference an explicit part of the `rdt` and `blockio` policy decisions.

The patch series *is deliberately* written with separate patches introducing the new controller-based request processing logic and removing the old one, to help testing and debugging during development. A temporary command line options is in place to select which code path is active. The related patches can be *squashed* before merging.

Likewise there are separate patches to introduce the basic controller mechanism (together with the more straightforward `rdt` and `blockio` controllers) and to implement the full controller based CRI request processing. These are merely kept separate for *easing the review burden* of such a large patch series and they can be *squashed* before merging.


klihub:
@kad  @marquiz  Any chances of getting this reviewed ? #84 could go in, but IIRC, we tentatively agreed not to merge it (since it temporarily removes RDT support) before this one is ready to go in as well (since this one readds RDT support).

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -284,6 +284,16 @@ type Container interface {
 
 	// GetAffinity returns the annotated affinity expressions for this container.
 	GetAffinity() []*Affinity
+
+	// SetRDTClass assigns this container to the given RDT class.
+	SetRDTClass(string)
+	// GetRDTClass returns the RDT class for this container.
+	GetRDTClass() string
```
marquiz: Just a thought: could this be a generic `SetResourceClass(resType string, class string)` or similar? Would make extensions in the future easier (with `resType` prolly a type of its own instead of a string).

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -294,6 +298,20 @@ type Container interface {
 	SetBlockIOClass(string)
 	// GetBlockIOClass returns the BlockIO class for this container.
 	GetBlockIOClass() string
+
+	// SetCRIRequest sets the current pending CRI request of the container.
+	SetCRIRequest(req interface{}) error
+	// GetCRIRequest returns the current pending CRI request of the container.
+	GetCRIRequest() (interface{}, bool)
+	// ClearCRIRequest returns the current pending CRI request of the container.
```
marquiz: Incorrect docstring?

pkg/cri/resource-manager/cache/container.go
```diff
@@ -319,6 +319,17 @@ func (c *container) GetResmgrLabel(key string) (string, bool) {
 	return value, ok
 }
 
+func (c *container) GetLabels() map[string]string {
+	if c.Labels == nil {
+		return nil
+	}
+	labels := make(map[string]string)
```
marquiz: nit: you could use `len(c.Labels)` as the size?

pkg/cri/resource-manager/cache/container.go
```diff
@@ -356,6 +367,17 @@ func (c *container) GetResmgrAnnotation(key string, objPtr interface{}) (string,
 	return c.GetAnnotation(kubernetes.ResmgrKey(key), objPtr)
 }
 
+func (c *container) GetAnnotations() map[string]string {
+	if c.Annotations == nil {
+		return nil
+	}
+	annotations := make(map[string]string)
```
marquiz: similar length nit...

pkg/cri/resource-manager/policy/policy.go
```diff
@@ -266,6 +268,13 @@ func (p *policy) UpdateResources(c cache.Container) error {
 	return p.backend.UpdateResources(c)
 }
 
+// ExportResourceData exports/updates resource data for the container.
+func (p *policy) ExportResourceData(c cache.Container, syntax DataSyntax) {
```
marquiz: What is this `syntax` here for?

pkg/utils/cgroups.go
```diff
@@ -51,8 +51,8 @@ func GetContainerCgroupDir(subsystemDir, containerID string) string {
 }
 
 // GetProcessInContainer gets the IDs of all processes in the container.
-func GetProcessInContainer(cgroupParentDir, containerID string) ([]int, error) {
-	var pids []int
+func GetProcessInContainer(cgroupParentDir, containerID string, pidsptr *[]int) ([]string, error) {
```
marquiz: We could drop this `pidsptr` if it's not used anywhere?

pkg/cri/resource-manager/cache/cache.go
klihub: True, these could be combined and/or generalized. We can do it if we start getting more of these.

pkg/cri/resource-manager/cache/cache.go
klihub: Yes it should be `clears and returns the current`. Will fix.

pkg/cri/resource-manager/cache/container.go
klihub: Fixed.

pkg/cri/resource-manager/cache/container.go
klihub: Fixed.

pkg/cri/resource-manager/policy/policy.go
klihub: The original idea was to support exporting the same policy-specific data in different formats, for instance as JSON or as a /bin/sh-syntax source-able file. Only the latter has been implemented.

It'd be probably more reasonable to let the policy backend only supply raw data (as a map[string]string) for the container and do the actual formatting in the agnostic layer.

pkg/utils/cgroups.go
klihub: Removed it.

pkg/cri/resource-manager/control/flags.go
```diff
@@ -0,0 +1,120 @@
...
+		return "relaxed"
+	default:
+		return fmt.Sprintf("<unknown mode %d>", m)
+	}
+}
+
+// MarshalJSON is the JSON marshaller for mode.
+func (m mode) MarshalJSON() ([]byte, error) {
+	str := m.String()
+	return json.Marshal(str)
```
marquiz: nit: simply `return json.Marshal(m.String())` ?

pkg/cri/resource-manager/control/flags.go
```diff
@@ -0,0 +1,120 @@
...
+		*m = Optional
+	case "relaxed":
+		*m = Relaxed
+	default:
+		return controlError("invalid mode %s", str)
+	}
+	return nil
+}
+
+// defaultOptions returns a new options instance, al linitialized to defaults.
```
marquiz: "*al linitialized*"??

pkg/cri/resource-manager/control/rdt/doc.go
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+
+package rdt
+
+var configHelp = `
+Resource Manager RDT enforcement controller.
+
+The RDT controller enforces container policy RDT decisions. The controller
+takes the assigned RDT class of a container and puts all processes in the
+containers' cgroup to that class using the resctrl pseudo-filesystem. If
```
marquiz: nit: **container's** instead of **containers'** (?)

pkg/cri/resource-manager/control/rdt/flags.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/intel/cri-resource-manager/pkg/config"
+	"github.com/intel/cri-resource-manager/pkg/rdt"
+)
+
+// options captures our configurable parameters.
+type options struct {
+	// ResctrlPath is the mount point of the resctrl pseudo-filesystem.
+	ResctrlPath string `json:",omitempty"`
+	// Class is a assigned to actual RDT class map.
+	Classes map[string]string `json:",omitempty"`
```
marquiz: Do we really want to use `omitempty` in the json tags? (the same question applies to all similar occurrences in other files, too)

pkg/cri/resource-manager/control/rdt/rdt.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/intel/cri-resource-manager/pkg/config"
+	logger "github.com/intel/cri-resource-manager/pkg/log"
+)
+
+const (
+	// RDTController is the name of the RDT controller.
+	RDTController = cache.RDT
+)
+
+// rdtctl encapsulates the runtime state of our RTD enforcment/controller.
```
marquiz: nit: s/enforcment/enforcement/

pkg/cri/resource-manager/control/rdt/rdt.go
```diff
@@ -0,0 +1,157 @@
...
+type rdtctl struct {
+	rdt   *rdt.Control // resctrl RDT control
+	cache cache.Cache  // resource manager cache
+}
+
+// Make sure rdtctl implements the control.Controller interface.
+var _ control.Controller = &rdtctl{}
```
marquiz: Is this check necessary? I think this is verified in other places, anyway, e.g. in getRDTController

pkg/cri/resource-manager/control/rdt/rdt.go
```diff
@@ -0,0 +1,157 @@
...
+
+	ctl.rdt = &rdtc
+	ctl.cache = cache
+
+	return nil
+}
+
+// Stop shuts down the controller.
+func (ctl *rdtctl) Stop() {
+}
```
marquiz: Just a mental note that maybe we should implement start/stop (or similar) in pkg/rdt, too(?) I.e. Stop() would delete all resctrl groups (managed by cri-rm) and thus return tasks within to the default/root group.

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -0,0 +1,147 @@
...
+package blockio
+
+import (
+	"fmt"
+
+	"github.com/intel/cri-resource-manager/pkg/cri/client"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/control"
+	"github.com/intel/cri-resource-manager/pkg/utils"
+
```
marquiz: nit: unintended extra linefeed?

pkg/cri/resource-manager/control/cri/cri.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// crictl encapsulated the runtime state of our CRI enforcement/controller.
+type crictl struct {
+	cache  cache.Cache
+	client client.Client
+}
+
+// Make sure crictl implements the control.Controller interface.
+var _ control.Controller = &crictl{}
```
marquiz: Is there a good reason for these?

pkg/cri/resource-manager/control/cri/cri.go
```diff
@@ -0,0 +1,141 @@
...
+	var update *criapi.UpdateContainerResourcesRequest
+
+	log.Debug("running post-update hook for %s...", c.PrettyName())
+
+	resources := c.GetLinuxResources()
+	if resources == nil {
+		return nil
+	}
+	request, ok := c.GetCRIRequest()
+	if !ok {
```
marquiz: Just a dumb question: when/why does this happen?

pkg/cri/resource-manager/control/cri/cri.go
```diff
@@ -0,0 +1,141 @@
...
+package cri
+
+import (
+	"fmt"
+
+	"github.com/intel/cri-resource-manager/pkg/cri/client"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/control"
+	criapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+
```
marquiz: nit: unintentional linefeed?

pkg/cri/resource-manager/control/control.go
```diff
@@ -0,0 +1,260 @@
...
+
+package control
+
+import (
+	"fmt"
+	"github.com/intel/cri-resource-manager/pkg/cri/client"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	logger "github.com/intel/cri-resource-manager/pkg/log"
+	"sort"
+	"strings"
```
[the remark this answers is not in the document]
klihub: I'd love to, but gofmt, the worst formatting and tabulation nazi in human history, disagrees and will insist on reformatting it differently.

pkg/cri/resource-manager/control/rdt/flags.go
klihub: Well, if I understand it correctly, if we don't tag it with `omitempty` the configuration always must explicitly set a value for it otherwise Unmarshal()ling will fail. IMO the biggest problem with that would be that it leaves no nice way to let something to default to a reasonable built-in value: instead of leaving something unset for the default you'd typically have to set it to the golang zero value of its internal type. So generally speaking, AFAICT all configuration fields with a reasonable built-in default should be tagged with `omitempty`.

pkg/cri/resource-manager/control/control.go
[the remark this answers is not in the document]
klihub: Hmm... actually I was wrong. It looks like gofmt is not doing it. I'll change it if all of of vet, lint, etc. accept it.

pkg/cri/resource-manager/control/rdt/flags.go
marquiz: AFAIU the `omitempty`s are only used in marshalling, not unmarshalling. If a field is not present unmarshalling will just use an "empty" value

pkg/cri/resource-manager/control/flags.go
klihub: Fixed.

pkg/cri/resource-manager/control/flags.go
klihub: Fixed.

pkg/cri/resource-manager/policy/policy.go
klihub: I moved formatting out of the backends to policy/policy.go with the backends only providing raw data in a map. Got rid of DataSyntax and the related constants. We always export in shell syntax only now.

pkg/cri/resource-manager/control/rdt/rdt.go
klihub: Fixed.

pkg/cri/resource-manager/control/rdt/rdt.go
klihub: I've been using it as an idio(ma)tic pattern everywhere, even when leaving it out would cause compilation a failure because of other code to the same effect (like with getRDTController). I don't know what the preferred go way.

pkg/cri/resource-manager/control/rdt/flags.go
marquiz: Actually, if the field is not present unmarshalling just leaves it as it was (does not replace with an "empty" value or anything)

pkg/cri/resource-manager/control/rdt/rdt.go
klihub: Yes, that's something we need to figure out for rdt, and also for the block I/O controller once we start filling in the stubs.

pkg/cri/resource-manager/control/control.go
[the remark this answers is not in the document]
klihub: Fixed.

pkg/cri/resource-manager/control/rdt/doc.go
klihub: Fixed.

pkg/cri/resource-manager/control/rdt/flags.go
klihub: Ah, ok. Then we could remove it from a lot of places. But I'd prefer to do that with a single commit addressing all misuses of `omitempty`.

klihub: Fixed (those introduced in this patch set).

pkg/cri/resource-manager/control/cri/cri.go
klihub: Same as above. Idiomatic pattern I used.

pkg/cri/resource-manager/control/cri/cri.go
klihub: It shouldn't happen if nothing is screwed up in request processing. I put the check here because a package boundary is involved here: the request is set (associated with the container) in resource-manager, then modified here in the CRI controller (and finally consumed again in the resource-manager).

pkg/cri/resource-manager/control/cri/cri.go
klihub: Fixed.

klihub:
> I started to wonder do we want to have this default enforcement to Pod QoS classes if the policy does not do any decisions? Is it a (unpleasant) "surprise" for policy writers that we do rdt/block-io enforcement even if the policy does not touch these?
> 
> It would be sort of simple and clean design if an effectively "empty" policy would act as a "pass-through" with no additional modifications done to the pod/container.

While you can already do that for empty/passthrough policies by disabling the RDT or blockio controllers entirely, I think that is a fair and valid point/concern. I think it somewhat falls to the same bin about more flexibility/control for policies about enforcement, which is already mentioned in the intro text to this patch set:

> With this setup, a policy cannot indicate for an rdt or blockio decision if it must succeed or is allowed to fail as this info comes from the configuration. If we want more dynamic choices for a policy, for instance to let them require decisions to be enforced for Guaranteed Pods but allow them to fail for others, then we need to make that preference an explicit part of the rdt and blockio policy decisions.

Considering that we're not sure how these things should be exactly handled and that we have other PR's waiting for this to go in, I'd prefer to address those concerns with further patches/PRs, once we have figured out the exact details of the desired behavior and functionality.

pkg/cri/resource-manager/control/cri/cri.go
klihub: Removed these from rdt, blockio, and the cri controller.

marquiz:
> > With this setup, a policy cannot indicate for an rdt or blockio decision if it must succeed or is allowed to fail as this info comes from the configuration. If we want more dynamic choices for a policy, for instance to let them require decisions to be enforced for Guaranteed Pods but allow them to fail for others, then we need to make that preference an explicit part of the rdt and blockio policy decisions.
> 
> Considering that we're not sure how these things should be exactly handled and that we have other PR's waiting for this to go in, I'd prefer to address those concerns with further patches/PRs, once we have figured out the exact details of the desired behavior and functionality.

Yeah, sure, there is no point in trying to do everything in one PR


> If that is considered to not be good enough, I think we can easily add an option to enable/disable whether the RDT- and Block I/O classes default to the Pod QoS class if otherwise unset.
> 
> @marquiz What do you think ?

My only question/concern was if we should default to Pod QoS class name if nothing is set by the policy (or just leave it empty instead). I really don't know the "correct" answer to this, just started to ponder the question. Any other opinions.

klihub:
> > I started to wonder do we want to have this default enforcement to Pod QoS classes if the policy does not do any decisions? Is it a (unpleasant) "surprise" for policy writers that we do rdt/block-io enforcement even if the policy does not touch these?
> > It would be sort of simple and clean design if an effectively "empty" policy would act as a "pass-through" with no additional modifications done to the pod/container.

I updated the RDT and Block I/O controllers so that
  - the class mapping is initialized to an identity-map of the 3 QoS classes
  - if a class is maps to the empty string, no assignment is done

The default behavior is still the same but this allows one to disable any RDT or Block I/O assignment based on the QoS classes.

marquiz:
Looks good

> * the class mapping is initialized to an identity-map of the 3 QoS classes

I think these are redundant and could be removed

PR #90, opened 2019-12-04
add unit tests to topology nodes

rojkov:
This is rather a suggestion for how the code base can be reorganized, not final code.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -323,6 +323,11 @@ func (sys *System) NodeIDs() []ID {
 	return ids
 }
 
+// PackageNodeIDs returns NUMA node ids for a given package.
+func (sys *System) PackageNodeIDs(id ID) []ID {
+	return sys.Package(id).NodeIDs()
+}
+
 // CPUIDs gets the ids of all CPUs present in the system.
 func (sys *System) CPUIDs() []ID {
 	ids := make([]ID, len(sys.cpus))
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -70,7 +70,7 @@ type System struct {
 // Package is a physical package (a collection of CPUs).
 type Package struct {
 	id    ID    // package id
-	cpus  IDSet // CPUs in this package
+	Cpus  IDSet // CPUs in this package
 	nodes IDSet // nodes in this package
 }
 
@@ -79,7 +79,7 @@ type Node struct {
 	path     string // sysfs path
 	id       ID     // node id
 	pkg      ID     // package id
-	cpus     IDSet  // cpus in this node
+	Cpus     IDSet  // cpus in this node
 	distance []int  // distance/cost to other NUMA nodes
 }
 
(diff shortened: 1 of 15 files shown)


rojkov:
Yes, that's correct. 

The primary goal is to make the code testable preferably without excessive boilerplate. The scope is limited to the `topologyaware` package.

If getters/setters are not used for checking invariants they are not needed usually and can be replaced with direct access to the fields. Go doesn't support data encapsulation in the scope of a package and having trivial accessors doesn't add value.

But the getters/setters become inevitable even in the scope of the same package if interfaces are used for type polymorphism solely (not even exported for external use). Considering the amount of boilerplate code required to exploit this feature (and corresponding tests) I'd prefer to replace it with function pointers set by constructors or even with `switch`es. This way the code of `node.go` is ~300 lines shorter.

Also I add test cases and mark them `SKIPPED` if the current code isn't testable (either requires interfaces or uses `os.Exit()` in non-main functions).

klihub:
All the unit tests and obvious cleanups/improvements, like replacing string consts prone to be switch'ed on with integer ones + String() are pretty much no-brainers to go in as such. The temporary sysfs interface is also fine since we'll need to change sysfs eventually to interfaces anyway so we can mock it easily in unit tests.

I let the others chime in on the interface vs. member function pointers + closure for polymorphism. Either way is fine with me as long as there is some degree of consensus about it. I personally tend to use member function pointers only for optional functionality. But I have neither a strong opinion nor enough golang written under my belt that I could afford to be too strong-minded about it.

rojkov:
Sure, I'll split to smaller PRs to digest them piecemeal.

rojkov:
I've cleaned the PR up.

The biggest logical changes this PR brings in are that `Node.DiscoverCPU()` and `Node.DiscoverMemset()` were reduced to `node.discoverCPU()` and `node.getMemset()` respectively. Both old methods were called for every node in the tree yet they themselves traverse their subtrees.

The new code does traversing the tree only once for CPUs starting from the root. And a similar change has been done for Memset (plus some added laziness to its evaluation).

This makes me thing that either the logic of the old code or mine is wrong. I'm not sure mine is correct. Would be nice to add a couple of test cases to `TestDiscoverCPU()` and `TestGetMemset()` illustrating the expected results for given inputs.

klihub:
> The new code does traversing the tree only once for CPUs starting from the root. And a similar change has been done for Memset (plus some added laziness to its evaluation).
> 
> This makes me thing that either the logic of the old code or mine is wrong. I'm not sure mine is correct. Would be nice to add a couple of test cases to `TestDiscoverCPU()` and `TestGetMemset()` illustrating the expected results for given inputs.

I need to reread the original code to recall the details but my vague recollection is that
  - CPU discovery discovered only the CPU directly attached to a tree node, however
  - free CPU calculation was recursively cumulative traversing the subtree rooted at the node

klihub:
> * CPU discovery discovered only the CPU directly attached to a tree node, however
> * free CPU calculation was recursively cumulative, traversing the sub-tree rooted at the node

Sorry, I was wrong. CPU discovery is not calculating the directly attached CPU. It is calculating the cumulative CPU available at the subtree below the node. It is just implemented stupidly and does several traversals of the tree.

pkg/sysfs/system.go
```diff
@@ -70,7 +70,7 @@ type System struct {
 // Package is a physical package (a collection of CPUs).
 type Package struct {
 	id    ID    // package id
-	cpus  IDSet // CPUs in this package
+	Cpus  IDSet // CPUs in this package
```
marquiz: Why is this made a publid member? Either all of them should be public or none?

rojkov: I can't recall why already. I suspect the reason not to make all public is to reduce the patch size.

marquiz: But, do you need to make this public? I don't think you do

rojkov: This needs to be public to set up unit tests. Otherwise there's no way to create a policy with a mocked system in https://github.com/intel/cri-resource-manager/pull/90/files#diff-3e20512209e1f47a7250162092dbd6b5R47.

ipuustin: I have another attempt at refactoring this file coming up soon, stay tuned. Maybe this work can be rebased on top of it.

klihub:
Putting this on hold until the dust settles a bit over the memtier policy which was forked from topology-aware.

PR #92, opened 2019-12-10
topologyaware: cover pod-preferences.go with unit tests

rojkov:
The tests cover 100% of pod-preferences.go. Hopefully the input and expected output make sense as I didn't fully understand the code's purpose. E.g. what is `elevate` in function results?

Also please check if the names of the test cases reflect their meaning.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
@@ -15,8 +15,9 @@
 package topologyaware
 
 import (
-	"github.com/ghodss/yaml"
 	"strconv"
+
+	"github.com/ghodss/yaml"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
```diff
@@ -59,6 +62,8 @@ func podIsolationPreference(pod cache.Pod, container cache.Container) (bool, boo
 }
 
 // podSharedCPUPreference checks if a container wants to opt-out from exclusive allocation.
+// The first return value indicates the container is ok to share resources (true).
+// FIXME(rojkov): what is elevate?
```
klihub: I think an accurate description should be: The first return value indicates if the container prefers to opt-out from exclusive (sliced-off or isolated) CPU allocation even if it was otherwise eligible for it.

The second return value, elevate, indicates how much to elevate the actual allocation of the container in the tree of pools. IOW, how many levles to go up in the tree starting at the best fitting pool, before assigning the container to an actual pool. That preference is currently ignored.

rojkov: Updated.

PR #93, opened 2019-12-11
rdt: simplify l3 configuration

marquiz:
This patch tries to simplify the configuration of the RDT L3 allocation
and aims to make it slightly more comprehensible and user friendly.
Instead of having independent configuration sections (l3Schema,
l3CodeSchema, l3DataSchema), we now have separate 'unified', 'code' and
'data' settings under the 'l3Schema' key. E.g.
```
    resctrlGroups:
      Guaranteed:
        l3Schema:
          all:
            unified: 90%
            code: 100%
            data: 80%
```
Moreover, 'unified' must always be specified, and, 'code' and 'data' go
together so that none or both of them must be specified (only specifying
one of them creates an error). These requirements were added in order to
reduce the risk of accidental misconfiguration.

Also. the fallback/failure logic has been simplified. Now the
configuration only fails if 'l3Schema' is specified but L3 CAT is not
available and the 'rdt.config.options.l3.optional' is set to false. In
addition the 'l3' setting (which is now mandatory) is used as a
default/fallback value for 'code' and 'data' settings.

### DIFF
--- a/pkg/rdt/bitmask.go
+++ b/pkg/rdt/bitmask.go
@@ -17,6 +17,7 @@ limitations under the License.
 package rdt
 
 import (
+	"fmt"
 	"math/bits"
 	"strconv"
 	"strings"
@@ -30,6 +31,11 @@ import (
 // interface. The "width" i.e. the number of bits available depends on the
 // underlying hardware.
 type CacheBitmask Bitmask
+
+// MarshalJSON is the marshaller function for "encoding/json"
+func (b CacheBitmask) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("\"%#x\"", b)), nil
+}
 
 // UnmarshalJSON is the unmarshaller function for "encoding/json"
 func (b *CacheBitmask) UnmarshalJSON(data []byte) error {
--- a/pkg/rdt/config.go
+++ b/pkg/rdt/config.go
@@ -29,33 +29,29 @@ import (
 // ResctrlGroupConfig represents configuration of one CTRL group in the Linux
 // resctrl interface
 type ResctrlGroupConfig struct {
-	L3Schema     L3Schema `json:"l3Schema,omitempty"`
-	L3CodeSchema L3Schema `json:"l3CodeSchema,omitempty"`
-	L3DataSchema L3Schema `json:"l3DataSchema,omitempty"`
-	MBSchema     MBSchema `json:"mbSchema,omitempty"`
+	L3Schema L3Schema `json:"l3Schema,omitempty"`
+	MBSchema MBSchema `json:"mbSchema,omitempty"`
 }
 
 // SchemaOptions contains the common settings for all resctrl groups
 type SchemaOptions struct {
-	L3     L3Options `json:"l3,omitempty"`
-	L3Code L3Options `json:"l3code,omitempty"`
-	L3Data L3Options `json:"l3data,omitempty"`
-	MB     MBOptions `json:"mb,omitempty"`
+	L3 L3Options `json:"l3"`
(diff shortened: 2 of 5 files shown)


marquiz:
Please take a look and provide feedback if this makes sense. One thing I was pondering is the naming of the `l3` key in the config, i.e. should we use something else like `unified`, instead

marquiz:
Any thoughts? Another way to change the setting (to avoid the individual l3/code/cache keys) would be to put all settings into a single line, e.g. like this:
```
    resctrlGroups:
      Guaranteed:
        l3Schema:
          all: "100%,code:90%,data:80%"
```

klihub:
> Please take a look and provide feedback if this makes sense. One thing I was pondering is the naming of the `l3` key in the config, i.e. should we use something else like `unified`, instead

To me it would sound more intuitive to change, as you suggest, `l3` to `unified`.
Moreover, I think it would be clearer to allow only two alternatives for the `unified/code/data` combo:
  - specify *all three* of them, or
  - specify *only `unified`*

IOW, don't allow a partial fine-grained configuration where `l3`/`unified` is given together with either `data` or `code` with `l3`/`unified` providing a fallback for the missing one.

klihub:
> Any thoughts? Another way to change the setting (to avoid the individual l3/code/cache keys) would be to put all settings into a single line, e.g. like this:
> 
> ```
>     resctrlGroups:
>       Guaranteed:
>         l3Schema:
>           all: "100%,code:90%,data:80%"
> ```

If we didn't use JSON/YAML for the configuration, I'd prefer this one. But I think cloud-native folks would consider this violently non-yamlish.

BTW, we have something like this for the log configuration which we should probably change to a more yamlish style.

marquiz:
> To me it would sound more intuitive to change, as you suggest, `l3` to `unified`.

Yes I agree, `unified` is probably more comprehensible and logical. I'll change this.

> I think it would be clearer to allow only two alternatives for the `unified/code/data` combo:
> 
> * specify _all three_ of them, or
> * specify _only `unified`_

Again, I agree. Makes sense and decreases the possibility of accidental misconfiguration. I'll change this, too

> If we didn't use JSON/YAML for the configuration, I'd prefer this one. But I think cloud-native folks would consider this violently non-yamlish.

Yes, stupid idea 😄 Plus this would just make our parsing (unmarshal) code slightly more complex.

marquiz:
Pushed a new version.
- `l3` under (`l3Schema`) renamed to `unified`
- both `code` and `data` must be specified together (or none of them)

marquiz:
Woops, forgot to update the sample config in previous version :smirk:  Now fixed

marquiz:
rebased

klihub:
@marquiz Could you add also a patch addressing these few remaining golint complaints:
```
./pkg/rdt/info.go:28:6: exported type Info should have comment or be unexported
Found 1 lint suggestions; failing.
./pkg/rdt/config.go:46:1: comment on exported type L3Options should be of the form "L3Options ..." (with optional leading article)
./pkg/rdt/config.go:51:1: comment on exported type MBOptions should be of the form "MBOptions ..." (with optional leading article)
./pkg/rdt/config.go:66:6: exported type L3SchemaType should have comment or be unexported
./pkg/rdt/config.go:69:2: exported const L3SchemaTypeUnified should have comment (or a comment on this block) or be unexported
./pkg/rdt/config.go:114:1: exported method L3Schema.UnmarshalJSON should have comment or be unexported
./pkg/rdt/config.go:190:1: exported method MBSchema.UnmarshalJSON should have comment or be unexported
```

marquiz:
> @marquiz Could you add also a patch addressing these few remaining golint complainta

Fixed

PR #98, opened 2019-12-18
introduce end to end tests

rojkov:
This is preliminary e2e tests.

All test cases run in isolation: for each case a brand new resource manager is created and started together with a fake CRI gRPC server.

I'm going to simplify the code structure to make it more readable, hence WIP.

### DIFF
--- a/test/functional/e2e_test.go
+++ b/test/functional/e2e_test.go
@@ -1,0 +1,179 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


klihub:
This looks interesting !

rojkov:
I'm kinda satisfied now with the code. The PR is ready for review.

test/functional/e2e_test.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+const (
+	testDir = "/tmp/cri-rm-test"
+)
+
+func setUp(t *testing.T) {
+	t.Helper()
+
+	if err := os.MkdirAll(testDir, 0700); err != nil {
```
marquiz: Should we use `ioutil.TempDir` or smth to avoid possible races?

rojkov: Agree. Even though mkdir is an atomic syscall there's no need to call it for every test function. I put it to `init()`.

PR #101, opened 2019-12-19
[RFC] resource manager: add event and metrics processing skeleton.

klihub:
This PR is based on/requires #84 and #85 .

This patch series adds skeletal support for making resource manager policy decisions based on unsolicited *events*, in addition to the current, *request*-based decisions. The various patches here
  - splits out and makes externally accessible the HTTP server in pkg/instrumentation
  - adds support for dynamically registered prometheus.Gatherer's to instrumentation (for proxying)
  - add a NUMA- block I/O, etc. metrics collector
  - add policy support for reallocating resources of a subset of the running containers
  - add support for event-triggered decision making
  - periodically trigger metrics collection
  - proxies the collected raw metrics to prometheus
  - generate events and related decisions based on the metrics
  - adds support for AVX512-related container affinity

Currently the only metric we generate events for is AVX512 instructions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -353,6 +353,8 @@ bin/webhook: $(wildcard cmd/webhook/*.go) \
                 find $$dir -name \*.go; \
             done | sort | uniq)
 
+bin/avx512-load: $(wildcard cmd/avx512-load/*.go)
+
 # phony targets
 .PHONY: all build install clean test images \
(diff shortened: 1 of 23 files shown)


klihub:
Closing. We've merged most of it on the [rebalancing]-series, with #151 pending for review and one more (proxying raw metrics to Prometheus) later.

PR #104, opened 2019-12-23
Makefile: always run 'go test' verbosely.

klihub:


klihub:
This is a bad idea. It can change timing (does in my tests) and hide data races that would otherwise get detected. Closing it.

PR #106, opened 2019-12-23
avx: disable AVX collector for tests.

klihub:
It is normal for a *build system* to *not have* the AVX collector's ELF binary for eBPF filtering installed. Currently,
 - a failure to load the ELF binary causes the AVX collector's constructor to fail,
 - which in turn causes the metrics collector's constructor to bail out with an error,
 - which then causes the resource manager's constructor to fail

This prevents even the trivial end-to-end tests from running.

As a workaround split out the code responsible for automatically registering the AVX collector to a file of its own with a compilation constraint on (the lack of) a tag. When running/building tests, set up the default build tags so that the AVX registration code does not get compiled into the test binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,10 @@ GO_CYCLO  := gocyclo
 GO_LINT   := golint
 GO_CILINT := golangci-lint
 
+# TEST_TAGS is the set of extra build tags passed for tests.
+TEST_TAGS :=
+GO_TEST   := $(GO_CMD) test $(TEST_TAGS)
+
 # Disable some golangci_lint checkers for now until we have an more acceptable baseline...
 GO_CILINT_CHECKERS := -D unused,staticcheck,errcheck,deadcode,structcheck,gosimple -E golint,gofmt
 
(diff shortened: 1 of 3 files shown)


mythi:
> We'll probably need a command line or configuration option as well for disabling the AVX collector altogether. 

Would a build (`// +build !foo`) constraint in `register_metrics.go` and the corresponding `go -tags foo test` in `Makefile` work?

klihub:
> > We'll probably need a command line or configuration option as well for disabling the AVX collector altogether.
> 
> Would a build (`// +build !foo`) constraint in `register_metrics.go` and the corresponding `go -tags foo test` in `Makefile` work?

Yes, that'd probably be a better idea.

klihub:
> Would a build (`// +build !foo`) constraint in `register_metrics.go` and the corresponding `go -tags foo test` in `Makefile` work?

I'll take a look at it if I could change this to use tags and build constraints. Not sure if it is doable in a straightforward manner (when you want to exclude code from a dependency and not from the package being built for testing). Other folks seem to struggle with this too...
https://github.com/golang/go/issues/12120
https://github.com/golang/go/issues/21360

mythi:
Hmm, I see. I found those issues too but they seem to be asking an *automatic* constraint during `go test`...

klihub:
True, it works fine with a manually passed tag dedicated for testing. Updated PR accordingly.

Makefile
```diff
@@ -201,11 +201,11 @@ golangci-lint:
 
 test:
 ifndef WHAT
-	$(Q)$(GO_CMD) test -race -coverprofile=coverage.txt -covermode=atomic \
+	$(Q)$(GO_CMD) test -tags testing -race -coverprofile=coverage.txt -covermode=atomic \
```
mythi: this could be an override that travis/any CI sets if it needs to skip avx but is run by default (or vice versa)?

klihub: Ok. Changed it into an overrideable set of build tags specific to building tests. By default those tags are set to disable the registration of the AVX collector.

PR #107, opened 2020-01-03
Fix: add missing docstrings for exported symbols.

klihub:
Add a few missing docstrings for exported types, functions in pkg/cgroups and pkg/metrics.

### DIFF
--- a/pkg/cgroups/cgroupid.go
+++ b/pkg/cgroups/cgroupid.go
@@ -10,12 +10,14 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// CgroupID implements mapping kernel cgroup IDs to cgroupfs paths with transparent caching.
 type CgroupID struct {
 	root  string
 	cache map[uint64]string
 	sync.Mutex
 }
 
+// NewCgroupID creates a new CgroupID map/cache.
 func NewCgroupID(root string) *CgroupID {
 	return &CgroupID{
 		root:  root,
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -12,8 +12,10 @@ var (
 	log                  = logger.NewLogger("metrics")
 )
 
+// InitCollector is the type for functions that initialize collectors.
 type InitCollector func() (prometheus.Collector, error)
 
+// RegisterCollector registers the names prometheus.Collector for metrics collection.
```
marquiz: nit: maybe it's just me but I don't understand the meaning of *the names prometheus.Collector* 😄

pkg/cgroups/cgroupid.go
```diff
@@ -10,12 +10,14 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// CgroupID is a cgroup id to path map/cache.
```
marquiz: I don't fully understand this either 😉

pkg/metrics/metrics.go
klihub: Argh... should have been *named*. Fixed.

pkg/cgroups/cgroupid.go
klihub: Hmm... fair enough. Let's try to clarify it then.

PR #109, opened 2020-01-09
rdt: extra hack for getting rdt properly configured and started

marquiz:
Fix segfault when NODE_NAME is unset. Drop some commented out code


klihub:
BTW, the description is incorrect, copied from a previous PR.

PR #112, opened 2020-01-14
instrumentation: expose HTTP server, support dynamic reconfiguration.

klihub:
This patch series reworks instrumentation by
  - making the HTTP server accessible/usable for other internal packages
  - implementing dynamic (re)configurability
  - adding support for dynamically registered gatherers (which we'll use for proxying raw metrics to Prometheus)


pkg/instrumentation/flags.go
```diff
@@ -148,8 +155,18 @@ func defaultOptions() interface{} {
 
 // configNotify is our configuration udpate notification handler.
 func configNotify(event config.Event, source config.Source) error {
-	log.Info("tracing configuration is now %v", opt.Trace)
-	Setup()
+	log.Info("instrumentation configuration is now %v", opt.Trace)
+
+	if opt.Collector == cfg.Collector || opt.Agent == cfg.Agent || opt.Metrics == cfg.Metrics {
```
marquiz: Should it be `&&` instead `||`?

pkg/instrumentation/flags.go
```diff
@@ -148,8 +155,18 @@ func defaultOptions() interface{} {
...
+		cfg.Trace = opt.Trace
+		ConfigureTracing(opt.Trace)
+	} else {
+		log.Info("some of the tracing enpoints have changes, trying to restart...")
```
marquiz: s/changes/changed/ ?

pkg/instrumentation/flags.go
klihub: Yes, it should test if any of the endpoints have changed, so it's broken now.

PR #126, opened 2020-01-14
resource-manager: default to connecting to containerd instead of dockershim.

klihub:
Try connecting by default to the containerd socket, both for the runtime and the image services.


kad:
lgtm, with small question: do we still use anywhere constant DockerShim ? e.g. in some help output?

klihub:
> lgtm, with small question: do we still use anywhere constant DockerShim ? e.g. in some help output?

Ah, good point. No, so we should remove the constant.

PR #131, opened 2020-01-16
[RFC] introspection: provide data for external introspection (for instance, visualization)

klihub:
This patch series adds initial support for external introspection of the current state of the resource manager. Introspection data is served over the instrumentation metrics HTTP endpoint using the '/introspect' path. The data is JSON and currently contains:
  - Pods (name, ID, containers)
  - Containers (name, IDs, command/args, resource requests, topology hints)
  - Pools (name, CPUs, memory, parent/children pools)
  - Assignments (container ID, pool, CPU, memory, RDT class, Block I/O class)
  - System info (HW: sockets, NUMA nodes, isolated&offline CPUs, RDT classes, active policy)

In it's current incarnation the patch set fetches pools and assignments from the active policy while other data is provided by the common agnostic policy layer. Consequently, new corresponding Introspect() functions have been added to the policy and backend interfaces as well as to the builtin policy implementations (only the topology-aware one provides real data FTTB).

The reason for this split of duty is that we have not standardized/unified the data structure(s) for resource assignment in the cache. Therefore it is now stored among the policy-specific data. Consequently the agnostic policy layer cannot interpret/map it to the data format provided externally by/for introspection. If/Once we do the unification, we'll probably be able to get rid of the backend-specific Introspect() interface altogether.

The external introspection interface currently only supports fetching a full snapshot of the current state without providing any means to easily track changes. This is true both for the internal interface which provides data for introspection and the external HTTP introspection interface. Some of the existing data structures (introspect.Update) and functions (introspect.Server.Update) are there with differential updates in mind, although they are currently unused or just stubs.

For this patch set I'd like to do/experiment with these additional things:
  - unify representation of container resource assignment
  - move storing resource assignment from the various policies to the generic policy layer
  - differential updates/provide extra data for indicating the changes since the last introspection

The first two we should do regardless of this patch set not because of it...

### DIFF
--- a/pkg/cri/resource-manager/introspect/introspect.go
+++ b/pkg/cri/resource-manager/introspect/introspect.go
@@ -1,0 +1,201 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package introspect
+
+import (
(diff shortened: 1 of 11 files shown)


mythi:
target to add "self-hosting visualization/UI" before final version

klihub:
Closing, superseded by #184 .

PR #135, opened 2020-01-17
rdt: support MBps mode of memory bandwidth allocation

marquiz:
Changes the configuration so that the MB allocation (rdt partitions) and
MB schema (rdt classes) values are now a specified as a list instead of
a single value. Also, a unit is now required in order to determine what
the user intended (100% is quite different than 100MBps in modern
systems). The value to use is picked depending on the mode active on the
os level (whether -o mba_MBps mount option was used or not).

A missing value (of the correct unit) is treated as a configuration
error. This means that normally all mb schemas must have a percentage
value, and, if 'o -mba_MBps' mount option of the resctrl fileystem, all
mb schemas must have a MBps value.

A configuration example:
```
  rdt:
    partitions:
      default:
        mbAllocation:
          all: [90% 9000MBps]
        classes:
          Burstable:
            mbSchema:
              all: [66%, 5000MBps]
```


marquiz:
Set to RFC status. Depends on #134

marquiz:
Changed the fallback logic so that a missing value of the correct unit is treated as a configuration error. Trying to avoid unintentional/invisible configuration mistakes with this change

marquiz:
Removed RFC status

PR #137, opened 2020-01-20
[rebalancing] cgroupstats: add cgroupfs based metrics collector.

klihub:
Add initial implementation of a cgroupfs based metrics collector, courtesy of Ismo.

### DIFF
--- a/pkg/cgroups/cgroupstats.go
+++ b/pkg/cgroups/cgroupstats.go
@@ -1,0 +1,500 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cgroups
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/intel/cri-resource-manager/pkg/sysfs"
+)
+
+// BlkioDeviceBytes contains a single operations line of blkio.throttle.io_service_bytes_recursive file
+type BlkioDeviceBytes struct {
+	Major      int
+	Minor      int
+	Operations map[string]int64
+}
+
+// BlkioThrottleBytes has parsed contents of blkio.throttle.io_service_bytes_recursive file
+type BlkioThrottleBytes struct {
+	DeviceBytes []*BlkioDeviceBytes
+	TotalBytes  int64
+}
+
+// CPUAcctUsage has a parsed line of cpuacct.usage_all file
+type CPUAcctUsage struct {
+	CPU    int
+	User   int64
+	System int64
+}
+
+// HugetlbUsage has parsed contents of huge pages usage in bytes.
+type HugetlbUsage struct {
+	Size     string
+	Bytes    int64
+	MaxBytes int64
+}
+
+// MemoryUsage has parsed contents of memory usage in bytes.
+type MemoryUsage struct {
+	Bytes    int64
+	MaxBytes int64
+}
+
+// NumaLine represents one line in the NUMA statistics file.
+type NumaLine struct {
+	Total int64
+	Nodes map[string]int64
+}
+
+// NumaStat has parsed contets of a NUMA statistics file.
+type NumaStat struct {
+	Total       NumaLine
+	File        NumaLine
+	Anon        NumaLine
+	Unevictable NumaLine
+
+	HierarchicalTotal       NumaLine
+	HierarchicalFile        NumaLine
+	HierarchicalAnon        NumaLine
+	HierarchicalUnevictable NumaLine
+}
+
+// GlobalNumaStats has the statistics from one global NUMA nodestats file.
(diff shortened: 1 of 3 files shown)


ipuustin:
I think this looks good. It only works on systemd cgroup driver though. Is that a problem?

klihub:
> I think this looks good. It only works on systemd cgroup driver though. Is that a problem?

There are a number of other place in the code that probably break down if kubelet is running with the cgroupfs driver. I think we should address all of those at the same time, in a single sweep and my hope was to take care of those in a later PR. That said if folks tend to lean towards doing it now, we can do that as well. In any case, I'll update this PR's commit adding a note+TODO about it...

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,541 @@
...
+
+	lines, err := readCgroupFileLines(path.Join(cgroupPath, "cpuacct.usage_all"))
+
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]CPUAcctUsage, len(lines)-1)
+
+	for i, line := range lines {
```
marquiz: just a suggestion:
```
for i, line := range lines[1:]
```
would make the logic in the loop slightly simpler

klihub: Updated. Also further simplified the loop by doing filtering with continue's instead of nested if's. 
Also fixed the combo bug of incorrectly sizing result + the loop index i not reflecting skipped lines which could cause zero-valued 'holes' or trailing items in result if lines were really skipped.

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,542 @@
...
+
+	lines, err := readCgroupFileLines(path.Join(cgroupPath, "memory.numa_stat"))
+
+	if err != nil {
+		return NumaStat{}, err
+	}
+
+	result := NumaStat{}
+	for _, line := range lines {
+		tokens := numaLineRe.FindStringSubmatch(line)
```
marquiz: I'd prolly not use regexps here, but, just go just with `strings.Split()` and a helper function for parsing `key=value` type tokens. But I guess this is fine if it works here :smile:

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,542 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]CPUAcctUsage, 0, len(lines)-1)
+
+	for _, line := range lines[1:] {
+		if len(line) == 0 {
+			continue
+		}
```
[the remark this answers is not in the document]
klihub: True, updated.

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,542 @@
...
+	//
+	// 124
+
+	usageFiles, err := filepath.Glob(path.Join(cgroupPath, "/hugetlb.*.usage_in_bytes"))
+
+	if err != nil {
+		return nil, err
+	}
+
+	maxUsageFiles, err := filepath.Glob(path.Join(cgroupPath, "/hugetlb.*.max_usage_in_bytes"))
```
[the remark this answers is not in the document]
klihub: Hmm... good point. Now that you drew my attention to it, I'll try to rework this eliminating all regexp usage from this function. It looks simple enough to do...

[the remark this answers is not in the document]
klihub: I reworked all code using rather expensive regexp matching for fairly trivial things, without using any regexp.

pkg/cgroups/cgroupstats.go
klihub: Removed regexp usage.

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,542 @@
...
+	// numa_hit 1851614569
+	// numa_miss 0
+	// numa_foreign 0
+	// interleave_hit 49101
+	// local_node 1851614569
+	// other_node 0
+
+	result := make(map[int]GlobalNumaStats)
+
+	nodeDirs, err := filepath.Glob("/sys/devices/system/node/node*")
```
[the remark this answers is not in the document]
klihub: Removed regexp usage.

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,499 @@
...
+				return BlkioThrottleBytes{}, err
+			}
+			min64, err := strconv.ParseInt(string(majmin[1]), 10, 32)
+			if err != nil {
+				return BlkioThrottleBytes{}, err
+			}
+			major := int(maj64)
+			minor := int(min64)
+
+			idx, ok := devidx[split[0]]
```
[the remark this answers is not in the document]
klihub: Yup, golangci-lint spotted it as well. Fixed.

pkg/cgroups/cgroupstats.go
```diff
@@ -0,0 +1,500 @@
...
+	// numa_hit 1851614569
+	// numa_miss 0
+	// numa_foreign 0
+	// interleave_hit 49101
+	// local_node 1851614569
+	// other_node 0
+
+	result := make(map[int]GlobalNumaStats)
+
+	nodeDirs, err := filepath.Glob(prefix + "*")
```
marquiz: My previous comment to this was a bit vague. What I tried to say (in addition to removing the regexp) is that you could get rid of the glob too by doing ReadDir in `/sys/bus/node/devices/`. This dir does not contain any additional files, just symlinks to the nodes)

klihub: No, I noticed the comment earlier but I ignored it deliberately. It's true that it probably would be faster to do it the way you suggest (kernel do an extra symlink dereference vs. golang implementation of glob). Our other parts of the code does the same thing as here, so I left it untouched for consistency. We can change that later for all of our code.

PR #138, opened 2020-01-21
[rebalancing] cache: add cache function for looking up containers by cgroup path.

klihub:
Add LookupContainerByCgroup(), a function for finding a container based on a cgroupfs path.
We need this for associating collected cgroupfs based metrics with actual running containers.

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -421,6 +421,8 @@ type Cache interface {
 	DeleteContainer(id string) Container
 	// LookupContainer looks up a container in the cache.
 	LookupContainer(id string) (Container, bool)
+	// LookupContainerByCgroup looks up a container for the given cgroup path.
+	LookupContainerByCgroup(path string) (Container, bool)
 
 	// GetPendingContainers returs all containers with pending changes.
 	GetPendingContainers() []Container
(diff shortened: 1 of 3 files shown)


mythi:
TODO: add a test case

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -694,6 +696,35 @@ func (cch *cache) LookupContainer(id string) (Container, bool) {
...
 }
 
+// LookupContainerByCgroup looks up the container for the given cgroup path.
+func (cch *cache) LookupContainerByCgroup(path string) (Container, bool) {
+	cch.Debug("resolving %s to a container...", path)
+
+	for id, c := range cch.Containers {
+		if id != c.CacheID {
+			continue
+		}
```
marquiz: When do we hit this, again? Maybe a short comment to clarify(?)

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -694,6 +696,35 @@ func (cch *cache) LookupContainer(id string) (Container, bool) {
...
+		if !strings.HasPrefix(path, parent) {
+			continue
+		}
+
+		if strings.Index(path, c.GetID()) != -1 {
+			return c, true
+		}
+	}
+
+	return nil, false
```
marquiz: Would `nil, true` be valid in some case? Could we just return `Container`?

pkg/cri/resource-manager/cache/cache.go
klihub: We hash containers by two IDs: the ID in the CRI runtime and our cache ID. We hit this for every container which has already been successfully created in the actual runtime. So it is rather the rule, not the exception.

marquiz: Ah yeah, of course

pkg/cri/resource-manager/cache/cache.go
klihub: > Would `nil, true` be valid in some case? Could we just return `Container`?

I think this is the classical go nil interface problem. If we returned (nil, true) here, how could you tell a valid Container apart from a nil Container interface without using reflect ?

marquiz: > I think this is the classical go nil interface problem. If we returned (nil, true) here, how could you tell a valid Container apart from a nil Container interface without using reflect ?

Yeah, that's true. What I was thinking was that is it a "legitimate" for these functions to return a `nil` Container interface in any condition?

klihub:
Added minimal tests for cache.LookupContainerByCgroup().

klihub:
> TODO: add a test case

Added (a minimal) test case.

PR #139, opened 2020-01-21
[rebalancing] cache: add cache functions for tagging containers.

klihub:
Add functions for manipulating container tags, which are basically dynamic labels which exist only in the scope of the resource manager itself. We'll use tags to attach dynamically changing attributes to running containers based on the collected metrics data.

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -323,6 +323,13 @@ type Container interface {
 	HasPending(string) bool
 	// ClearPending clears the pending change marker for the given controller.
 	ClearPending(string)
+
+	// GetTag gets the value of the given tag.
+	GetTag(string) (string, bool)
+	// SetTag sets the value of the given tag and returns its previous value..
+	SetTag(string, string) (string, bool)
+	// DeleteTag deletes the given tag, returning its deleted value.
+	DeleteTag(string) (string, bool)
 }
 
 // A cached container.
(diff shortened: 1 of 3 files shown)


pkg/cri/resource-manager/cache/cache.go
```diff
@@ -344,6 +351,7 @@ type container struct {
 	Mounts        map[string]*Mount   // mounts
 	Devices       map[string]*Device  // devices
 	TopologyHints sysfs.TopologyHints // Set of topology hints for all containers within Pod
+	Tags          map[string]string   // container tags (local dyanmic labels)
```
marquiz: s/dyanmic/dynamic/

pkg/cri/resource-manager/cache/container.go
```diff
@@ -790,3 +793,20 @@ func (c *container) HasPending(controller string) bool {
...
+}
+
+func (c *container) SetTag(key string, value string) (string, bool) {
+	prev, ok := c.Tags[key]
+	c.Tags[key] = value
+	return prev, ok
+}
+
+func (c *container) DeleteTag(key string) (string, bool) {
+	prev, ok := c.Tags[key]
```
marquiz: nit: `prev` is a bit strange name in this context :smile:

klihub: True, I'll change it to `value`...

pkg/cri/resource-manager/cache/cache.go
klihub: Thanks, fixed.

PR #143, opened 2020-01-21
[rebalancing] cache: allow affinity expressions to reference other metadata than just labels

klihub:
Allow affinity expressions to reference any of the following metadata:
  - name
  - namespace
  - QoS class
  - labels
  - tags
  - any of the above (sans tags) of the containers parent pod

This mechanism enables the use more flexible and fine-grained affinities. For instance, once can now create an affinity with a scope of all containers of Guaranteed QoS class pods.

### DIFF
--- a/docs/container-affinity.md
+++ b/docs/container-affinity.md
@@ -1,0 +1,224 @@
+# Container Affinity and Anti-Affinity
+
+## Introduction
+
+Some policies allow the user to give hints about how particular containers
+should be *co-located* within a node. In particular these hints express whether
+containers should be located *'close'* to each other or *'far away'* from each
+other, in a hardware topology sense.
+
+Since these hints are interpreted always by a particular *policy implementation*,
+the exact definitions of 'close' and 'far' are also somewhat *policy-specific*.
+However as a general rule of thumb containers running
+
+  - on CPUs within the *same NUMA nodes* are considered *'close'* to each other,
+  - on CPUs within *different NUMA nodes* in the *same socket* are *'farther'*, and
+  - on CPUs within *different sockets* are *'far'* from each other
+
+These hints are expressed by `container affinity annotations` on the Pod.
+There are two types of affinities:
+
+  - `affinity` (or `positive affinty`): cause affected containers to *pull* each other closer
+  - `anti-affinity` (or `negative affinity`): cause affected containers to *push* each other further away
+
+Policies try to place a container
+  - close to those the container has affinity towards
+  - far from those the container has anti-affinity towards.
+
+## Affinity Annotation Syntax
+
+*Affinities* are defined as the `cri-resource-manager.intel.com/affinity` annotation.
+*Anti-affinities* are defined as the `cri-resource-manager.intel.com/anti-affinity`
+annotation. They are specified in the `metadata` section of the `Pod YAML`, under
+`annotations` as a dictionary, with each dictionary key being the name of the
+*container* within the Pod to which the annotation belongs to.
+
+```yaml
+metadata:
+  anotations:
+    cri-resource-manager.intel.com/affinity: |
+      container1:
+        - scope:
+            key: [optional-key-domain/]key
+            operator: op
+            values:
+            - value1
+            ...
+            - valueN
+        - match:
+            key: [optional-key-domain/]key
+            operator: op
+            values:
+            - value1
+            ...
+            - valueN
+          weight: w
+```
+
+An anti-affinity is defined similarly but using `cri-resource-manager.intel.com/anti-affinity`
+as the annotation key.
+
+```yaml
+metadata:
+  anotations:
+    cri-resource-manager.intel.com/anti-affinity: |
+      container1:
+        - scope:
+            key: [optional-key-domain/]key
+            operator: op
+            values:
+            - value1
+            ...
+            - valueN
+        - match:
+            key: [optional-key-domain/]key
+            operator: op
+            values:
+            - value1
+            ...
+            - valueN
+          weight: w
+```
+
+## Affinity Semantics
+
+An affinity consists of three parts:
+
+  - `scope expression`: defines which containers this affinity is evaluated against
+  - `match expression`: defines for which containers (within the scope) the affinity applies to
+  - `weight`: defines how *strong* a pull or a push the affinity causes
+
+*Affinities* are also sometimes referred to as *positive affinities* while
+*anti-affinities* are referred to as *negative affinities*. The reason for this is
+that the only difference between these are that affinities have a *positive weight*
+while anti-affinities have a *negative weight*.
+
+The *scope* of an affinity defines the *bounding set of containers* the affinity can
+apply to. The affinity *expression* is evaluated against the containers *in scope* and
+it *selects the containers* the affinity really has an effect on. The *weight* specifies
+whether the effect is a *pull* or a *push*. *Positive* weights cause a *pull* while
+*negative* weights cause a *push*. Additionally, the *weight* specifies *how strong* the
+push or the pull is. This is useful in situations where the policy needs to make some
+compromises because an optimal placement is not possible. The weight then also acts as
+a way to specify preferences of priorities between the various compromises: the heavier
+the weight the stronger the pull or push and the larger the propbability that it will be
+honored, if this is possible at all.
+
+The scope can be omitted from an affinity in which case it implies *Pod scope*, in other
+words the scope of all containers that belong to the same Pod as the container for which
+which the affinity is defined.
+
+The weight can also be omitted in which case it defaults to -1 for anti-affinities
+and +1 for affinities.
+
+Both the affinity scope and the expression select containers, therefore they are identical.
+Both of them are *expressions*. An expression consists of three parts:
+
+  - key: specifies what *metadata* to pick from a container for evaluation
+  - operation (op): specifies what *logical operation* the expression evaluates
+  - values: a set of *strings* to evaluate the the value of the key against
+
+Essentially an expression defines a logical operation of the form (key op values).
+Evaluating this logical expression will take the value of the key in  which
+either evaluates to true or false. 
+a boolean true/false result. Currently the following operations are supported:
+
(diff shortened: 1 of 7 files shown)


mythi:
TODO: document the feature

pkg/cri/resource-manager/cache/pod.go
```diff
@@ -109,6 +111,32 @@ func (p *pod) GetContainers() []Container {
...
+	}
+
+	return found
+}
+
+// Get container by its name.
+func (p *pod) GetContainer(name string) (Container, bool) {
+	c := p.getContainer(name)
+
+	return c, c != nil
```
marquiz: Again, was there a reason that we want to return two values?

pkg/cri/resource-manager/cache/affinity.go
```diff
@@ -178,11 +164,88 @@ func (e *Expression) Evaluate(container Container) bool {
...
+
+	obj = c
+	ref := strings.Split(path, "/")
+	if len(ref) == 1 {
+		ref = []string{"labels", path}
+	}
+	for {
+		key := ref[0]
+		c.cache.Debug("* resolve: walking %s, @%s, obj %T...", path, key, obj)
+		switch obj.(type) {
```
marquiz: Could you use something like `switch v := obj.(type)` here? You should be able to avoid the type assertions below with that.

pkg/cri/resource-manager/cache/affinity.go
```diff
@@ -178,11 +164,88 @@ func (e *Expression) Evaluate(container Container) bool {
...
+		case *container:
+			c := obj.(*container)
+			switch key {
```
marquiz: Should/could we just ignore case here by using strings.ToLower as we're doing some ignoring (ignorance(?) :wink: ) below, anyway?

pkg/cri/resource-manager/cache/affinity.go
```diff
@@ -178,11 +164,88 @@ func (e *Expression) Evaluate(container Container) bool {
...
+	var obj interface{}
+
+	c.cache.Debug("resolving %s/%s...", c.PrettyName(), path)
+
+	obj = c
+	ref := strings.Split(path, "/")
+	if len(ref) == 1 {
+		ref = []string{"labels", path}
+	}
+	for {
```
marquiz: Would `for len(ref) > 0` work?

pkg/cri/resource-manager/cache/pod.go
klihub: We want to return one interface value that we can tell apart from a nil Container without using reflect... ;-)

pkg/cri/resource-manager/cache/affinity.go
klihub: If that works, then definitely. I'm just still such a golang n00b that I didn't know you can do that.

klihub:
> nit: the commit message subject could be shortened a bit. E.g.:
> 
> ```
> cache: add pod.GetContainer()
> 
> New method for looking up containers from pod by name.
> ```

Rewrote commit message accordingly.

pkg/cri/resource-manager/cache/affinity.go
klihub: I'm not sure I understand... The idea here is that we default to looking up labels if only a key is provided without any `key-domain/`-prefix.

pkg/cri/resource-manager/cache/affinity.go
klihub: Changed it to be like that.

pkg/cri/resource-manager/cache/affinity.go
klihub: Changed both switches on key to strings.ToLower(key).

pkg/cri/resource-manager/cache/affinity.go
marquiz: We have this
```
		if len(ref) == 0 {
			break
		}
```
later in the loop so it looks like a while loop to me

pkg/cri/resource-manager/cache/affinity.go
marquiz: > If that works, then definitely. I'm just still such a golang n00b that I didn't know you can do that.

Ha, I'm the total n00b here. But it's comforting to see that even you don't know everything, yet 😄

pkg/cri/resource-manager/cache/affinity.go
klihub: Ah, that one... Yes, you're right, it is essentially a while loop. I'll update it.

klihub:
> TODO: document the feature

Added documentation.

docs/container-affinity.md
```diff
@@ -0,0 +1,225 @@
...
+
+  - Equals: equality, true if the *value of key* equals the single item in *values*
+  - NotEqual: inequality, true if the *value of key* is not equal to the single item in *values*
+  - In: membership, true if *value of key* equals to any among *values*
+  - NotIn: negated membership, true if the *value of key* is not equal to any among *values*
+  - Exists: true if the given *key* exists with any value
+  - NotExists: true if the given *key* does not exist
+
+
+The effective affinity between containers C_1 and C_1, A(C_1, C_2) is the sum of the
```
marquiz: s/C_1 and C_1/C_1 and C_2/

docs/container-affinity.md
```diff
@@ -0,0 +1,225 @@
...
+            - valueN
+          weight: w
+```
+
+## Affinity Semantics
+
+An affinity consists of three parts:
+
+  - `scope`: defines which containers this affinity is evaluated against
+  - `expression`: defines for which containers (within the scope) the affinity applies to
```
marquiz: Just for coherency: should we use term "`match` expression (just *expression* later in this document)" or something similar here as the key used in the annotation is *match*?

docs/container-affinity.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Affinity Annotation Syntax
+
+*Affinities* are defined as the `cri-resource-manager.intel.com/affinity` annotation.
+*Anti-affinities* are defined as the `cri-resource-manager.intel.com/anti-affinity`
+annotation. They are specified in the `metadata` section of the `Pod YAML`, under
+`annotations` as a dictionary, with each dictionary key being the name of the
+*container* within the Pod to which the annotation belongs to.
+
+```
```
grahamwhaley: if you make this block enclosed in ```` ```yaml ````, then it should get syntax highlighted ;-)

docs/container-affinity.md
```diff
@@ -0,0 +1,225 @@
...
+it *selects the containers* the affinity really has an effect on. The *weight* specifies
+whether the effect is a *pull* or a *push*. *Positive* weights cause a *pull* while
+*negative* weights cause a *push*. Additionally, the *weight* specifies *how strong* the
+push or the pull is. This is useful in situations where the policy needs to make some
+compromises because an optimal placement is not possible. The weight then also acts as
+a way to specify preferences of priorities between the various compromises: the heavier
+the weight the stronger the pull or push and the larger the propbability that it will be
+honored, if this is possible at all.
+
+The scope can be omitted from an affinity in which case it implies *Pod scope*, IOW
```
grahamwhaley: I'd make that "I.O.W", or preferably to my eyes fully expand the acronym.

docs/container-affinity.md
```diff
@@ -0,0 +1,225 @@
...
+  - key: specifies what *metadata* to pick from a container for evaluation
+  - operation (op): specifies what *logical operation* the expression evaluates
+  - values: a set of *strings* to evaluate the the value of the key against
+
+Essentially an expression defines a logical operation of the form (key op values).
+Evaluating this logical expression will take the value of the key in  which
+either evaluates to true or false. 
+a boolean true/false result. Currently the following operations are supported:
+
+  - Equals: equality, true if the *value of key* equals the single item in *values*
```
grahamwhaley: I'd consider putting the operator names in `backticks` to make them stand out.

docs/container-affinity.md
```diff
@@ -0,0 +1,225 @@
...
+A(C_1, C_2) and A(C_2, C_1) can and will be different unless the affinity annotations
+crafted to prevent this (IOW to be symmetric). Moreover, A(C_1, C_2) is calculated
+and taken into consideration during resource allocation for C_1, while A(C_2, C_1)
+is calculated and taken into account during resource allocation for C_2. This might be
+changed in a future version.
+
+
+## Examples
+
+Put the container `peter` close to the container `sheeps` but far away from the
```
grahamwhaley: humorous nitpick.... the plural of sheep... is sheep (not sheeps :-) )

docs/container-affinity.md
klihub: Ah, thanks. I'll update.

docs/container-affinity.md
klihub: Thanks ! I didn't know that. Updated.

grahamwhaley: np. github markdown supports a huge list of languages it knows about: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

docs/container-affinity.md
klihub: Good point. Added backticks.

docs/container-affinity.md
klihub: Expanded (and rephrased in a few other places).

docs/container-affinity.md
klihub: I changed both 'scope' and 'expression' to 'scope expression' and 'match expression', since they are both expressions (as a bit later stated in the document).

docs/container-affinity.md
klihub: Ah, thanks. If only Shaun the Sheep had a proper spoken audio track... with so many hours under my belt I might have realized this then. ;-)

klihub:
Oh... and Peter is careless with his affinities and tries to put himself both close to and far away from the wolf. I'll fix that one too...

PR #144, opened 2020-01-21
[rebalancing] cache: add support for implicit container affinities.

klihub:
Add support for registering implicit affinities. An implicit affinity is defined by a container eligibility test function and an affinity expression. When affinities for a container are looked up by the cache, all those implicit affinities get included for which the eligibility function called with the Container returns true.

Implicit affinities can be used to inject policy-specific affinities to all or a selected subset of containers, for instance to inject automatic AVX512 affinities and anti-affinities to containers.

### DIFF
--- a/pkg/cri/resource-manager/cache/affinity.go
+++ b/pkg/cri/resource-manager/cache/affinity.go
@@ -321,3 +321,22 @@ func (pca *podContainerAffinity) parseFull(pod *pod, value string, weight int32)
 
 	return nil
 }
+
+// GlobalAffinity creates an affinity with all containers in scope.
+func GlobalAffinity(key string, weight int32) *Affinity {
+	return &Affinity{
+		Scope: &Expression{
+			Op: AlwaysTrue, // evaluate against all containers
+		},
+		Match: &Expression{
+			Key: key,
+			Op:  Exists,
+		},
+		Weight: weight,
+	}
+}
+
+// GlobalAntiAffinity creates an anti-affinity with all containers in scope.
+func GlobalAntiAffinity(key string, weight int32) *Affinity {
+	return GlobalAffinity(key, -weight)
+}
--- a/pkg/cri/resource-manager/cache/affinity.go
+++ b/pkg/cri/resource-manager/cache/affinity.go
@@ -68,6 +68,12 @@ const (
 	AlwaysTrue = "AlwaysTrue"
 )
 
+// ImplicitAffinity is an affinity that gets implicitly added to all eligible containers.
+type ImplicitAffinity struct {
+	Eligible func(Container) bool // function to determine if Affinity is added to a Container
+	Affinity *Affinity            // the actual implicitly added Affinity
+}
+
 // Validate checks the affinity for (obvious) invalidity.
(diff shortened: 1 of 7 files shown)


marquiz:
> I didn't yet fully understand this? Maybe a short example (e.g. config snippet) in the commit message could clarify this

OK, now I understood - it is meant to be used with the `Scope` parameter of the affinity. But still, maybe a bit more clarification about scope and/or an example in the commit message would be useful(?)

pkg/cri/resource-manager/cache/affinity.go
```diff
@@ -341,3 +347,17 @@ func GlobalAffinity(key string, weight int32) *Affinity {
 func GlobalAntiAffinity(key string, weight int32) *Affinity {
 	return GlobalAffinity(key, -weight)
 }
+
+// AddImplicitAffinities registers a set of implicit affinities.
+func (cch *cache) AddImplicitAffinities(implicit map[string]*ImplicitAffinity) error {
+	for name := range implicit {
+		if existing, ok := cch.implicit[name]; ok {
+			return cacheError("implicit affinity %s already defined (%s)",
+				name, existing.Affinity.String())
```
marquiz: Is an error or just overriding the previous rule "the right thing to do"? I don't know. But what if a policy would like to change the implicit affinity rules. for example?

klihub:
> > I didn't yet fully understand this? Maybe a short example (e.g. config snippet) in the commit message could clarify this
> 
> OK, now I understood - it is meant to be used with the `Scope` parameter of the affinity. But still, maybe a bit more clarification about scope and/or an example in the commit message would be useful(?)

Yes, it's for Scope. I'll add an example in the commit message and maybe to the comments in the code as well.

pkg/cri/resource-manager/cache/affinity.go
klihub: That is true and I was considering adding a DeleteImplicitAffinities(name string), but then I didn't bother. Mostly because we don't need it ATM, but also because, since we have the Eligible callback/function in ImplicitAffinity, you can technically always use that work around the lack of Delete if you need to dynamically change the per-container affinity behavior for your policy. I can add the Delete functions nevertheless...

klihub: Added cache.DeleteImplicitAffinities([]string).

pkg/cri/resource-manager/cache/pod.go
```diff
@@ -109,6 +111,32 @@ func (p *pod) GetContainers() []Container {
...
+		p.containers = make(map[string]string)
+	}
+
+	for _, c := range p.GetContainers() {
+		cptr := c.(*container)
+		p.containers[cptr.Name] = cptr.ID
+		if cptr.Name == name {
+			found = cptr
+		}
+	}
```
klihub: Hmm... this looks wrong. Seems like the idea was to cache successful lookups on the slow path, but only the slow-path has been now implemented. The fast-path (else branch) is missing altogether... I'll fix this as well.

klihub: Fixed.

PR #147, opened 2020-01-23
[rebalancing] policy: add Rebalance(), stub backend implementations.

klihub:
Add a mechanism for optimizing resource allocation for the current set of running containers.

Rebalance() is the primary mechanism for rebalancing/optimizing the resource allocations for the current set of running containers. The resource manager will (eventually) trigger rebalancing periodically. It is up to the policy backend to decide what kind of rebalancing strategy it implements.

The current API presents no way for the upper layer to restrict which containers are in the scope of rebalancing. Regardless, implementations should honor the existing/implied QoS-semantics of pods/containers: containers in Guaranteed QoS pods are not to be disrupted/touched.

### DIFF
--- a/pkg/cri/resource-manager/policy/policy.go
+++ b/pkg/cri/resource-manager/policy/policy.go
@@ -104,6 +104,8 @@ type Backend interface {
 	ReleaseResources(cache.Container) error
 	// UpdateResources updates resource allocations of a container.
 	UpdateResources(cache.Container) error
+	// Rebalance tries an optimal allocation of resources for the current container.
+	Rebalance() (bool, error)
 	// ExportResourceData provides resource data to export for the container.
 	ExportResourceData(cache.Container) map[string]string
 }
@@ -120,6 +122,8 @@ type Policy interface {
 	ReleaseResources(cache.Container) error
 	// UpdateResources updates resource allocations of a container.
 	UpdateResources(cache.Container) error
+	// Rebalance tries to find an optimal allocation of resources for the current containers.
+	Rebalance() (bool, error)
 	// ExportResourceData exports/updates resource data for the container.
 	ExportResourceData(cache.Container)
 }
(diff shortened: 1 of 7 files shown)


klihub:
Added a note to the commit message about the topology-aware policy having a dumb Rebalance() implementation instead of a stub one.

klihub:
> Added a note to the commit message about the topology-aware policy having a dumb Rebalance() implementation instead of a stub one.

Added a note to the commit message about the topology-aware policy having a dumb Rebalance() implementation instead of a stub one.

klihub:
> Looks good to me. Maybe the commit message could state that there is a not-quite-stub example implementation of `Rebalance()` in the topology aware policy(?)

Added a note to the commit message about the topology-aware Rebalance() implementation being dumb instead of stub.

pkg/cri/resource-manager/policy/policy.go
```diff
@@ -235,6 +239,11 @@ func (p *policy) UpdateResources(c cache.Container) error {
 	return p.backend.UpdateResources(c)
 }
 
+// Rebalance tries to find a more optimal allocation of resources for the current containers.
+func (p *policy) Rebalance() (bool, error) {
+	return p.backend.Rebalance()
```
mythi: Would `Rebalance()` be relevant/useful for all policies? It'd also be possible to leave it optional and check for 

```
if pr, ok := p.backend.(Rebalance()); ok {
    return pr.Rebalance()
} else {
   return false, nil
}
```

klihub: Well, we could split out Rebalance() to a policy.Rebalance interface which would not be part of policy.Backend. Then we could do an attempted assertion like that and not call the function if the policy does not support it. So far we have tried to keep the backends identical, without any optional interfaces, and let them provide dummy functions for functionality they prefer not to implement.

mythi: OK, makes sense, +1

PR #149, opened 2020-01-29
cgroupstats: further unify cgroupstats getters.

klihub:
This patchset unifies parsing in the complex cgroupstats getters. It
  - adds a configurable parser for parsing multiple single-line entries from a file
  - reworks GetBlkioThrottleBytes and GetNumaStats using the above function
  - adds support for delayed evaluation of (expensive) arguments for log.Debug()

### DIFF
--- a/pkg/cgroups/cgroupstats.go
+++ b/pkg/cgroups/cgroupstats.go
@@ -17,11 +17,11 @@ package cgroups
 import (
 	"fmt"
 	"io/ioutil"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
 
+	logger "github.com/intel/cri-resource-manager/pkg/log"
 	"github.com/intel/cri-resource-manager/pkg/sysfs"
 )
 
@@ -86,6 +86,9 @@ type GlobalNumaStats struct {
 	LocalNode     int64
 	OtherNode     int64
 }
+
+// our logger instance
+var log = logger.NewLogger("cgroup-metrics")
 
 func readCgroupFileLines(filePath string) ([]string, error) {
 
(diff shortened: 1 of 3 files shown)


pkg/sysfs/parsers.go
```diff
@@ -115,6 +133,65 @@ func parseNumeric(path, value string, ptr interface{}) error {
...
+		vmap.(map[string]uint)[key] = uint(unum * uint64(unit))
+	case map[string]uint8:
+		unum, err = strconv.ParseUint(numstr, 0, 8)
+		vmap.(map[string]uint8)[key] = uint8(unum * uint64(unit))
+	case map[string]uint16:
+		unum, err = strconv.ParseUint(numstr, 0, 16)
+		vmap.(map[string]uint16)[key] = uint16(unum * uint64(unit))
+	case map[string]uint32:
+		unum, err = strconv.ParseUint(numstr, 0, 32)
+		vmap.(map[string]uint32)[key] = uint32(unum * uint64(unit))
```
ipuustin: Why would we want anything other than `{int64,uint64,float64}` as internal result values? When I see code such as `uint8(unum * uint64(unit))` I start thinking that this is going overflow.

pkg/cgroups/cgroupstats.go
```diff
@@ -368,72 +379,73 @@ func GetNumaStats(cgroupPath string) (NumaStat, error) {
...
-		for _, nodeEntry := range split[1:] {
-			nodeamount := strings.Split(nodeEntry, "=")
-			if len(nodeamount) != 2 {
-				return NumaStat{}, fmt.Errorf("error parsing file %s", entry)
+	path := filepath.Join(cgroupPath, cgroupEntry)
+	err := sysfs.ParseFileByLines(path,
+		// split line to fields by spaces
+		func(line string) ([]*sysfs.Field, error) {
+			split := strings.Split(line, " ")
+			fields := make([]*sysfs.Field, 0, len(split))
```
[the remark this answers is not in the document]
klihub: It's an idiomatic pattern I picked up for slice-construction in a loop: voluntarily pay the price of a single call per iteration to the fast path of the builtin append instead of ending up hard-to find bugs with trailing zero-value entries in case we ever break out early form the loop and (try to) return the items collected so far.

But you're right, with this simple loop it cannot happen. Updated code accordingly.

pkg/sysfs/parsers.go
```diff
@@ -166,3 +246,74 @@ func ParseFileEntries(path string, values map[string]interface{}, pickFn PickEnt
...
+
+	for _, line := range strings.Split(string(data), "\n") {
+		fields, err := splitFn(line)
+		switch {
+		case err == ErrSkip:
+			continue
+		case err != nil:
+			return nil
+		}
+		sort.Slice(fields, func(i, j int) bool { return fields[i].Index <= fields[j].Index })
```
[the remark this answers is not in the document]
klihub: Added a comment explaining why what the caller-provided Index indicates and why we sort.

pkg/sysfs/parsers.go
klihub: Valid point. Updated to use highest-precision internal representations and added checks for trivial overflow/underflow of the numeric part without the unit.

klihub:
> I'm bit on the fence with the generalization of the parsing logic... on the other hand it's good to have reusable code and there may well be more files which we are parsing in the future, but on the other hand the parsing logic starts to be pretty abstract and tricky to follow.

I'm pretty sure that with a bit of head scratching it's possible to come up with a simpler yet equally generic and customizable parser for line-oriented files. It's just not something I'd personally want to spend too much time with right now. If someone has a much better alternative, I'm glad to take it though.

pkg/log/log.go
```diff
@@ -55,6 +56,62 @@ type Logger interface {
...
 
+// Delayed implement delayed evaluation, useful for lowering the overhead suppressed log.Debug.
+type Delayed interface {
+	String() string
+}
+
+// delay implements Delayed.
+type delay struct {
+	o interface{}
+	Delayed
```
marquiz: You could drop this `Delayed`

pkg/sysfs/parsers.go
```diff
@@ -115,6 +134,145 @@ func parseNumeric(path, value string, ptr interface{}) error {
...
+		vm[key] = num
+	case map[string]uint64:
+		typeName, max := "uint64", uint64(math.MaxUint64)
+		if unum, err = parseUintWithUnit(path, numstr, unit); err != nil {
+			return err
+		}
+		if unum > max {
+			return sysfsError(path, "numeric value %s overflows %s", value, typeName)
+		}
+		vm[key] = unum
```
marquiz: There's a lot of duplicate code. Just a thought: could this be simplified by doing the limits checking in `parseUintWithUnit()` / `parseIntWithUnit()`? Either by giving the number of bits or min/max limits as arguments. WDYT?

klihub: Yes, that's probably a good idea. I'll change it accordingly.

pkg/sysfs/parsers.go
```diff
@@ -166,3 +327,76 @@ func ParseFileEntries(path string, values map[string]interface{}, pickFn PickEnt
...
+		for idx, f := range fields {
+			ptr, err := pickFn(idx, f)
+			switch {
+			case err == ErrSkip:
+				continue
+			case err != nil:
+				return err
+			}
+
+			switch ptr.(type) {
```
marquiz: Would be possible to use the `p := ptr.(type)` construct here, agani?

marquiz:
> I'm bit on the fence with the generalization of the parsing logic... on the other hand it's good to have reusable code and there may well be more files which we are parsing in the future, but on the other hand the parsing logic starts to be pretty abstract and tricky to follow.

After finally having time to read through the last patch, I'm inclined to second @ipuustin here. The code starts to be hard to follow, especially at this time on Friday evening 😄

pkg/sysfs/parsers.go
klihub: I wonder how long it will take to get this into my thick skull... if ever. ;-)
I'll update the code.

klihub:
Closing this for now...

PR #151, opened 2020-01-31
[rebalancing] resource-manager: add skeletal events- and metrics-based rebalancing

klihub:
Note: *this PR set requires (IOW includes) #156 .*

This patch set adds a new metrics collector subpackage to the top-level generic resource-manager and initial code to the resource manager itself to trigger periodic rebalancing of the allocated container resources.

The metrics collector subpackage periodically
  - polls the low-level collectors to acquire up-to-date *raw data*,
  - preprocesses the data to produce events for the resource manager, and
  - delivers these events to the resource manager

The events produced are (supposed to be) more *domain-specific* and better suited for consumption by the resource manager than just the collected low-level raw metrics data.

The resource manager
  - pulls events from the metrics collector
  - periodically triggers rebalancing

This initial patch set only implements events related to AVX512 instruction usage. Producing relevant data for the policy-specific rebalancing algorithm needs further patches.

### DIFF
--- a/pkg/cgroups/v2path.go
+++ b/pkg/cgroups/v2path.go
@@ -1,0 +1,29 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cgroups
+
+import (
+	"flag"
+)
+
+var (
+	// V2path is the mount point for the cgroup V2 pseudofilesystem.
+	V2path string
+)
+
+func init() {
+	flag.StringVar(&V2path, "cgroupv2-path", "/sys/fs/cgroup/unified",
+		"Path to cgroup-v2 mountpoint")
+}
--- a/pkg/cri/resource-manager/events.go
+++ b/pkg/cri/resource-manager/events.go
@@ -1,0 +1,139 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resmgr
+
+import (
+	"time"
+
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/metrics"
+	logger "github.com/intel/cri-resource-manager/pkg/log"
+)
+
+// Our logger instance for events.
+var evtlog = logger.NewLogger("events")
+
+// setupEventProcessing sets up event and metrics processing.
+func (m *resmgr) setupEventProcessing() error {
+	var err error
+
+	m.events = make(chan interface{}, 8)
+	m.stop = make(chan interface{})
+	options := metrics.Options{
+		PollInterval: opt.MetricsTimer,
+		Events:       m.events,
+	}
+	if m.metrics, err = metrics.NewMetrics(options); err != nil {
+		return resmgrError("failed to create metrics (pre)processor: %v", err)
+	}
+
+	return nil
+}
+
+// startEventProcessing starts event and metrics processing.
+func (m *resmgr) startEventProcessing() error {
+	if err := m.metrics.Start(); err != nil {
+		return resmgrError("failed to start metrics (pre)processor: %v", err)
+	}
+
+	stop := m.stop
+	go func() {
+		rebalanceTimer := time.NewTicker(opt.RebalanceTimer)
+		for {
+			select {
+			case _ = <-stop:
+				return
+			case event := <-m.events:
+				m.processEvent(event)
+			case _ = <-rebalanceTimer.C:
+				if err := m.RebalanceContainers(); err != nil {
+					evtlog.Error("rebalancing failed: %v", err)
+				}
+			}
+		}
+	}()
+
+	return nil
+}
+
+// stopEventProcessing stops event and metrics processing.
+func (m *resmgr) stopEventProcessing() {
+	close(m.stop)
+	m.metrics.Stop()
+}
+
+// SendEvent injects the given event to the resource manaager's event processing loop.
+func (m *resmgr) SendEvent(event interface{}) error {
+	if m.events == nil {
+		return resmgrError("can't send event, no event channel")
+	}
+	select {
+	case m.events <- event:
+		return nil
+	default:
+		return resmgrError("can't send event of type %T, event channel full", event)
+	}
+}
+
+// processEvent processes the given event.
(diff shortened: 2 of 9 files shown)


pkg/cri/resource-manager/metrics/avx.go
```diff
@@ -0,0 +1,62 @@
...
+
+	avx, ok := raw["avx_switch_count_per_cgroup"]
+	if !ok {
+		return nil
+	}
+	dump("AVX context switches", avx)
+
+	ratio := map[string]float64{}
+	for _, v := range avx.Metric {
+		cgroup := strings.TrimPrefix(v.Label[0].GetValue(), cgroups.V2path)
```
mythi: `filepath.Rel()`?

pkg/cri/resource-manager/metrics/avx.go
```diff
@@ -0,0 +1,62 @@
...
+	for _, v := range all.Metric {
+		cgroup := strings.TrimPrefix(v.Label[0].GetValue(), cgroups.V2path)
+		ratio[cgroup] /= v.Gauge.GetValue()
+	}
+
+	usage := map[string]bool{}
+	for cgroup, use := range ratio {
+		active := use >= m.opts.AvxThreshold
+		log.Debug(" %s AVX ratio = %f, active?: %v", cgroup, use, active)
+		usage[cgroup] = active
```
mythi: this could now be enhanced with the time stamp value. it'd filter scenarios where the ratio is high due to high avx512 bursts that happen infrequently.

pkg/cri/resource-manager/metrics/metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	m.raw = f
+	return nil
+}
+
+// process processes the collected raw metrics.
+func (m *Metrics) process() error {
+	raw := map[string]*model.MetricFamily{}
+	for _, f := range m.raw {
+		log.Debug("got event %s", *f.Name)
```
mythi: reuse `dump()` here?

pkg/metrics/metrics.go
```diff
@@ -9,7 +9,7 @@ import (
 var (
 	builtInCollectors    = make(map[string]InitCollector)
 	registeredCollectors = []prometheus.Collector{}
-	log                  = logger.NewLogger("metrics")
+	log                  = logger.NewLogger("raw-metrics")
```
mythi: Should we rename this package to `collectors`. That's what they are and CRI-RM uses collectors to build metrics... just a thought.

klihub: Yes, that'd probably be a good idea.

pkg/cri/resource-manager/metrics/avx.go
klihub: Good point, I'll update.

pkg/cri/resource-manager/metrics/avx.go
klihub: Yes, the current code is in not (even close to) what it should be in the end. The AVX PoC bits were put in because I did not want to put in an empty wireframe implementation/plumbing without any triggers for policy decision making.

With this patch series I was hoping/my aim was just to get something minimal/enablers merged to start building upon. The next steps then IMO would be to
  - distill more meaningful metrics data for policy consumption
  - inject metrics data (and the related interfaces) to the cache
  - prototype with actual policy rebalancing algorithms

pkg/metrics/metrics.go
klihub: But I'd like to do that in connection with other API changes there. I talked about this briefly in the last Friday chat, when I mentioned that I'd like to turn the current model inside out. Currently we 
  1. collect raw data from the kernel
  2. turn it into prometheus format and collect it
  3. try to turn it into something better suited for policies
 [4. forward raw data + chewed upon/correlated metrics to prometheus. (This still missing, it's only on the original proto branch, I'll try to clean it up and file a PR for it.)

Now the prometheus data model is not well suited / very inefficient for any type of correlation/processing other than plotting or something else a time series DB is good at.
Instead of the current model I'd like to change to this one:

  1. low-level collect data either in some resource-manager/metrics.SomeMetrics format, or something which is a natural reflection of what the kernel provides. This we already have in practice in the low-level collectors. They just convert it into the prometheus format when polled.
  2. pass events in the native format up in the stack when polled by resource-manager/metrics
  3. define an interface collector.ToMetricFamily(interface{}, []*model.MetricFamily) which all the collectors must implement (or actually I think the data types should implement them)
  4. do the conversion on-demand only when/if we proxy data to prometheus

I think this would be a much more sensible model avoiding double conversion/correlation and having to operate on data in a very unsuitable format.

klihub: But I changed the logger source name to collectors. ;-)

pkg/cri/resource-manager/metrics/metrics.go
klihub: Updated.

pkg/cri/resource-manager/metrics/avx.go
mythi: yes, let's keep it as it is now

pkg/metrics/metrics.go
mythi: > Now the prometheus data model is not well suited / very inefficient for any type of
> correlation/processing other than plotting or something else a time series DB is good at.

For avx I still want to experiment with prometheus summary/histogram functions to see if they'd somehow help with the tagging and policy/rebalancing: for instance histograms for per-CPU activity to verify "non-AVX" cores indeed remain "zero activity" OR per-container activity to rebalance "top peak" containers... (maybe in this case separate tagging/conversion is even not needed?)

pkg/cgroups/v2path.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
marquiz: Maybe we should update the copyright year :smile:

pkg/metrics/metrics.go
klihub: Yes, sure. In the above model nothing prevents us, in princpiple, from declaring that the native data type for AVX usage collection is a *MetricFamily, if that turns out to offer tangible benefits. Then the above (using histrogam, etc. functions from prometheus) should become straightforward, without any intermediary steps.

However, to reiterate my point about the prometheus data being inconvenient for us, look at how even for AVX, which requires almost no correlation beyond the eventual resolution (in the resource-manager layer) from cgroup to cache.Container, we need to loop through two MetricFamilies, hashing everything by cgroup path to two different tables, just to determine how actively a cgroup has been using AVX instructions.

And this gets at least linearly worse (scaling by the number of pieces of information you want to cross-correlate) as you try to stitch back together pieces of information which was first chopped up by the low-level raw metrics collectors. It would make more sense IMO to do the necessary chopping up at the edge, on the cri-resmgr -> prometheus edge, and only if/when we proxy to prometheus.

pkg/cgroups/v2path.go
klihub: That code was written last year... ;-)

klihub: But I'll update anyway...

pkg/cri/resource-manager/events.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// processEvent processes the given event.
+func (m *resmgr) processEvent(e interface{}) {
+	evtlog.Debug("received event of type %T...", e)
+
+	m.Lock()
+	defer m.Unlock()
+
+	switch e.(type) {
```
marquiz: you could use the `v := e.(type)` construct here to avoid the type assertions later

pkg/cri/resource-manager/metrics/avx.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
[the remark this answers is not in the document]
klihub: updated

pkg/cri/resource-manager/events.go
klihub: For some reason I never seem to learn this trick. I remember you reminded me at least once in another PR about the same. Updated.

klihub:
@mythi Is this good to merge in its current form ?

klihub:
> +1. we should continue brainstorming the metrics collection later

Yes, definitely. We can start discussing it already during the tomorrow session, if we have time.

PR #153, opened 2020-02-04
topology: get hints for VFIO devices

mythi:
An IOMMU group is the smallest set of physical devices that can be
passed to a VM. For each vfio virtual devices, we return topology hints
for all devices in the corresponding IOMMU group.


mythi:
would like to get this included in #158

Issue #152, opened 2020-02-03
go get fails to import cri-resource-manager


bart0sh:
Here is how it happens:
```
$ go get github.com/intel/cri-resource-manager/pkg/sysfs
go: finding github.com/intel/cri-resource-manager v0.2.0
go: finding github.com/intel/cri-resource-manager/pkg latest
go: finding github.com/intel/cri-resource-manager/pkg/sysfs latest
go: finding github.com/intel/cri-resource-manager v0.2.0
go: downloading github.com/intel/cri-resource-manager v0.2.0
verifying github.com/intel/cri-resource-manager@v0.2.0: github.com/intel/cri-resource-manager@v0.2.0: reading https://sum.golang.org/lookup/github.com/intel/cri-resource-manager@v0.2.0: 410 Gone
```

The actual error is this:
```
not found: unzip /tmp/gopath/pkg/mod/cache/download/github.com/intel/cri-resource-manager/@v/v0.2.0.zip: malformed file path "pkg/sysfs/testdata/sys/devices/pci0000:00/0000:00:02.0/class": invalid char ':'
```

I'll try to come up with a fix soon.

klihub:
Hmm, strange. Looks like the toolchain cannot deal with files with a colon in their name.

@bart0sh Since the problem seems to be only with sysfs test data, where colons do appear in file names, one ugly workaround would be to hide those files from the toolchain during go get, for instance movin them to an in-repo tarball which is then extracted  prior to running the tests (by appropriately crafted Makefile rules).

klihub:
This is a toolchain problem. See any of the many bug reports, for instance [this](https://github.com/golang/go/issues/27642).

The suggested semi-official upstream workaround is to scatter around empty go.mod files as needed in your repository to fool the toolchain to consider the subtrees with offending file names separate packages so they get excluded from the zip created during go get.

bart0sh:
I'm going to fix it by removing pkg/sysfs/testdata and generating test data dynamically in the test code. apart from fixing this particular issue it would also improve synchronization between test data and code.

klihub:
> I'm going to fix it by removing pkg/sysfs/testdata and generating test data dynamically in the test code. apart from fixing this particular issue it would also improve synchronization between test data and code.

Hmm... I'm not sure if that is the right fix. I think it would be good to keep real world HW sysfs-dumps in the repo as test data and verify that our discovery code works correctly against those.

bart0sh:
[...]

PR #154, opened 2020-02-04
Generate test data for TestNewTopologyHints

bart0sh:
Got rid of keeping test content in pkg/sysfs/testdata by
dynamically generating it.

This should fix 'go get github.com/intel/cri-resource-manager/pkg/sysfs'
failure caused by unzip failing to unpack files with ':' in their names:
```
  not found: unzip github.com/intel/cri-resource-manager/@v/v0.2.0.zip:
    malformed file path "pkg/sysfs/testdata/sys/devices/pci0000:00/0000:00:02.0/class":
      invalid char ':'
```

Fixes: #152


klihub:
I filed a PR, #155 as an alternative fix.
Please, @kad , @marquiz , @mythi review, so we can resolve the blocker for the device plugins which try to import this package as a dependency.

klihub:
Closing as #155 got merged as an alternative workaround to the same problem.

PR #156, opened 2020-02-05
[rebalancing] resource-manager: chop/clean up {set,start}up sequence code.

klihub:
Clean up the generic resource manager code slightly by
chopping up the setup/startup sequence to smaller, more
orthogonal pieces.

### DIFF
--- a/pkg/cri/resource-manager/resource-manager.go
+++ b/pkg/cri/resource-manager/resource-manager.go
@@ -53,76 +53,42 @@ type resmgr struct {
 
 // NewResourceManager creates a new ResourceManager instance.
 func NewResourceManager() (ResourceManager, error) {
-	var err error
+	m := &resmgr{Logger: logger.NewLogger("resource-manager")}
 
-	if opt.ForceConfig != "" && opt.FallbackConfig != "" {
-		return nil, resmgrError("both fallback (%s) and forced (%s) configuration given",
-			opt.FallbackConfig, opt.ForceConfig)
-	}
-
-	m := &resmgr{
-		Logger: logger.NewLogger("resource-manager"),
+	if err := m.setupCache(); err != nil {
+		return nil, err
 	}
 
-	// Set-up connection to cri-resmgr agent
-	m.agent, err = agent.NewAgentInterface(opt.AgentSocket)
-	if err != nil {
-		m.Warn("failed to connect to cri-resmgr agent: %v", err)
+	if err := m.checkOpts(); err != nil {
+		return nil, err
 	}
 
-	ropts := relay.Options{
-		RelaySocket:   opt.RelaySocket,
-		ImageSocket:   opt.ImageSocket,
-		RuntimeSocket: opt.RuntimeSocket,
-	}
-	if m.relay, err = relay.NewRelay(ropts); err != nil {
-		return nil, resmgrError("failed to create resource manager: %v", err)
+	if err := m.setupConfigAgent(); err != nil {
+		return nil, err
 	}
 
-	copts := cache.Options{
-		CacheDir: opt.RelayDir,
-	}
-	if m.cache, err = cache.NewCache(copts); err != nil {
-		return nil, resmgrError("failed to create resource manager: %v", err)
+	if err := m.loadConfig(); err != nil {
+		return nil, err
 	}
 
-	if err = m.loadInitialConfig(); err != nil {
-		return nil, resmgrError("failed to load initial configuration: %v", err)
+	if err := m.setupConfigServer(); err != nil {
+		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/cri/resource-manager/resource-manager.go
```diff
@@ -53,188 +53,290 @@ type resmgr struct {
...
-	if opt.ForceConfig != "" && opt.FallbackConfig != "" {
-		return nil, resmgrError("both fallback (%s) and forced (%s) configuration given",
-			opt.FallbackConfig, opt.ForceConfig)
+	if err := m.setupCache(); err != nil {
+		return nil, err
 	}
 
-	m := &resmgr{
-		Logger: logger.NewLogger("resource-manager"),
+	if err := m.checkConfig(); err != nil {
```
marquiz: Just a thought, should this be named `checkOpts` or smth? The startup configuration management a bit convoluted mechanism and naming could help in understanding the code

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -53,188 +53,290 @@ type resmgr struct {
...
 	}
 
 	if err := m.relay.Start(); err != nil {
-		return resmgrError("failed to start relay: %v", err)
+		return resmgrError("failed to start CRI relay: %v", err)
 	}
 
-	if m.conf != nil {
-		m.cache.SetConfig(m.conf)
+	if opt.ForceConfig == "" {
```
marquiz: question: this is already checked in `setupConfigServer()`. Should/could we check `m.configServer != nil` here, instead? To me this would prolly be a bit more 
 readable/understandable but I guess this is very subtle and a personal preference.

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -53,188 +53,290 @@ type resmgr struct {
...
+		if err := m.configServer.Start(opt.ConfigSocket); err != nil {
+			return resmgrError("failed to start configuration server: %v", err)
+		}
+
+		if m.conf != nil {
+			m.cache.SetConfig(m.conf)
+			m.conf = nil
+		}
```
marquiz: There could be a comment clarifying the placement of this if/when we don't want to stash the forced config in cache.

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -53,188 +53,290 @@ type resmgr struct {
...
+	var err error
+
+	if opt.ForceConfig != "" {
+		m.Warn("using forced configuration %s, not starting configuration server",
+			opt.ForceConfig)
+	} else {
+		if m.configServer, err = config.NewConfigServer(m.SetConfig); err != nil {
+			return resmgrError("failed to create configuration notification server: %v", err)
+		}
+	}
```
marquiz: Related to my previous comment and thinking aloud: maybe we could remove the check here and create the `configServer` instance in any case. Then, just not start it in `resmgr.Start()`?

pkg/cri/resource-manager/resource-manager.go
klihub: > Should/could we check `m.configServer != nil` here, instead?

Well, this is an interface, so I think that wouldn't work (I mean testing nilness without reflect). If we want to do that, we need to change it to a *config.Server. Then we could check for nil, but then we'd always need to (*m.configServer).Func() it, IIRC.

pkg/cri/resource-manager/resource-manager.go
klihub: Hmm, it was called checkOpts in the beginning. The reason for changing it to checkConfig was mostly cosmetic, to have it group better with the other config-related functions. I'll change it back since it looks to be more intuitive.

pkg/cri/resource-manager/resource-manager.go
klihub: Yes, maybe we should do just that. Actually I have a related/similar patch in the kubernetes-v1.17.2 PR for the agent. When we force the config, we should disable the agent interface as well. There I had the same dilemma: how to administer it. Change it to pointer and test for nil, etc. I ended up adding support for explicit disabling to the agent interface and it errors when called if it is disabled. 

If we create the configServer without starting here, then after the other patch lands these would be symmetric.

pkg/cri/resource-manager/resource-manager.go
klihub: Added a comment.

pkg/cri/resource-manager/resource-manager.go
klihub: Changed the code accordingly.

pkg/cri/resource-manager/resource-manager.go
klihub: Changed to checkOpts.

PR #165, opened 2020-02-12
avx: fix building avx512.o on Linux 5.5

askervin:
- Prevent accidently including userspace headers (-nostdinc).
- Fix include order - #include <linux/types.h> included
  `include/uapi/linux/types.h` when `include/linux/types.h`
  was wanted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@ PROTOCODE := $(patsubst %.proto,%.pb.go,$(PROTOBUFS))
 
 CLANG := clang
 KERNEL_VERSION ?= $(shell uname -r)
-KERNEL_SRC_DIR ?= /usr/src/kernels/$(KERNEL_VERSION)
+KERNEL_SRC_DIR ?= /lib/modules/$(KERNEL_VERSION)/source
+KERNEL_BUILD_DIR ?= /lib/modules/$(KERNEL_VERSION)/build
 
 # Binaries and directories for installation.
 PREFIX     ?= /usr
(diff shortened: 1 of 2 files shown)


askervin:
Fixes the issue on Clear Linux 32280

`Linux single 5.5.2-903.native #1 SMP Fri Feb 7 14:50:11 PST 2020 x86_64 GNU/Linux`

Needs testing in other environments as well.

askervin:
@mythi , autodetect header directories works now on both Clear Linux and Debian.

Makefile
```diff
@@ -107,10 +108,20 @@ images: $(foreach dir,$(IMAGE_DIRS),image-$(dir))
...
+                      /arch/x86/include/generated/uapi
+
+KERNEL_INCLUDES := $(foreach kernel_dir,$(KERNEL_SRC_DIR) $(KERNEL_BUILD_DIR),$(addprefix -I,$(wildcard $(addprefix $(kernel_dir),$(KERNEL_INCLUDE_DIRS)))))
+
 libexec/%.o: elf/%.c
+	@if [ -z $(KERNEL_INCLUDES) ]; then echo "Cannot build $@: invalid KERNEL_SRC_DIR=$(KERNEL_SRC_DIR)"; exit 1; fi
 	$(Q)echo "Building $@"
 	$(Q)mkdir -p libexec
-	$(CLANG) -I$(KERNEL_SRC_DIR)/arch/x86/include -I$(KERNEL_SRC_DIR)/arch/x86/include/uapi -I$(KERNEL_SRC_DIR)/include -O2 -Wall -target bpf -c $< -o $@
+	$(CLANG) -nostdinc -D __KERNEL__ $(KERNEL_INCLUDES) -O2 -Wall -target bpf -c $< -o $@
```
klihub: Only one nitpick: since you're touching this line, can you also properly prefix it with $(Q), so it's consistent with the rest of the Makefile wrt. suppressed echoing vs. Q not being overridden ?

Otherwise this is good.

Makefile
```diff
@@ -107,10 +108,20 @@ images: $(foreach dir,$(IMAGE_DIRS),image-$(dir))
...
+                      /include/uapi \
+                      /include/generated/uapi \
+                      /arch/x86/include \
+                      /arch/x86/include/uapi \
+                      /arch/x86/include/generated/uapi
+
+KERNEL_INCLUDES := $(foreach kernel_dir,$(KERNEL_SRC_DIR) $(KERNEL_BUILD_DIR),$(addprefix -I,$(wildcard $(addprefix $(kernel_dir),$(KERNEL_INCLUDE_DIRS)))))
+
 libexec/%.o: elf/%.c
+	@if [ -z $(KERNEL_INCLUDES) ]; then echo "Cannot build $@: invalid KERNEL_SRC_DIR=$(KERNEL_SRC_DIR)"; exit 1; fi
```
klihub: Oh, and I think this is wrong... it should be if [ -z "$(KERNEL_INCLUDES)" ]; ... to avoid shell evaluation errors.

Makefile
askervin: $(Q) added

Makefile
askervin: Good findings.
- Added missing doublequotes.
- Added `$(strip ... )` around `KERNEL_INCLUDES`, because otherwise this `if` never prints user-friendly error. Without trimming `KERNEL_INCLUDES` it seems to always include one space.
- Added `$(Q)`.

PR #168, opened 2020-02-13
logging: fix logging races, rework/clean up logger.

klihub:
This patch set reworks the pkg/log, the common logger implementation with the primary aim being
  - to make the logger race-free/properly locked
  - clean up the implementation
  - add unit tests for logging

The set of fixes, changes and improvements are:
  - logging now should be race-free and multiple goroutine-safe (sans bugs)
  - filter early messages until/unless the daemon decides to start up, or there is an error
  - support for toggling full debugging on/off using SIGUSR1
  - format, serialize and emit messages from an asynchronous goroutine
  - properly handle and format multi-line messages without dedicated multi-line function calls
  - unit test

### DIFF
--- a/pkg/log/backend.go
+++ b/pkg/log/backend.go
@@ -1,0 +1,233 @@
+// Copyright 2019-2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"fmt"
+	"strings"
+)
+
+//
+// Logging backend interface and default fmt-based backend implementation.
+//
+
+// BackendFn is a functions that creates a Backend instance.
+type BackendFn func() Backend
+
+// Backend can format and emit log messages.
+type Backend interface {
+	// Name returns the name of this backend.
+	Name() string
+	// Log emits log messages with the given severity, source, and Printf-like arguments.
+	Log(Level, string, string, ...interface{})
+	// Block emits a multi-line log messages, with an additional line prefix.
+	Block(Level, string, string, string, ...interface{})
+	// Flush flushes and stops initial buffering synchronously
+	Flush()
+	// Sync waits for all messages to get emitted.
+	Sync()
+	// Stop stops the backend instance.
+	Stop()
+	// SetSourceAlignment sets the maximum prefix length for optional alignment.
+	SetSourceAlignment(int)
+}
+
+// RegisterBackend registers a logger backend.
+func RegisterBackend(name string, fn BackendFn) {
+	log.backend[name] = fn
+}
+
+const (
+	// FmtBackendName is the name of our simple fmt-based logging backend.
+	FmtBackendName = "fmt"
+	// fmtBackendQueueLen is the length of the internal fmt message queue.
+	fmtBackendQueueLen = 1024
+)
+
+const (
+	levelNop Level = iota + levelHighest
+	levelStop
+)
+
+// severity tags fmtBackend uses to prefix emitted messages with.
+var fmtTags = map[Level]string{
+	LevelDebug: "D: ",
+	LevelInfo:  "I: ",
+	LevelWarn:  "W: ",
+	LevelError: "E: ",
+	LevelFatal: "FATAL ERROR: ",
+	LevelPanic: "PANIC: ",
+}
+
+// fmtBackend is our simple, default fmt.Printf-based Backend.
+type fmtBackend struct {
+	q     chan *fmtReq // request channel
+	align int          // source alignment
+}
+
+// fmtReq is the request the fmt logger goroutine uses to emit messages.
+type fmtReq struct {
+	level  Level         // logging severity level
+	source string        // logger source
+	prefix string        // block prefix
+	format string        // log message format string
+	args   []interface{} // any arguments for format
+	sync   chan struct{} // reverse-ack for synchronous requests
+	flush  bool          // Flush() and stop queueing
+}
+
+// createFmtBackend creates an fmt Backend and starts its emitter goroutine.
+func createFmtBackend() Backend {
+	f := &fmtBackend{
+		q: make(chan *fmtReq, fmtBackendQueueLen),
+	}
+	go f.run()
+	return f
+}
+
+func (*fmtBackend) Name() string {
+	return FmtBackendName
+}
+
+func (f *fmtBackend) Log(level Level, source, format string, args ...interface{}) {
+	f.log(level, source, "", format, args...)
+}
+
+func (f *fmtBackend) Block(level Level, source, prefix, format string, args ...interface{}) {
+	f.log(level, source, prefix, format, args...)
+}
+
+func (f *fmtBackend) Flush() {
+	sync := make(chan struct{})
+	f.q <- &fmtReq{
+		level: levelNop,
+		flush: true,
+		sync:  sync,
+	}
+	_ = <-sync
+	close(sync)
+}
+
+func (f *fmtBackend) Sync() {
+	sync := make(chan struct{})
+	f.q <- &fmtReq{
+		level: levelNop,
+		sync:  sync,
+	}
+	_ = <-sync
+	close(sync)
+}
+
+func (f *fmtBackend) Stop() {
+	sync := make(chan struct{})
+	f.q <- &fmtReq{
+		level: levelStop,
+		sync:  sync,
+	}
+	_ = <-sync
+	close(sync)
+}
+
+func (f *fmtBackend) SetSourceAlignment(len int) {
+	f.align = len
+}
+
+// log pushes a new log message for emitting.
+func (f *fmtBackend) log(level Level, source, prefix, format string, args ...interface{}) {
+	var sync chan struct{}
+
+	// fatal errors are synchronous
+	if level > LevelError {
+		sync = make(chan struct{})
+	}
+
+	req := &fmtReq{
+		level:  level,
+		source: source,
+		prefix: prefix,
+		format: format,
+		args:   args,
+		sync:   sync,
+		flush:  level >= LevelError, // errors force buffer flush
+	}
+	f.q <- req
+
+	if sync != nil {
+		_ = <-sync
+		close(sync)
+	}
+}
+
+// run emits log messages for the fmtBackend.
+func (f *fmtBackend) run() {
+	buf := make([]*fmtReq, 0, fmtBackendQueueLen)
+
+	//
+	// logging loop
+	//
+	//   1. if we're not buffering, emit immediately
+	//   2. if we're buffering
+	//      - run out of space or asked to flush, flush buffer then emit message
+	//
+
+	for req := range f.q {
+		if buf == nil {
+			// past initial buffering => emit
+			f.emit(req)
+		} else {
+			// flush request or buffer full => flush, stop buffering, emit
+			if req.flush || len(buf) == cap(buf) {
+				for _, r := range buf {
+					f.emit(r)
+				}
+				f.emit(req)
+				buf = nil
+			} else {
+				buf = append(buf, req)
+			}
+		}
+		if req.sync != nil {
+			req.sync <- struct{}{}
+		}
+		if req.level == levelStop {
+			return
+		}
+	}
+}
+
+// emit formats and emits a single log message.
+func (f *fmtBackend) emit(req *fmtReq) {
+	if req.level > levelHighest {
+		return
+	}
+	length := len(req.source)
+	suflen := (f.align - length) / 2
+	prelen := (f.align - (length + suflen))
+	source := "[" + fmt.Sprintf("%*s", prelen, "") + req.source + fmt.Sprintf("%*s", suflen, "") + "]"
+
+	if req.prefix == "" {
+		for _, line := range strings.Split(fmt.Sprintf(req.format, req.args...), "\n") {
+			fmt.Println(fmtTags[req.level], source, line)
+		}
+	} else {
+		for _, line := range strings.Split(fmt.Sprintf(req.format, req.args...), "\n") {
+			fmt.Println(fmtTags[req.level], source, req.prefix, line)
+		}
+	}
+}
+
+func init() {
+	RegisterBackend(FmtBackendName, createFmtBackend)
+}
--- a/pkg/log/flags.go
+++ b/pkg/log/flags.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Intel Corporation. All Rights Reserved.
+// Copyright 2019-2020 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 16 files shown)


klihub:

pkg/log/logger.go
```diff
@@ -0,0 +1,298 @@
...
+	log.RLock()
+	defer log.RUnlock()
+
+	source, _ := log.sources[l]
+
+	return source
+}
+
+// Debug logs a debug message.
+func (l logger) Debug(format string, args ...interface{}) {
```
ipuustin: There could be one backend function for all these, maybe something like

```
func (l logger) logMessage(level Level, format string, args ..interface{}) {
	if cfg, active, emit := l.config(level); emit {
		active.Log(level, cfg.source(), format, args...)
	}
}
```

pkg/log/logger.go
```diff
@@ -0,0 +1,298 @@
...
+// config is the configuration parameters for a single logger stuffed into a single uint64
+type config struct {
+	id     uint16
+	enable uint16
+}
+
+// mkConfig creates a configuration with the given parameters.
+func mkConfig(logger logger, logging, tracing bool) config {
+	cfg := config{id: uint16(logger)}
+	if logging {
```
ipuustin: Maybe `cfg.setEnabled(logging, tracing)`

pkg/log/log_test.go
```diff
@@ -0,0 +1,528 @@
...
+			loggers[idx] = NewLogger(fmt.Sprintf("%d-<3>-log", idx))
+		case 4:
+			loggers[idx] = NewLogger(fmt.Sprintf("%d-<4>-logger", idx))
+		case 5:
+			loggers[idx] = NewLogger(fmt.Sprintf("%d-<5>-tester", idx))
+		default:
+			loggers[idx] = NewLogger(fmt.Sprintf("%d-<->-logger-tester", idx))
+		}
+
+		if (idx % 11) == 0 {
```
ipuustin: Maybe a comment for explaining the indexing would be nice. :-)

klihub:
> This PR is difficult to review thoroughly because it's so big. I have a bunch of mostly superficial code review comments, and some general questions:
> 
> * This is a pretty big component for logging. Was there no existing logging framework that we could have used?

We started with using glog (well, it was in the pluggable CPU policies, but still). Then kubernetes switched to klog (forked glog into klog). That turned into a disaster as it was impossible to include both glog and klog in the same binary without a runtime panic. I learned my lesson (once more, this time for golang): I don't want to depend on any single external logger. Instead I want to have a (relatively) simple logging infra which can be used to plug in external loggers but can fulfill our needs on its own. Maybe this is not it, maybe it is not good enough. But that's the basic intention/idea.

> * It's bit unclear to me what's the critical path in performance sense. There is a bitmask for finding out if logging or tracing is enabled, I guess this is to maximize perf?

For debug messages (because of the potential volume) the critical path is to decide as early as possible to not produce them. For messages we end up logging (both debug and other), taking out the penalty of message formatting from the goroutine which emits the log message.

> * It's great to have SIGUSR1 turn logging on. Can the log level be set also by environmental variable?

Not at the moment but it should be easy to add.

> * Should log rate limiting be left for the component which consumes the logs?

Our rate-limiter is opt-in. None of the log(ger)s are doing any rate-limited. You need to create a logger then wrap it in a rate-limiter for doing so. It is there, as an option, because it is ultimately the one producing log messages who know whether rate-limiting might be necessary. And if there is, it is a much better idea to drop log messages early in the log pipeline (preferable before they enter the pipeline at all) than late, especially at the end, once everybody went through a lot of trouble, bandwidth and processing power to deliver the messages.

klihub:
I'll try to take another look at this to see if I could easily make this less controversial...

pkg/log/logger.go
klihub: The lack of any kind of macros is a nuisance in golang. Here is a primary example...

For debug messages, I want to drop the suppressed ones as early as possible. Before they cross any potential goroutine boundary, IOW gets injected into a channel, then eventually pulled out at the other end by the backend just to get dropped, just like with the default fmt backend.

For the others it's true we could combine them to a single function. However, since we don't have macros I don't see a reasonable way of keeping the API like it is, yet combining Info, Warn, Error, Fatal, and Panic into a single function, while leaving Debug a separate one.

pkg/log/log_test.go
klihub: Added a clarifying comment.

pkg/log/logger.go
klihub: Good point. Updated acordingly.

klihub:
> * It's bit unclear to me what's the critical path in performance sense. There is a bitmask for finding out if logging or tracing is enabled, I guess this is to maximize perf?

The reason for that bitmask is related to performance but in a slightly different than you probably think. The code tries to minimize the length of locked basic blocks by creating a copy of the logger state at the package ingress functions/in logger.config(). The actual copying is done by not using a map[...]*config but a map[...]config, so when you look it up you automatically get a copy. Before writing this, I had absolutely no idea about performance of that in golang, so I wrote a simple piece of test code. It was doing a large number of map lookups of structures with various members and calculating and comparing average lookup times.

The results were so surprising that we spent a few hours with @marquiz playing around with the code, running it on different hardware. The surprising bit was that on some hardware the types of members had a much bigger effect on performance than the amount of memory moved around, even so that moving the exact same amount of memory could become orders of magnitude slower. I think (did not check the generated code) the compiler is simplistic and always generates a member by member copy even if it could be done smarter. Hence for some of the structure layouts with the same size the penalty of misaligned access gets overemphasized.

Since in golang your only control over alignment is the types you use (all types are aligned on their natural size) I ended up with that setup as a compromise.

pkg/log/logger.go
```diff
@@ -0,0 +1,295 @@
...
+	maxLoggers = math.MaxUint16
+	loggingBit = (1 << iota)
+	tracingBit
+)
+
+// config is the configuration parameters for a single logger stuffed into a single uint64
+type config struct {
+	id     uint16
+	enable uint16
+}
```
marquiz: Hmm, at a quick glance this bitwise logging config handling somehow feels a bit too complex. Would it be possible to get rid of this with function pointers (set to nil if the corresponding log level/functionality is disabled) or someting, WDYT?

pkg/log/backend.go
```diff
@@ -0,0 +1,240 @@
...
+// createFmtBackend creates an fmt Backend and starts its emitter goroutine.
+func createFmtBackend() Backend {
+	f := &fmtBackend{
+		q: make(chan *fmtReq, fmtBackendQueueLen),
+	}
+	go f.run()
+	return f
+}
+
+// Name returns the name FmtBackendName.
```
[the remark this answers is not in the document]
klihub: I'll remove the duplicates.

pkg/log/logger.go
```diff
@@ -0,0 +1,295 @@
...
+	// LevelInfo is the severity for informational messages.
+	LevelInfo
+	// LevelWarn is the severity for warnings.
+	LevelWarn
+	// LevelError is the severity for errors.
+	LevelError
+	// LevelFatal is the severity for fatal errors.
+	LevelFatal
+	// LevelPanic is the severity for panic messages.
+	LevelPanic
```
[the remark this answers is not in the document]
klihub: At least it was not my intention. Do you ask that because a panic() could be recover()ed form in some situations but you can't prevent os.Exit() from happening ? If you do, then I see your point and, yes, we should probably do that.

pkg/log/ratelimit.go
```diff
@@ -0,0 +1,124 @@
...
+	defer rl.Unlock()
+
+	msg := fmt.Sprintf(format, args...)
+	lim, ok := rl.limits[msg]
+
+	if !ok {
+		if len(rl.window) > 0 {
+			rl.window = rl.window[1:]
+		}
+		rl.window = append(rl.window, msg)
```
marquiz: What is the purpose of the window here and overall??

pkg/log/logger.go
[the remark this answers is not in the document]
marquiz: Yes, that was my thinking. Panic can probably be recovered (just a bug in some functionality) but Fatal cannot

pkg/log/backend.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+		if req.level == levelStop {
+			return
+		}
+	}
+}
+
+// emit formats and emits a single log message.
+func (f *fmtBackend) emit(req *fmtReq) {
+	if req.level > LevelPanic {
```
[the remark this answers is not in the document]
klihub: Changed.

pkg/log/flags.go
```diff
@@ -17,328 +17,324 @@ package log
...
 type options struct {
-	Level  Level       // lowest unsuppressed severity
-	Logger backendName // name of active backend
-	Enable stateMap    // enabled logger sources, all if nil
-	Debug  stateMap    // enable debug flags, all if nil
+	// Level is the logging severity/level.
+	Level Level
+	// Enable is a map for enabling/disabling normal logging for sources.
+	Enable srcmap
+	// Enable is a map for enabling/disabling debug logging for sources.
```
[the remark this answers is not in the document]
klihub: Fixed.

pkg/log/backend.go
```diff
@@ -0,0 +1,240 @@
...
+	levelStop
+)
+
+// severity tags fmtBackend uses to prefix emitted messages with.
+var fmtTags = map[Level]string{
+	LevelDebug: "D: ",
+	LevelInfo:  "I: ",
+	LevelWarn:  "W: ",
+	LevelError: "E: ",
+	LevelFatal: "Fatal error: ",
```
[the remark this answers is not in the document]
klihub: Changed to all CAPS.

pkg/log/logger.go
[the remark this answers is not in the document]
klihub: Swapped LevelFatal and LevelPanic.

pkg/log/flags.go
```diff
@@ -17,328 +17,324 @@ package log
...
+	Debug srcmap
+	// name of the logger backend to use
```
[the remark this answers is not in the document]
klihub: Fixed.

pkg/log/flags.go
```diff
@@ -17,328 +17,324 @@ package log
...
-		case "off", "disable", "disabled":
-			state = false
-		default:
-			if state, err = strconv.ParseBool(status); err != nil {
-				return loggerError("invalid state '%s' in spec '%s': %v", status, value, err)
-			}
+		if state != "" {
+			prev = state
+		}
+		if state == "" {
```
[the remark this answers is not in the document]
klihub: Updated.

pkg/log/flags.go
```diff
@@ -17,328 +17,324 @@ package log
...
-			switch f {
-			case "all", "*":
-				(*m)["*"] = state
-			case "none":
-				(*m)["*"] = !state
-			default:
-				(*m)[f] = state
-			}
+		switch strings.ToLower(state) {
+		case "on", "true", "enable":
```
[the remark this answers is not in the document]
klihub: Added pkg/utils.ParseEnabled() and switched to using it.

pkg/log/flags.go
```diff
@@ -17,328 +17,324 @@ package log
...
+					src = "*"
+				}
+				switch strings.ToLower(state) {
+				case "on", "enable", "true":
+					(*m)[src] = true
+				case "off", "disable", "false":
+					(*m)[src] = false
+				default:
+					return loggerError("source '%s' has invalid state '%s' in logger source map",
+						src, state)
```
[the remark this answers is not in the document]
klihub: Switched likewise to the newly added pkg/utils.ParseEnabled().

pkg/log/flags.go
```diff
@@ -17,328 +17,324 @@ package log
...
+	}
+
+	return loggerError("failed to unmarshal logger source map '%s': %v",
+		string(raw), err)
 }
 
-func (o *options) sourceEnabled(source string) bool {
-	return o.Enable.isEnabled(source)
+// copy enabled sources to options
+func (o *options) copyEnabled(m srcmap) {
```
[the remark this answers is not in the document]
klihub: Added srcmap.copy() and replaced calls to any of these with that.

pkg/log/log.go
```diff
@@ -16,394 +16,279 @@ package log
...
+	if !ok {
+		return loggerError("can't activate unknown backend '%s'", name)
+	}
 
-// Block emits a block of messages with the given emitting/printing function.
-func Block(fn func(string, ...interface{}), prefix string, format string, args ...interface{}) {
-	defLogger.Block(fn, prefix, format, args...)
-}
+	log.active.Stop()
+	log.active = createFn()
```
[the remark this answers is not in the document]
klihub: AFAICT, it's already called with log locked.

pkg/log/log.go
```diff
@@ -16,394 +16,279 @@ package log
...
-		fn("%s%s", prefix, line)
-	}
+	return log.create(source)
 }
 
-// Emit a block of debug messages.
-func (l *logger) DebugBlock(prefix string, format string, args ...interface{}) {
-	if !l.debug {
-		return
+// create creates a new logger for a source (call write-locked).
```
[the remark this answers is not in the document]
klihub: I changed all these locking comment to "should be called with log write-locked", "should be called with log read-locked", "read-locks log", and "write-locks log".

pkg/log/log.go
```diff
@@ -16,394 +16,279 @@ package log
...
+// logging encapsulates the full runtime state of logging.
+type logging struct {
+	sync.RWMutex
+	loggers map[string]logger    // source to logger mapping
+	sources map[logger]string    // logger to source mapping
+	configs map[logger]config    // logger configuration
+	maxname int                  // longest enabled/tracing source name
+	level   Level                // logging severity level
+	disable srcmap               // logging source configuration
+	tracing srcmap               // tracing source configuration
```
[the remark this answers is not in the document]
klihub: Replaced all instances of trace/tracing with debug/debugging.

pkg/log/ratelimit.go
klihub: To rate limit per message content. Not just per source.

klihub:
> What is `pkg/old-log` for?

Oops. Was committed accidentally. Removed.

pkg/log/ratelimit.go
```diff
@@ -0,0 +1,124 @@
...
+	sync.Mutex
+	rate   Rate
+	window []string
+	limits map[string]*goxrate.Limiter
+}
+
+const (
+	// DefaultWindow is the default message window size for rate limiting.
+	DefaultWindow = 256
+	// MinimumWindow is the smalled message window size for rate limiting.
```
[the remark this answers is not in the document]
klihub: Updated.

pkg/log/logger.go
klihub: I'd rather not. IMO, it's rather subjective/questionable whether `if foo.fn != nil` is any clearer/simpler than `if foo.isDebugging()`.

pkg/log/ratelimit.go
klihub: Maybe a test case would have been a good idea...

marquiz: Yeah, I kind of got that impression from some of the comments. I just don't undestand at all the logic here and how `window` is wired to the rate limiting logic.

ipuustin:
I'm thinking the same as @marquiz . Let's merge this now to have functional logging, but IMHO we should consider eventually spinning this off into a side project or using some external logging library, just to make the maintenance simpler (or rather someone else's concern :-). There a list of golang logging components here: https://www.programming-books.io/essential/go/logging-libraries-ead0070067424993bb40c754426bbb58

pkg/log/logger.go
marquiz: That was just a quick thought/gut feeling. And yes, the nil pointer was probably a bad idea. An alternative would be using function pointers and setting the pointer to point to a stub when disabling logging. That could possibly eliminate all if's and hopefully the config type.

But, no need to get stuck with this here. I think I can live with this 😄 We could create a new issue (assigned to me?) about this and try to address this separately.

pkg/log/ratelimit.go
klihub: Hmm... true. I don't understand either how that was supposed to work. Should be fixed now.

pkg/log/log.go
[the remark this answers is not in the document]
marquiz: Yes, true

pkg/log/ratelimit.go
marquiz: You haven't pushed the fixed version yet(?)

klihub: It should be there now.

pkg/log/ratelimit.go
```diff
@@ -0,0 +1,127 @@
...
+	if ok {
+		return limit
+	}
+
+	limit = goxrate.NewLimiter(rl.rate.Limit, rl.rate.Burst)
+	if len(rl.limits) == rl.rate.Window {
+		delete(rl.limits, rl.window[0])
+		rl.window = append(rl.window[1:], msg)
+	}
+	rl.window = append(rl.window, msg)
```
marquiz: Shouldn't you only append `msg` once? Now, you you append twice and let `rl.window` to outgrow `rl.rate.Window`

klihub: Ugh... *facepalm*

PR #169, opened 2020-02-14
build against kubernetes v1.17.2

klihub:
Switch to building against kubernetes v1.17.2.


klihub:

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -201,6 +201,13 @@ func (m *resmgr) setupCache() error {
...
 	var err error
 
+	if opt.ForceConfig != "" {
+		if opt.AgentSocket != "" {
+			m.Warn("using forced configuration %s, disabling agent", opt.ForceConfig)
+			opt.AgentSocket = agent.DontConnect
+		}
+	}
+
 	if m.agent, err = agent.NewAgentInterface(opt.AgentSocket); err != nil {
```
ipuustin: It seems bit strange that there's only a facility for disabling the agent. Why do we create a non-working interface in the first place, if there is no chance of anyone enabling it later?

kad: it looks like that this PR has more commits than initially intended. @klihub  can you clean-up, so we can move to 1.17.2 without other changes?

klihub: > it looks like that this PR has more commits than initially intended. @klihub can you clean-up, so we can move to 1.17.2 without other changes?

Yes, it was based on #168 . Rebased on latest master, chopped down to minumal commits.

klihub:
Let me add one more commit to this, after lunch...

pkg/cri/resource-manager/resource-manager.go
klihub: > It seems bit strange that there's only a facility for disabling the agent. Why do we create a non-working interface in the first place, if there is no chance of anyone enabling it later?

We always want to provide a non-nil agent interface among the policy.Options for policy backends. If the agent can't be used ATM, it is `Disabled()` instead of being `nil`. Whether the agent can later be(come) enabled is an implementation detail.

klihub:
> Let me add one more commit to this, after lunch...

@kad On a second thought I rather file another PR for the thing(s) I had in mind (to allow disabling bits in instrumentation piece by piece). It'd become too big for this.

pkg/cri/resource-manager/resource-manager.go
klihub: Split out agent disabling bits from this PR.

PR #172, opened 2020-02-17
[rebalancing] resource-manager: save and proxy metrics to prometheus.

klihub:
This patch set adds the last missing bits from the original rebalancing PoC. It implements proxying raw metrics to prometheus.

The patch set is split in two. One implements the actual gathering vs. proxying bits in the resource-manager. The other refactors the instrumentation code, combining most of the instrumentation runtime state into an instrumentation.Service object, and adds proper locking to it.


pkg/instrumentation/instrumentation.go
```diff
@@ -15,70 +15,63 @@
...
 
-// Stop shuts down instrumentation.
+// Stop stops our internal instrumentation services.
 func Stop() {
-	unregisterJaegerExporter()
-	unregisterGrpcTraceViews()
-	unregisterPrometheusExporter()
-	HTTPShutdown()
+	if service != nil {
+		service.Stop()
```
marquiz: should we log the possible error, just in case?

pkg/instrumentation/jaeger.go
```diff
@@ -0,0 +1,58 @@
...
+		return instrumentationError("failed to create Jaeger exporter: %v", err)
+	}
+
+	return nil
+}
+
+// startJaegerExporter 'starts' the Jeager exporter by registering it.
+func (s *Service) startJaegerExporter() {
+	if s.jexport != nil {
+		if opt.Trace != Disabled {
```
marquiz: nit: you could drop one level of indentation by using `&&` (?)

pkg/instrumentation/prometheus.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
marquiz: Copyright year 2020

pkg/instrumentation/prometheus.go
```diff
@@ -0,0 +1,101 @@
...
+		return instrumentationError("failed to create Prometheus exporter: %v", err)
+	}
+
+	return nil
+}
+
+// startPrometheusExporter registers and starts the Prometheus exporter.
+func (s *Service) startPrometheusExporter() {
+	if s.pexport != nil {
+		if opt.Trace != Disabled {
```
marquiz: nit: combine the nested ifs with `&&`?

pkg/instrumentation/service.go
```diff
@@ -0,0 +1,136 @@
...
+	log.Info("starting instrumentation service...")
+
+	s.createHTTP()
+	s.startJaegerExporter()
+	s.startPrometheusExporter()
+
+	if err := s.registerGrpcViews(); err != nil {
+		s.stopJaegerExporter()
+		s.stopPrometheusExporter()
+		s.stopHTTP()
```
marquiz: Do we want to stop http when that's not started yet?

pkg/instrumentation/instrumentation.go
klihub: Changed error handling to be more consistent. Now
  - stopping a non-existent service is not an error
  - starting/restarting a non-existent service is an error
  - the caller is expected to deal with/log the top-level errors
Fixed a couple of potential crashes for service == nil (Restart(), GetHTTPMux(), ...).

instrumentation.s

pkg/instrumentation/jaeger.go
klihub: Fixed.

pkg/instrumentation/prometheus.go
klihub: Fixed (and the others, too).

pkg/instrumentation/prometheus.go
klihub: Fixed.

pkg/instrumentation/service.go
klihub: It was bit of a misnomer, so renamed to closeHTTP(). It both does a http.Server.Close()'s and lets the mux/server go by resetting them to nil (which is what we want here).

PR #184, opened 2020-02-28
introspection+visualization: add introspection HTTP interface, container placement visualization and a PoC UI

klihub:
This PR needs (IOW is rebased on top of) #233 .

This patch series adds initial support for external introspection of the current state of the resource manager. It also introduces a package for serving self-hosted container placement/resource allocation visualization UIs, with an initial/proof-of-concept UI implementation ('bubbles') by Arnaud Mandy.

Introspection data is served over the instrumentation metrics HTTP endpoint using the '/introspect' path. The data is JSON and currently contains:
  - Pods (name, ID, containers)
  - Containers (name, IDs, command/args, resource requests, topology hints)
  - Pools (name, CPUs, memory, parent/children pools)
  - Assignments (container ID, pool, CPU, memory, RDT class, Block I/O class)
  - System info (HW: sockets, NUMA nodes, isolated&offline CPUs, RDT classes, active policy)

In it's current incarnation the patch set fetches pools and assignments from the active policy while other data is provided by the common agnostic policy layer. Consequently, new corresponding Introspect() functions have been added to the policy and backend interfaces as well as to the builtin policy implementations (only the topology-aware one provides real data FTTB). The reason for this split of duty is that we have not standardized/unified the data structure(s) for resource assignment in the cache.

The external introspection interface currently only supports fetching a full snapshot of the current state without providing any means to easily track changes. This is true both for the internal interface which provides data for introspection and the external HTTP introspection interface.

Workload placement visualization Web UIs are served over the instrumentation metrics HTTP endpoint using the '/ui' path. In particular for visualization this PR adds
  - support for serving UI collateral over HTTP
  - embedding UI collateral in the binary using [vfsgen](https://github.com/shurcooL/vfsgen)
  - support for serving external UIs (`--external-visualizers <dirs>` option)
  - a first PoC UI implementation (`bubbles`), courtesy of Arnaud Mandy
  - command line option (`--disable-ui`) for disabling UIs

The UIs use the introspection interface of `cri-resmgr` to fetch up-to-date data and visualize it.
Currently only the `topology-aware` policy produces any real introspection data.

You can give it a test drive by starting up `cri-resmgr`, scheduling a few workloads then pointing your browser at `http://<host-running-cri-resmgr>:8888/ui/bubbles`. Currently `bubbles` cannot refresh itself, so you need to manually reload the UI to see any changes reflected.


klihub:

marquiz:
I tried to fix the RDT interface problem (also addressed in this PR) in #208

klihub:
> I tried to fix the RDT interface problem (also addressed in this PR) in #208

Thanks, I rebased this on latest master with #208 merged.

pkg/cri/resource-manager/visualizer/flags.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// dirList is a comma-separated list of directory names to search for visualizers.
+type dirList string
+
+// Set registers external visualizer implementations found in the directory list.
+func (d *dirList) Set(value string) error {
+	for _, dir := range strings.Split(value, ",") {
+		log.Info("searching for external visualizers in %s...", dir)
+		visualizers.discoverExternal(dir)
```
marquiz: Just a comment: I think this is fine for now but I think it would be nicer to separate the discovery from flag parsing. But I guess that can be done later if we want to allow "hot-plugging" of visualizers, for example

pkg/cri/resource-manager/visualizer/visualizer.go
```diff
@@ -0,0 +1,213 @@
...
+
+	switch {
+	case len(split) > 1 && kind == "internal":
+		dir, ok = v.internal[split[1]]
+	case len(split) > 1 && kind == "external":
+		dir, ok = v.external[split[1]]
+	default:
+		ok = false
+	}
+
```
marquiz: Could be simplified with smth like:
```
	ok = false
	if len(split) > 1 {
		kind := split[0]
		switch kind {
		case "internal":
			dir, ok = v.internal[split[1]]
		case "external":
			dir, ok = v.external[split[1]]
		}
	}
```
WDYT?

pkg/cri/resource-manager/visualizer/visualizer.go
```diff
@@ -0,0 +1,213 @@
...
+		name := v.uniqueExternalName(dir)
+
+		log.Debug("found external visualizer '%s' (%s)", name, dir)
+		v.register(name, http.FileSystem(http.Dir(dir)), false)
+
+		return nil
+	})
+}
+
+// uniqueExternalName generate a unique name for the external visualizer.
```
marquiz: s/generate/generates/

klihub:
> Thanks for the PR. Some minor comments, but, looks good.
> 
> I didn't read through the .js stuff, I'm sure it's fine

Heh... The JS stuff is a rough proto and luckily we have other folks from the Web team working on it. IMO, its biggest functional problem ATM is that it can't refresh itself.

pkg/cri/resource-manager/policy/builtin/topology-aware/topology-aware-policy.go
```diff
@@ -233,6 +234,50 @@ func (p *policy) ExportResourceData(c cache.Container) map[string]string {
...
+			ExclusiveCPUs: g.ExclusiveCPUs().Union(g.IsolatedCPUs()).String(),
+			RDTClass:      g.GetContainer().GetRDTClass(),
+			BlockIOClass:  g.GetContainer().GetBlockIOClass(),
+			Pool:          g.GetNode().Name(),
+		}
+		if g.SharedPortion() > 0 || a.ExclusiveCPUs == "" {
+			a.SharedCPUs = g.SharedCPUs().String()
+		}
+		if a.RDTClass == "" {
+			a.RDTClass = "<root>"
```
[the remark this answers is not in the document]
klihub: Hmm, true. The general idea is that we set this to what class the policy assigned the container to. Where that is mapped is then another story. We'll probably deal with that by simply including the RDT class map in the introspection data. Ditto for Block I/O. It's then up to the visualization part how to show the effective class if it's not the set one. But you're right, better not change the class here. And we have to move the logic of defaulting to the QoS class to the cache.

pkg/cri/resource-manager/policy/policy.go
```diff
@@ -274,6 +282,89 @@ func (p *policy) ExportResourceData(c cache.Container) {
...
+		sys.Isolated = p.system.Isolated().String()
+		sys.Offlined = p.system.Offlined().String()
+		p.inspsys = sys
+	}
+
+	rdtClassNames := []string{}
+	for _, rdtClass := range rdt.GetClasses() {
+		rdtClassNames = append(rdtClassNames, rdtClass.Name())
+	}
+	p.inspsys.RDTClasses = rdtClassNames
```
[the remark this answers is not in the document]
klihub: IMO by including eventually the controller class maps in the introspection data. It will be then up to the visualization UI to decide how to show the case where the set class is not the effective one.

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -49,17 +51,18 @@ type ResourceManager interface {
...
+	policy       policy.Policy      // resource manager policy
+	configServer config.Server      // configuration management server
+	control      control.Control    // policy controllers/enforcement
+	agent        agent.Interface    // connection to cri-resmgr agent
+	conf         *config.RawConfig  // pending for saving in cache
+	metrics      *metrics.Metrics   // metrics collector/pre-processor
+	events       chan interface{}   // channel for delivering events
+	stop         chan interface{}   // channel for signalling shutdown to goroutines
+	signals      chan os.Signal     // signal channel
+	ispect       *introspect.Server // server for external introspection
```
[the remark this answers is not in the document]
klihub: Updated.

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -430,3 +440,40 @@ func (m *resmgr) startControllers() error {
 
 	return nil
 }
+
+// setupIntrospection prepares the resource manager for serving external introspection requests.
+func (m *resmgr) setupIntrospection() error {
+	if policy.ActivePolicy() == policy.NullPolicy {
+		return nil
+	}
```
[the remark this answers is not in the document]
klihub: Moved bypassing active policy introspection in case of null policy is now done in the policy layer, so we don't need to care here.

pkg/cri/resource-manager/visualizer/visualizer.go
klihub: True. Updated.

pkg/cri/resource-manager/visualizer/visualizer.go
klihub: Fixed.

pkg/cri/resource-manager/visualizer/flags.go
klihub: Hmm, you mean we'd try to look discover external visualizers on demand, without having to restart cri-resmgr ? That's a good idea. I'll update the code to do that.

marquiz: Yeah, that was my thinking. Cool, if you're implementing that 👍

klihub: @marquiz I pushed a version implementing on-demand external UI discovery.

The current code does the discovery without caching any results. When an external UI is chosen with the normal browsing flow (-> index page -> select UI) we do discovery twice: once to generate the index page, then once when the external UI is selected. Caching the results should be easy to add if it is considered necessary.

marquiz: Cool, I think this is enough or now.

pkg/cri/resource-manager/policy/policy.go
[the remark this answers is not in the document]
klihub: > IMO by including eventually the controller class maps in the introspection data. It will be then up to the visualization UI to decide how to show the case where the set class is not the effective one.

This is now tracked separately in #232 .

klihub:
Removed RDT and BlockIO classes from introspection data, until #232 is done.
Rebased on top of #233 which is necessary for successfully running `make vet`.

PR #186, opened 2020-02-28
system: change system structs to interfaces

ipuustin:
Make the `sysfs` package use interfaces instead of plain objects so we can mock them in tests. Use `CPUPackage` (and corresponding type `cpuPackage struct`) because "package" is a reserved word in Go. Also add `DiscoverSystemAt()` method for testing with fake sysfs directory hierarchy.

Note: relates to PR https://github.com/intel/cri-resource-manager/pull/90.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/static-plus/static-plus-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/static-plus/static-plus-policy.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"strconv"
 
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
@@ -62,7 +62,7 @@ type staticplus struct {
 	reserved    cpuset.CPUSet // pool (primarily) for system-/kube-tasks
 	isolated    cpuset.CPUSet // primary pool for exclusive allocations
 	allocations Allocations   // container cpu allocations
-	sys         *sysfs.System // system/topologu information
+	sys         sysfs.System  // system/topologu information
 	cache       cache.Cache   // system state/cache
 	shared      cpuset.CPUSet // pool for fractional and shared allocations
 }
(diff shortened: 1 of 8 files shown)


pkg/cri/resource-manager/policy/builtin/static-plus/static-plus-policy.go
```diff
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"strconv"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
klihub: nitpick: I'd understand the explicit naming if you called it corev1 to be in line with metav1, but I fail to see any benefit here if it's explicitly named the same as the default.

pkg/cri/resource-manager/policy/builtin/static/static-policy.go
```diff
@@ -18,7 +18,7 @@ import (
 	"fmt"
 	"strconv"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
klihub: nitpick: same nitpick as above...

pkg/sysfs/system.go
```diff
@@ -54,37 +54,97 @@ const (
...
+	CPUCount() int
+	NUMANodeCount() int
+	ThreadCount() int
+	CPUSet() cpuset.CPUSet
+	Package(id ID) CPUPackage
+	Node(id ID) Node
+	CPU(id ID) CPU
+	Offlined() cpuset.CPUSet
+	Isolated() cpuset.CPUSet
+	discoverCache(path string) error
```
klihub: Do these unexported functions really need to be part of the interface ? I think there is no reason to force any internal mocked test implementation to structurally mimick the discovery code so closely. And for externally mocked tests they do not matter as they are not visible.

pkg/sysfs/system.go
```diff
@@ -95,6 +155,8 @@ type CPU struct {
 	isolated bool    // whether this CPU is isolated
 }
 
+var _ CPU = &cpu{}
```
klihub: I think this construct does not produce any benefit here, since both CPU and cpu are part of this same package. Any compile-time error caught by this would be caught anyway, since there must be a function returning a &cpu as a CPU (otherwise having CPU wouldn't make sense with only the single cpu implementing it).

pkg/cri/resource-manager/policy/builtin/static-plus/static-plus-policy.go
ipuustin: They seem to come automagically from golint... fixed them to `corev1` to keep the tooling happy.

pkg/sysfs/system.go
ipuustin: Right, removed them.

pkg/sysfs/system.go
ipuustin: Ok, makes sense. Fixed.

PR #190, opened 2020-03-06
Rudimentary support for CPU prioritization

marquiz:
This patch adds support for per core priority in the cpu allocator used
by static, static-plus and topology-aware built-in policies. At this
point there is only support for two stages of priority (low and high).
The new allocator heuristics added in this patch are straightforward:
depending on the caller's guidance, the allocator always simply prefers
or tries to avoid high-priority CPUs.

Detection of high priority CPUs is currently based on the base_frequency
attribute. If CPUs with different base frequencies are detected, all
higher frequency CPUs (higher than the lowest detected base frequency)
are flagged as high priority.

### DIFF
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -18,7 +18,6 @@ import (
 	"flag"
 	"fmt"
 	"sort"
-	"sync"
 
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 
@@ -49,9 +48,9 @@ type CPUAllocator struct {
 	sys           sysfs.System  // sysfs CPU and topology information
 	flags         AllocFlag     // allocation preferences
 	from          cpuset.CPUSet // set of CPUs to allocate from
+	preferred     cpuset.CPUSet // set of preferred CPUs
 	cnt           int           // number of CPUs to allocate
 	result        cpuset.CPUSet // set of CPUs allocated
-	offline       cpuset.CPUSet // set of CPUs currently offline
 
 	pkgs []sysfs.CPUPackage // physical CPU packages, sorted by preference
 	cpus []sysfs.CPU        // CPU cores, sorted by preference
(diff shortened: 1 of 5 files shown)


klihub:
Overall this looks good to me. The only thing which I was pondering (as we  discussed) when reading the code is how to use the preferred set:

  - use the original algorithm, first confined to the full preferred set then falling back to the full set, or
  - take the preferred set into account when picking/prioritizing cores/threads

This implementation is the latter one.

The original allocator code seems a bit stupid, because in some cases when it could fit a workload within one socket it will instead spread it across multiple ones (takeIdleCores and takeIdleThreads should sort cores and threads primarily by the amount of cores/threads free in their socket and only secondarily by core/thread ID). Also it does not care about NUMA nodes even when SNC is enabled. These we should fix eventually. Also it might make sense to provide some way to pass in preferences (or override the core/thread sorting logic by the caller) if the number of parameters we want to take into account increases.

pkg/cpuallocator/allocator.go
```diff
@@ -366,30 +420,43 @@ func allocateCpus(from *cpuset.CPUSet, cnt int) (cpuset.CPUSet, error) {
...
+func ReleaseCpus(from *cpuset.CPUSet, cnt int, preferHighPrio bool) (cpuset.CPUSet, error) {
 	var oset cpuset.CPUSet
 
 	if debug {
 		oset = from.Clone()
 	}
 
-	result, err := allocateCpus(from, from.Size()-cnt)
+	preferred := system.priorityCpus
+	if !preferHighPrio {
```
klihub: nitpick: not that it would matter because none of the policies do partial releases of CPUs... but if the logic by the caller for deducing preferHighPrio for AllocateCpus is to take it from an optional annotation, then the caller needs to invert that logic here. IOW, if it allocates with preferHighPrio v, then it would need to pass !v to ReleaseCpus for shrinking the container.

marquiz: Yeah, that's true. Any suggestions how to fix the logic/naming? I seem to be quite blind to that at this time of the week and day 😄 I think it sort of does what it "promises" i.e. releases high prio cpus if `preferHighPrio` is true. But OTOH, I understand the inherent risk of confusion, too

marquiz:
Mm, I played with different variants and ended up with this eventually which prefers full cores over single high priority hyperthreads. Effectively, in practice it tries harder to pack the workload onto fewer numa nodes, instead of spreading it out to high priority hyperthreads all over.

The best strategy is subject to discussion and also depends on the workload. I think we could merge whatever seems reasonable here and fix/improve the allocator logic in #191.

pkg/cpuallocator/allocator.go
klihub: I'm really not sure if it can be confusing for anybody else than myself. I mentioned just in case somebody else might want take a look at it and see how they feel. I honestly suspect that this really might be a neurotic personality disorder vs. normal human beings thing. Let's not touch this unless somebody else gets confused.

klihub:
> The best strategy is subject to discussion and also depends on the workload. I think we could merge whatever seems reasonable here and fix/improve the allocator logic in #191.

I agree. Makes sense.

Issue #175, opened 2020-02-21
Documentation overhaul


marquiz:
We need to go through all the documentation, restructure it and supplement missing areas, before the next release. Motivated by #171 

Backlog (please append new items as you encounter areas to fix):
- [ ] #321 README should provide references to all the sub-documents
- [ ] #320 update RDT documentation, better describing partitions, classes, configuration, exclusivity etc.
- [ ] #121 (Documentation: “how to write policies”)
- [x] #159 (Documentation about how to start cri-resmgr is outdated)
- [ ] #171 (Readme Improvements)
- [ ] #318 document (per policy) handling of isolcpus

msivosuo:
Related issues: #171, #159 , #121

marquiz:
I spent some time on this and decided to use [Github Pages](https://help.github.com/en/github/working-with-github-pages). I think that would allow us much more flexible and easier to navigate documentation.

I did some experimentation with [Hugo](https://gohugo.io/), an easy-to-use static site generator and was quite happy with it. With a minimal effort I was able to create a clean and easily navigable documentation wireframe. It's very nice as you can basically just write markdown files like before. The only overhead is publishing the content, but, even that should be relatively easy to automate with e.g. github workflows (I already created some scripting for that).

You can check out the result on my personal github fork:
https://marquiz.github.io/cri-resource-manager/

My [devel/documentation-overhaul/ghpages](https://github.com/marquiz/cri-resource-manager/tree/devel/documentation-overhaul/ghpages) branch contains the source

WDYT?

marquiz:
My experiment in https://marquiz.github.io/cri-resource-manager/ also (kind-of-implicitly) contains an initial draft of a possible documentation hierarchy/structure. Please comment that, too

klihub:
I have extremely limited experience with this, so this is more of a note than an opinion... I personally ended up using Jekyll for github.io. My unscientific algorithm for choosing that was googling, which gave me the *impression* that
  - Jekyll is the 'natively supported' choice for GitHub pages, and
  - the normal combos seemed to be GitHub pages+Jekyll and Netlify+Hugo

That said, I use Jekyll for occasionally experimenting with a github-based blog and not for a repo-related 'site'. If Hugo is as simple and does everything we need, I'm sure we can use that too.

klihub:
[...]

PR #192, opened 2020-03-06
static-policy: drop the implicit logic of allocating isolcpus

marquiz:
Stop allocating isolated cores if the user requests for exactly two
logical cpus. Instead, allocate ordinary cores similarly to other cases.
The removed implicit logic was very unintuitive, assuming the user to
know/understand the internals of cri-resmgr, plus, giving varying
results e.g. depending on whether hyper-threading was enabled or not.


klihub:
I'd be fine with this change. @kad How about you ?

kad:
I'm ok with that, however, I have a feeling that we need to have documentation about handling kernel isolcpus and table with policies/conditions when/how it will be allocated. Otherwise we will forget that ourselves.

marquiz:
> I have a feeling that we need to have documentation about handling kernel isolcpus and table with policies/conditions when/how it will be allocated. Otherwise we will forget that ourselves.

I added an item to the tasklist in #175

PR #195, opened 2020-03-09
cpuallocator: drop '-cpu-allocator-debug' command line flag

marquiz:
It was mostly unused, anyway. Just use normal debug logging.


marquiz:
> LGTM, once that build failure is fixed.

Uh-oh, I had fatfingered something when pushing the code. Should be fixed, now

PR #196, opened 2020-03-09
memtier: add policy prototype

klihub:
A 'memory tiering' policy, currently in prototype/WiP state.


ipuustin:
@klihub: use this as the "memtier: fix tests." commit to fix the tests: https://github.com/ipuustin/cri-resource-manager/commit/6f391fb60f64633e3b12bd86b8098e567a89e8b4

klihub:
> @klihub: use this as the "memtier: fix tests." commit to fix the tests: [ipuustin@6f391fb](https://github.com/ipuustin/cri-resource-manager/commit/6f391fb60f64633e3b12bd86b8098e567a89e8b4)

@ipuustin Updated with that commit.

ipuustin:
Next thing I'm working on: conservative memory reservation from NUMA nodes to prevent OOM situations. If a container is allocated on a second-level node, we need to assume that the memory is allocated as a worst-case scenario from both nodes. Then containers should be moved upwards in the NUMA tree until the memory allocation is safe again.

ipuustin:
I think the key questions deal with how the memory allocation logic is done. Currently the containers are able to provide a bitmask telling which memory types they would like to have. This makes sense from the memory controller point of view, but now when we need to think of allocating memory bytes (due to the extra reservation scheme mentioned above), it gets bit tricky, especially when we can't really tell what kind of memory the container will actually get when it does the allocations.

It might be better to have a containers request "minimum speed class", bit like QoS. Then we could just reserve "at least this fast memory". Thoughts?

klihub:
> I think the key questions deal with how the memory allocation logic is done. Currently the containers are able to provide a bitmask telling which memory types they would like to have. This makes sense from the memory controller point of view, but now when we need to think of allocating memory bytes (due to the extra reservation scheme mentioned above), it gets bit tricky, especially when we can't really tell what kind of memory the container will actually get when it does the allocations.
> 
> It might be better to have a containers request "minimum speed class", bit like QoS. Then we could just reserve "at least this fast memory". Thoughts?

I think that for now we should assume that under normal conditions containers will fit into their assigned pool/NUMA nodes and the algorithm to relax/expand the memory pinning will not need to kick in. So, for the time being, I'd primarily tweak the memtier policy for the ideal case: enough memory and containers requesting/pinned to a single type of memory. In that case, it is not true that we don't know what kind of memory the containers are going to get, at least not for containers that explicitly request a particular type of memory.

In the fallback case where all of these assumptions are not true, treating faster memory than requested as the requested one is a reasonable assumption... at least for PMEM/DRAM. I'm not so sure about HBMEM. AFAIK, usually it's a really scarce resource dedicated to special workloads. It might be a better idea to fail a container which does not explicitly request HBMEM then to give it HBMEM and later fail a special container that really could use it. Or at least make this configurable as a policy config flag, so the admin can easily and clearly indicate if HBMEM will be requested and should be reserved for special applications.

pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
```diff
@@ -0,0 +1,290 @@
...
+// memoryType is bitmask of types of memory to allocate
+type memoryType int
+
+// memoryType bits
+const (
+	memoryUnspec memoryType = 0x0
+	memoryDRAM   memoryType = 0x1
+	memoryPMEM   memoryType = 0x2
+	memoryHBMEM  memoryType = 0x4
+	memoryAll    memoryType = 0x7
```
kad, line 63: not: https://golang.org/ref/spec#Iota, use `1 << iota` style bitmasks definitions.

pkg/sysfs/system.go
```diff
@@ -53,6 +54,16 @@ const (
...
 )
 
+// MemoryType is an enum for the Node memory
+type MemoryType int
+
+const (
+	// MemoryTypeDRAM means that the node has regular DRAM-type memory
+	MemoryTypeDRAM MemoryType = iota
+	// MemoryTypePMEM means that the node has persistent memory
+	MemoryTypePMEM
```
kad, line 64: hbmem?

pkg/sysfs/system.go
```diff
@@ -614,12 +628,39 @@ func (c *cpu) SetFrequencyLimits(min, max uint64) error {
...
+	return cpuset.Parse(strings.Trim(string(blob), "\n"))
+}
+
 // Discover NUMA nodes present in the system.
 func (sys *system) discoverNodes() error {
 	if sys.nodes != nil {
 		return nil
 	}
 
+	// FIXME assumption: if a node only has memory (and no CPUs), it's PMEM. Otherwise it's DRAM.
```
kad, line 648: we need to check size. If $amount is < than DRAM, then it is HBMEM. if > DRAM then PMEM.

kad:
> I think the key questions deal with how the memory allocation logic is done. Currently the containers are able to provide a bitmask telling which memory types they would like to have. This makes sense from the memory controller point of view, but now when we need to think of allocating memory bytes (due to the extra reservation scheme mentioned above), it gets bit tricky, especially when we can't really tell what kind of memory the container will actually get when it does the allocations.
> 
> It might be better to have a containers request "minimum speed class", bit like QoS. Then we could just reserve "at least this fast memory". Thoughts?

RIght now, the biggest problem is DRAM amounts tracking. If for workload DRAM+PMEM is enabled, kernel will start migrate pages to PMEM if there is pressure on available DRAM. The tricky part is DRAM only containers in this node - their pages can't be migrated.

pkg/sysfs/system.go
ipuustin: That's another assumption, which has it's own problems...

A proper way to do this would be to go through the ACPI tree and parse the entries there. or use existing tooling which does the same. I have the feeling that traversing the ACPI tree approach is also tricky, because we don't know beforehand the ACPI tree layout of each and every system, and testing it widely enough is hard. Using existing tooling would give us an  external dependency and possibly some headache with parsing tool output data and maintaining compatibility between versions.

I wrote the test below to find the PMEM nodes (and the NUMA nodes which map to them. It gives some idea of the potential complexity. I don't know if the ACPI tree files and their data layout is under the kernel<->userspace API stability promise...

```go
package main

import (
        "fmt"
        "io/ioutil"
        "os"
        "path"
        "path/filepath"
        "strconv"
        "strings"
)

const (
        acpiRoot = "/sys/devices/LNXSYSTM\\:00/"
)

type pmemregion struct {
        numaNode   int
        targetNode int
}

func main() {

        nodes := make([]pmemregion, 0)

        buses, _ := filepath.Glob(path.Join(acpiRoot, "LNXSYBUS\\:*"))
        for _, bus := range buses {
                devices, _ := filepath.Glob(path.Join(bus, "ACPI*"))
                for _, device := range devices {
                        ndbuses, _ := filepath.Glob(path.Join(device, "ndbus*"))
                        for _, ndbus := range ndbuses {
                                regions, _ := filepath.Glob(path.Join(ndbus, "region*"))
                                for _, region := range regions {
                                        buf, err := ioutil.ReadFile(path.Join(region, "numa_node"))
                                        if err != nil {
                                                fmt.Println("Error reading file")
                                                os.Exit(1)
                                        }
                                        numaNode, _ := strconv.ParseInt(strings.Trim(string(buf), "\n"), 0, 64)

                                        buf, err = ioutil.ReadFile(path.Join(region, "target_node"))
                                        if err != nil {
                                                fmt.Println("Error reading file")
                                                os.Exit(1)
                                        }

                                        targetNode, _ := strconv.ParseInt(strings.Trim(string(buf), "\n"), 0, 64)

                                        entry := pmemregion{numaNode: int(numaNode), targetNode: int(targetNode)}

                                        nodes = append(nodes, entry)
                                }
                        }
                }
        }

        for _, node := range nodes {
                fmt.Printf("Numa node: %d -> PMEM node %d\n", node.numaNode, node.targetNode)
        }
}
```

pkg/sysfs/system.go
ipuustin: So far only DRAM and PMEM were detected. I'll add HBMEM with the "alwasy-smaller-than-dram" assumption discussed below.

klihub:
@kad , @marquiz If @ipuustin considers this a good enough initial version for merging, could either of you ACK it. I can't do it since I filed the PR.

ipuustin:
I think this is in "good enough" shape. The problem domain is quite complex and the e2e testing could be better, but it doesn't have any known issues.

marquiz:
I'm basically ok with merging this as it's virtually a "self-contained" patchset, adding one built-in policy and not touching the other parts of the codebase.

My wish is that could the commit history be tidied? E.g. squash everything into one commit with a good/meaningful commit message.

Another comment regarding the test data: would a plain tar be better (than zipped) suited for git, WDYT?

klihub:
> I'm basically ok with merging this as it's virtually a "self-contained" patchset, adding one built-in policy and not touching the other parts of the codebase.
> 
> My wish is that could the commit history be tidied? E.g. squash everything into one commit with a good/meaningful commit message.
> 

Yes, definitely. Not  into a single commit, but a few sensibly grouped ones (for instance, I don't want to mix sysfs changes/additions with the actual memtier policy commit(s)).

klihub:
> Another comment regarding the test data: would a plain tar be better (than zipped) suited for git, WDYT?

Do you mean an uncompressed tarball ? TBH, I don't see why that would be any better (suited for git). IMO both are a bit of an abuse. But the real problem with the plain tar in this case, and why we decided to compress it, is that the size of the tarball would grow from 131K to 22M.

marquiz:
> Do you mean an uncompressed tarball ? TBH, I don't see why that would be any better (suited for git).

Yeah, forget that comment, I don't know what I was thinking 😬 


> IMO both are a bit of an abuse. But the real problem with the plain tar in this case, and why we decided to compress it, is that the size of the tarball would grow from 131K to 22M.

Yup, but I think we can live with this as the tarball isn't bigger than that.


One thing that I started to think is the location of test data. Should we have it in the same package as the code or separately? Maybe we should have a uniform convention – I put RDT test data under `test/data/rdt/`. WDYT?

ipuustin:
I'll try to squash my commits today. I can't force push them to this branch, but I'll put there available on some branch.

The test data question is a good one, and I gave it some thought when I was writing the tests. On the other hand we should probably make the sysfs test data available for all packages to avoid duplication and to make it easier to write basic tests for many machines ("parsing this data should not crash on any machine"). On the other hand, the sysfs filesystem is really big, and I only took the directories directly relevant to the memory/cpu testing in the test data archive. It could be that some other packages do not have all the data they need in the sysfs snapshots.

One solution might be to have a full sysfs snapshots, and move the data to live outside of git in some large static archive, which could then be downloaded for running "full test suite". Otherwise only a subset of tests would be run.

I used directory `testdata` because that's the "official" directory name for go test data. `go help test` says:

    The go tool will ignore a directory named "testdata", making it available
    to hold ancillary data needed by the tests.

klihub:
> I'll try to squash my commits today. I can't force push them to this branch, but I'll put there available on some branch.

I tried to come up with a squashed history with the minimal number of commits. It's pushed to this branch, so you can take a look to see if it's ok. Basically I split out some accidental sysfs bits that slipped into memtier commits and moved all of them them to the beginning of the series. I removed an unrelated comment typo fix in pkg/agent. Then squashed all the memtier bits into a single commit under your name.

I know it's not how it happened, so if you can come up with something simple enough that better reflects reality then just shoot...

klihub:
> > Do you mean an uncompressed tarball ? TBH, I don't see why that would be any better (suited for git).
> 
> Yeah, forget that comment, I don't know what I was thinking 
> 
> > IMO both are a bit of an abuse. But the real problem with the plain tar in this case, and why we decided to compress it, is that the size of the tarball would grow from 131K to 22M.
> 
> Yup, but I think we can live with this as the tarball isn't bigger than that.
> 
> One thing that I started to think is the location of test data. Should we have it in the same package as the code or separately? Maybe we should have a uniform convention – I put RDT test data under `test/data/rdt/`. WDYT?

Yes, I agree with consistency uber alles as a principle we should at least try to adhere to... About the exact location for test data, I'm fine with anything that doesn't upset any of the toolchain. If there is an existing prevalent/dominant go convention we should probably try to go with that.

marquiz:
> If there is an existing prevalent/dominant go convention we should probably try to go with that.

I don't know if there's something like that. But, at least the standard libraries seem to use a separate testdata directory under each package, e.g. [`encoding/json`](https://golang.org/src/encoding/json/testdata/)

Maybe we should follow the same convention? If so, I can create a PRs to move all the other test data accordingly.

ipuustin:
@klihub Thanks for the rebase. Memtier plugin looks good, but I think https://github.com/ipuustin/cri-resource-manager/commit/b8dac23eb47b50af0d48bd8988ef1ecbab62ae35 is missing from the sysfs patches? I think @kad wanted to have something like this already for HBMEM.

marquiz:
Let's go with the convention used here (and in the golang standard libraries)

ipuustin:
Using `testdata` under the individual packages is good because the tests can use relative paths to the test data (the test working directory is the package directory). The problem is the duplication of data if many packages need to access sysfs.

marquiz:
> Using `testdata` under the individual packages is good because the tests can use relative paths to the test data (the test working directory is the package directory).

I don't think that will make a difference. Paths can be made easy in a separate package as well. This is what I did for rdt: https://github.com/intel/cri-resource-manager/blob/master/test/data/rdt/rdt.go



> The problem is the duplication of data if many packages need to access sysfs.

That could be a problem but otoh it's good that the packages don't have inter-dependencies via test data and it can be changed at will

klihub:
> @klihub Thanks for the rebase. Memtier plugin looks good, but I think [ipuustin@b8dac23](https://github.com/ipuustin/cri-resource-manager/commit/b8dac23eb47b50af0d48bd8988ef1ecbab62ae35) is missing from the sysfs patches? I think @kad wanted to have something like this already for HBMEM.

Is it ok with you if I squash that into [13a346adca81e84e10c70733165bbb0587fa9757](https://github.com/intel/cri-resource-manager/pull/196/commits/13a346adca81e84e10c70733165bbb0587fa9757) ?

ipuustin:
> Is it ok with you if I squash that into [13a346adca81e84e10c70733165bbb0587fa9757](https://github.com/intel/cri-resource-manager/pull/196/commits/13a346adca81e84e10c70733165bbb0587fa9757) ?

Sure, squash it!

klihub:
If everybody is happy enough with the current patch set, could you @marquiz ACK it so we can merge this ?

PR #197, opened 2020-03-09
cpuallocator: re-factor code

marquiz:
Make the CPU allocator an interface. One motivation of this change is to
get rid of the sysfs discovery in the cpuallocator package.

### DIFF
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -40,10 +40,11 @@ const (
 	logSource = "cpuallocator"
 )
 
-// CPUAllocator encapsulates state for allocating CPUs.
-type CPUAllocator struct {
-	logger.Logger               // allocator logger instance
+// allocatorHelper encapsulates state for allocating CPUs.
+type allocatorHelper struct {
+	logger.Logger               // allocatorHelper logger instance
 	sys           sysfs.System  // sysfs CPU and topology information
+	topology      topologyCache // cached topology information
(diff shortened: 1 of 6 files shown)


pkg/cpuallocator/allocator.go
```diff
@@ -157,16 +155,16 @@ func (s *sysfsSingleton) pick(idSlice []sysfs.ID, f IDFilter) []sysfs.ID {
 	return ids[0:idx]
 }
 
-func (s *sysfsSingleton) discoverPriorityCpus() {
-	s.priorityCpus = cpuset.NewCPUSet()
-	if s.sys == nil {
+func (ca *cpuAllocator) discoverPriorityCpus() {
+	ca.priorityCpus = cpuset.NewCPUSet()
+	if ca.sys == nil {
```
ipuustin: Is this an error case or is `sys` expected to be empty in this situation? If this is an error, you could check this already in `NewCPUAllocator()`.

marquiz: Thanks for looking this! The original code is written so that sys could be `nil`, in theory. Now that I took a closer look that actually seems to cause a crash. I'll update the PR with revised error handling...

marquiz: #201 addresses the issue

marquiz:
Let's get #201 resolved before continuing with this one

marquiz:
Pushed a new revised version.

PR #198, opened 2020-03-10
fixes: push most of the handling of NullPolicy to the policy layer.

klihub:
This patch series pushes most of the logic for special treatment of the 'null' pseudo-policy to the policy layer itself.

### DIFF
--- a/pkg/cri/resource-manager/events.go
+++ b/pkg/cri/resource-manager/events.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/metrics"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/policy"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
 )
 
@@ -44,6 +45,10 @@ func (m *resmgr) setupEventProcessing() error {
 
 // startEventProcessing starts event and metrics processing.
 func (m *resmgr) startEventProcessing() error {
+	if policy.Bypassed() {
+		return nil
+	}
+
 	if err := m.metrics.Start(); err != nil {
 		return resmgrError("failed to start metrics (pre)processor: %v", err)
 	}
@@ -70,8 +75,11 @@ func (m *resmgr) startEventProcessing() error {
 
 // stopEventProcessing stops event and metrics processing.
 func (m *resmgr) stopEventProcessing() {
-	close(m.stop)
-	m.metrics.Stop()
+	if m.stop != nil {
+		close(m.stop)
+		m.metrics.Stop()
+		m.stop = nil
+	}
 }
 
 // SendEvent injects the given event to the resource manaager's event processing loop.
--- a/pkg/cri/resource-manager/metrics/metrics.go
+++ b/pkg/cri/resource-manager/metrics/metrics.go
@@ -125,8 +125,10 @@ func (m *Metrics) Start() error {
 
 // Stop stops metrics collection and processing.
 func (m *Metrics) Stop() {
-	close(m.stop)
-	m.stop = nil
+	if m.stop != nil {
+		close(m.stop)
+		m.stop = nil
+	}
 }
 
 // poll does a single round of raw metrics collection.
--- a/pkg/cri/resource-manager/policy/flags.go
+++ b/pkg/cri/resource-manager/policy/flags.go
@@ -30,7 +30,7 @@ const (
 	// NullPolicy is the reserved name for disabling policy altogether.
 	NullPolicy = "null"
 	// NullPolicyDescription is the description for the null policy.
-	NullPolicyDescription = "A policy to bypass even the agnostic policy layer."
+	NullPolicyDescription = "A policy to bypass local policy processing."
 )
 
 // Options captures our configurable parameters.
--- a/pkg/cri/resource-manager/policy/policy.go
+++ b/pkg/cri/resource-manager/policy/policy.go
@@ -132,8 +132,9 @@ type Policy interface {
 // Policy instance/state.
 type policy struct {
 	cache   cache.Cache   // system state cache
-	backend Backend       // our active backend
+	active  Backend       // our active backend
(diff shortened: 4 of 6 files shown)


mythi:
> "deeper" and created a null/nil policy something similar to none, would it make any sense?

This question came to my mind too but after thinking it a bit my conclusion was that `null` is about not processing the policies. If `null` would be handled deeper we would be processing the policies.

marquiz:
> If `null` would be handled deeper we would be processing the policies.

Well, the general policy layer is activated anyway. It would just remove the `if`s from this upport policy layer and replace them with stubs in the policy.

pkg/sysfs/system.go
```diff
@@ -467,9 +467,10 @@ func (sys *system) discoverCPUs() error {
 
 	offline, err := sys.SetCpusOnline(true, nil)
 	if err != nil {
-		return fmt.Errorf("failed to set CPUs online: %v", err)
+		sys.Warn("failed to set CPUs online: %v", err)
+	} else {
+		defer sys.SetCpusOnline(false, offline)
 	}
-	defer sys.SetCpusOnline(false, offline)
```
ipuustin: This is an excellent change because the old behavior didn't work well with running tests using synthetic `/sys` data.

mythi:
> replace them with stubs in the policy.

Isn't this what `none` already does.  How would `none` and `null` be different?

marquiz:
`null` also completely disables cri request interception

mythi:
OK, I'm not quite clear on this. Should `null` policy then set `passthrough` interception?

marquiz:
> Should `null` policy then set `passthrough` interception

Yes. In practice, the policy methods (even stubs) should never get called

klihub:
> > "deeper" and created a null/nil policy something similar to none, would it make any sense?
> 
> This question came to my mind too but after thinking it a bit my conclusion was that `null` is about not processing the policies. If `null` would be handled deeper we would be processing the policies.

`none` is a full blown policy from the infra point of view. It just does nothing, it produces no decisions, IOW makes no changes to any containers.

`null` on the other hand disables message interception altogether. So it is pretty much as close to short-circuit relaying as possible without too much special-case handling in the upper layer (resource-manager) code.

klihub:
> Looks good. I just started to wonder what would it look like if we pushed the handling even one level "deeper" and created a null/nil policy something similar to none, would it make any sense?
> 
> The last commit seems a bit unrelated to the PR(?)

Yes, I know it looks like it. I can split it out, but then this PR should depend on the split out one. Without the last commit but with the rest included some of the tests start failing as they run on different code paths and try unconditionally toggling CPUs online.

klihub:
> Looks good. I just started to wonder what would it look like if we pushed the handling even one level "deeper" and created a null/nil policy something similar to none, would it make any sense?

What would be then the difference between `none` and `null` ? `none` is already exactly that.

pkg/sysfs/system.go
klihub: Would you like me to split that out to a PR of its own. We could merge it alone then before considering the rest of this.

marquiz:
> What would be then the difference between `none` and `null` ? `none` is already exactly that.

Disabling message interception in the server. Looking at the code, `null` looks like effectively a `none` policy backend but implemented with `if`s in the upper policy layer. But let's not get stuck with this 😄 I think the code is fine.

klihub:
> > What would be then the difference between `none` and `null` ? `none` is already exactly that.
> 
> Disabling message interception in the server. Looking at the code, `null` looks like effectively a `none` policy backend but implemented with `if`s in the upper policy layer. But let's not get stuck with this  I think the code is fine.

Ok. Just one final related comment/point... ;-) Exactly because the `null` policy disables interception, it does not really make sense to have it as a real policy implementation: the stubs would never get called since we are short-circuit-relaying messages (and actually even an instance of the null policy would never get created).

marquiz:
> the stubs would never get called since we are short-circuit-relaying messages (and actually even an instance of the null policy would never get created).

Yeah, that's why we could in principle drop the `p.active` checks from the upper policy layer, right? But, if we want to have these "anti-crash measures" a way to reduce the complexity and loc of the upper policy layer would be to e.g. re-use the stubs from none policy. Anyway, that was just a thought

klihub:
> > the stubs would never get called since we are short-circuit-relaying messages (and actually even an instance of the null policy would never get created).
> 
> Yeah, that's why we could in principle drop the `p.active` checks from the upper policy layer, right? But, if we want to have these "anti-crash measures" a way to reduce the complexity and loc of the upper policy layer would be to e.g. re-use the stubs from none policy. Anyway, that was just a thought

That's a fair point. And I think now I understand what bothers you. If I paraphrase it, either
  1. we should trust the short-circuit logic and remove the nil checks from the policy layer, or
  2. if we don't trust the logic, still remove the nil checks and add a dummy `none`-like `null` policy for less cluttered code/better readability.

Did I get that correctly ? If yes, then I'd go for trusting the short-circuit logic, removing the nil-checks, dealing with the consequences and fixing any remaining related problems.

marquiz:
> Did I get that correctly ? If yes, then I'd go for trusting the short-circuit logic, removing the nil-checks, dealing with the consequences and fixing any remaining related problems.

Thanks for reading my thoughts and writing them down 😄 Yes, you put it correctly 👍

klihub:
> > Did I get that correctly ? If yes, then I'd go for trusting the short-circuit logic, removing the nil-checks, dealing with the consequences and fixing any remaining related problems.
> 
> Thanks for reading my thoughts and writing them down  Yes, you put it correctly

Hmm... unfortunately, it won't be that clear-cut. Short-circuiting cuts out those policy calls which are directly related to CRI request processing. But then there are others, for instance Rebalance(). We should not start the event/metrics processing loop either if null is active. That'll then cut out Rebalance(). That would leave us with Bypassed()-checks for preventing the initial sync and starting the event loop in resource-manager, and one more in policy.Start() in addition to the constructor.

I'm not happy with this but I think that's the best we can do with this small changes. Maybe it is a reasonable compromise for now. We will anyway need to revisit this when we add support for changing the policy dynamically via configuration updates. We can try to come up with something better then.

PR #203, opened 2020-03-11
resource-manager: free any lingering instance upon pod/container recreation. 

### LINKED ISSUE #146: Properly detect and act on container crashes.
We used to piggyback on ListContainers periodically sent by kubelet to detect state changes of containers. We stopped doing that at one point and as a side-effect now we don't always properly detect when a container crashes. This eventually leads to resources being allocated to crashed containers and new container creation requests failing with insufficient resources.

I *think* one way to trigger/test this is to run a container that constantly leaks memory, with a memory limit set, until the OOM-killer decides to kill it. At this point, if the container's CPU request is more than the amount of remaining free CPU when the container was still running, the kubelet's attempt to (re)create the crashed container will fail with an error about insufficient available CPU.

klihub:
Free any lingering old instance of a container before recreating it.

kubelet occasionally fails to properly free crashed containers
before attempting to recreate and restart them. Check for this
and internally release any such instance if found. Likewise, do
a similar cleanup when a pod is being recreated.

Fixes #146 .

### DIFF
--- a/pkg/cri/resource-manager/no-test-api.go
+++ b/pkg/cri/resource-manager/no-test-api.go
@@ -1,0 +1,21 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !test
+
+package resmgr
+
+// ResourceManagerTestAPI is dummy if we're compiling without test build flag.
+type ResourceManagerTestAPI interface {
+}
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -135,6 +135,34 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
 	pod := m.cache.InsertPod(podID, request)
 	m.updateIntrospection()
 
+	// search for any lingering old version and clean up if found
+	released := false
+	for _, p := range m.cache.GetPods() {
+		if p.GetUID() != pod.GetUID() || p == pod {
+			continue
+		}
+		m.Warn("re-creation of pod %s, releasing old one", p.GetName())
+		for _, c := range pod.GetInitContainers() {
+			m.Info("%s: removing stale init-container %s...", method, c.PrettyName())
+			m.policy.ReleaseResources(c)
+			c.UpdateState(cache.ContainerStateStale)
+			released = true
+		}
+		for _, c := range pod.GetContainers() {
+			m.Info("%s: removing stale container %s...", method, c.PrettyName())
+			m.policy.ReleaseResources(c)
+			c.UpdateState(cache.ContainerStateStale)
+			released = true
+		}
+		m.cache.DeletePod(p.GetID())
+	}
+	if released {
+		if err := m.runPostReleaseHooks(ctx, method); err != nil {
+			m.Error("%s: failed to run post-release hooks for lingering pod %s: %v",
+				method, pod.GetName(), err)
+		}
+	}
+
 	m.Info("created pod %s (%s)", pod.GetName(), podID)
 
 	return reply, nil
@@ -198,6 +226,18 @@ func (m *resmgr) CreateContainer(ctx context.Context, method string, request int
 
 	m.Lock()
 	defer m.Unlock()
+
+	// kubelet doesn't always clean up crashed containers so we try doing it here
+	if msg, ok := request.(*criapi.CreateContainerRequest); ok {
+		if pod, ok := m.cache.LookupPod(msg.PodSandboxId); ok {
+			if msg.Config != nil && msg.Config.Metadata != nil {
+				if c, ok := pod.GetContainer(msg.Config.Metadata.Name); ok {
+					m.Warn("re-creation of container %s, releasing old one", c.PrettyName())
+					m.policy.ReleaseResources(c)
+				}
+			}
+		}
+	}
 
 	container, err := m.cache.InsertContainer(request)
 	if err != nil {
--- a/pkg/cri/resource-manager/test-api.go
+++ b/pkg/cri/resource-manager/test-api.go
@@ -1,0 +1,31 @@
+// Copyright 2029 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build test
+
+package resmgr
+
+import (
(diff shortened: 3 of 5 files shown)


pkg/cri/resource-manager/requests.go
```diff
@@ -141,6 +141,31 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
 	podID := reply.(*criapi.RunPodSandboxResponse).PodSandboxId
 	pod := m.cache.InsertPod(podID, request)
 
+	// search for any lingering old version and clean up if found
+	released := false
+	for _, p := range m.cache.GetPods() {
+		if p.GetUID() != pod.GetUID() || p == pod {
```
ipuustin: Can `p == pod` happen if the UIDs are different? Maybe there should be a utility function for finding out if two pods are the same.

pkg/cri/resource-manager/requests.go
```diff
@@ -204,6 +229,18 @@ func (m *resmgr) CreateContainer(ctx context.Context, method string, request int
 	m.Lock()
 	defer m.Unlock()
 
+	// kubelet doesn't always clean up crashed containers so we try doing it here
+	if msg, ok := request.(*criapi.CreateContainerRequest); ok {
+		if pod, ok := m.cache.LookupPod(msg.PodSandboxId); ok {
+			if msg.Config != nil && msg.Config.Metadata != nil {
+				if c, ok := pod.GetContainer(msg.Config.Metadata.Name); ok {
+					m.Warn("re-creation of container %s, releasing old one", c.PrettyName())
+					m.policy.ReleaseResources(c)
```
ipuustin: I don't maybe get the logic here... are the stale resources released both in `CreateContainer()` and `RunPod()`? Wouldn't one place be enough?

pkg/cri/resource-manager/requests.go
klihub: > If Kubelet should clean after the containers but it doesn't, should we rather try to address the issue upstream or at least file bug?

I was thinking about that, but I'm not sure if this is really a bug in Kubernetes or just something related to timing of its container manager reconciliation or other internal goroutines. Of course one could say, that the behavior depending for this so much on timing is a bug in itself. 

> Or does Kubelet assume that the container runtime monitors and cleans after the containers anyway?

Again, I'm not sure. I don't think it assumes too much autonomous intelligence on on the runtime side. This would not be a problem for us if we didn't do resource accounting. If you blindly obey what the kubelet says it does not matter if you have a number of entries for crashed/exited containers and pods. Also this could be fixed/worked around in other ways as well, for instance by trying to figure out every single time from scratch how many of the containers are still running and do dynamic resource accounting/calculations based on that. I just really don't like that idea.

> Also, if this is difficult to test in the real world, a synthetic test would be nice to validate that the release logic is correct.

Yes, we need synthetic tests for this. I'll check if I can add one to Dima's basic/PoC end-to-end tests and if not what changes it requires to start adding tests.

klihub: > Can `p == pod` happen if the UIDs are different?

No.

> Maybe there should be a utility function for finding out if two pods are the same.

IMO that wouldn't make sense (would be used only once, here) if this gets merged, since we exactly protect against having two instances of the same pod here.

This should probably be easier to understand if we turn the current 'check to see if we should skip' into a 'check if we should clean up', by nesting the cleanup like this:
```
  if p != pod && p.GetUID() == pod.GetUID() { // a lingering instance of the same pod
    // do the cleanup...
  }
```

pkg/cri/resource-manager/requests.go
klihub: > I don't maybe get the logic here... are the stale resources released both in `CreateContainer()` and `RunPod()`? Wouldn't one place be enough?

Those are cleanups for different objects: 1) a lingering pod (with any potential lingering containers), 2) a single lingering container.

pkg/cri/resource-manager/requests.go
mythi: Quoting #146 
> We used to piggyback on ListContainers periodically sent by kubelet to detect state changes of

What were the problems with ListContainers?

klihub: The problem was not the idea, it was the implementation. The basic idea of piggybacking on ListContainers/ListPodSandbox is fine. However, cleaning up detected containers and pods that look like lingering ones should happen only after a timeout. Otherwise we end up second-guessing what the kubelet will do, since we see and react to the reply (being relayed back) before the kubelet itself has a chance. In the best case, if nothing else goes wrong this causes a number of unnecessary temporary reallocations for the short duration between us processing and relaying the reply, and the kubelet receiving+processing the reply and then restarting containers. The implementation had other bugs too, so it was removed when we reworked request/event processing. We'll be adding piggybacked cleanup back eventually. But since we'll be delaying the piggybacked cleanups, we anyway need to check for restart/duplicates during pod and container creation anyway, which this patch series tries to implement.

klihub:
> Also, if this is difficult to test in the real world, a synthetic test would be nice to validate that the release logic is correct.

I've added synthetic tests for this to the end-to-end test set.

test/functional/e2e_test.go
```diff
@@ -40,7 +45,25 @@ func init() {
...
+	t           *testing.T
+	handlers    map[string]interface{}
+	client      api.RuntimeServiceClient
+	forceConfig string
+	mgr         resmgr.ResourceManager
+	cache       cache.Cache
+}
+
+// a big freakin' test lock to protect until we fix racy components...
+var BFTL sync.Mutex
```
marquiz, line 58: Is this big hammer still needed? If it is, what are the racy components left? We should create an issue tracking those.

pkg/cri/resource-manager/no-test-api.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
marquiz, line 1: nit: year 2020(?)

pkg/cri/resource-manager/test-api.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
marquiz, line 1: 2020(?)

pkg/cri/resource-manager/requests.go
```diff
@@ -135,6 +135,31 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
...
+			m.policy.ReleaseResources(c)
+			c.UpdateState(cache.ContainerStateStale)
+		}
+		for _, c := range pod.GetContainers() {
+			m.Info("%s: removing stale container %s...", method, c.PrettyName())
+			m.policy.ReleaseResources(c)
+			c.UpdateState(cache.ContainerStateStale)
+		}
+	}
+	if released {
```
marquiz, line 157: Where do we set this to true?

pkg/cri/resource-manager/requests.go
```diff
@@ -152,6 +152,7 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
 			m.policy.ReleaseResources(c)
 			c.UpdateState(cache.ContainerStateStale)
 		}
+		m.cache.DeletePod(p.GetID())
```
marquiz, line 155: nit: this line should be part of the first patch(?)

test/functional/e2e_test.go
ipuustin: And if we can expose those bugs by running this test without the lock, even better!

pkg/cri/resource-manager/requests.go
klihub: Definitely.

test/functional/e2e_test.go
klihub: > Is this big hammer still needed? 

Hopefully not. Let's remove the BFTL and find out...

pkg/cri/resource-manager/no-test-api.go
klihub: Updated.

pkg/cri/resource-manager/test-api.go
klihub: Updated.

pkg/cri/resource-manager/requests.go
klihub: Hmm... apparently nowhere. Fixed.

PR #206, opened 2020-03-12
Add initial version of the blockio control

askervin:
- Assign cgroup blkio parameters to containers according to pod's
  block I/O or QOS class.
- Support cgroup v1.
- Support static configuration: existing block devices are scanned
  only on configNotify to keep pod launch overhead minimal.

### DIFF
--- a/pkg/blockio/blockio.go
+++ b/pkg/blockio/blockio.go
@@ -1,0 +1,407 @@
+/*
+Copyright 2020 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package blockio
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/intel/cri-resource-manager/pkg/cgroups"
+	pkgcfg "github.com/intel/cri-resource-manager/pkg/config"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	logger "github.com/intel/cri-resource-manager/pkg/log"
+	"github.com/intel/cri-resource-manager/pkg/utils"
+)
+
+const (
+	// sysfsBlockDeviceIOSchedulerPaths expands (with glob) to block device scheduler files.
+	// If modified, check how to parse device node from expanded paths.
+	sysfsBlockDeviceIOSchedulerPaths = "/sys/block/*/queue/scheduler"
+)
+
+// BlockDeviceInfo holds information on a block device to be configured.
+// As users can specify block devices using wildcards ("/dev/disk/by-id/*SSD*")
+// BlockDeviceInfo.Origin is maintained for traceability: why this
+// block device is included in configuration.
+// BlockDeviceInfo.DevNode contains resolved device node, like "/dev/sda".
+type BlockDeviceInfo struct {
+	Major   int64
+	Minor   int64
+	DevNode string
+	Origin  string
+}
+
+// Our logger instance.
+var log logger.Logger = logger.NewLogger("blockio")
+
+// staticOciBlockIO connects user-defined block I/O classes to
+// corresponding OCI BlockIO parameters. "Static" means that
+// new/current block devices matching device wildcards in these
+// classes are not expanded every time new containers are assigned to
+// these classes. Devices are scanned on only at the beginning and on
+// blockio configuration changes.
+var staticOciBlockIO = map[string]cgroups.OciBlockIOParameters{}
+
+// currentIOSchedulers contains io-schedulers (found in
+// sysfsBlockDeviceIOSchedulerPaths) of device nodes:
+// {"/dev/sda": "bfq"}
+var currentIOSchedulers map[string]string
+
+// Initialize handles initial configuration and starts listening to config changes
+func Initialize() error {
+	pkgcfg.GetModule("blockio").AddNotify(configNotify)
+	applyConfig(true)
+	return nil
+}
+
+// configNotify is blockio class definition configuration notification callback.
+func configNotify(event pkgcfg.Event, source pkgcfg.Source) error {
+	log.Info("blockio I/O class configuration updated")
+	ignoreErrors := (event == pkgcfg.RevertEvent)
+	err := applyConfig(ignoreErrors)
+	return err
+}
+
+// applyConfig takes the configuration from the opt variable into use for new pods.
+func applyConfig(ignoreErrors bool) error {
+	currentIOSchedulers, ioSchedulerDetectionError := getCurrentIOSchedulers()
+	if ioSchedulerDetectionError != nil {
+		log.Warn("configuration validation partly disabled due to IO scheduler detection error %#v", ioSchedulerDetectionError.Error())
+	}
+
+	staticOciBlockIO = map[string]cgroups.OciBlockIOParameters{}
+	// Create static OCI BlockIO structures for each blockio class
+	for class := range opt.Classes {
+		ociBlockIO, err := devicesParametersToOci(opt.Classes[class], currentIOSchedulers)
+		if err != nil {
+			if ignoreErrors {
+				log.Error("ignoring: %v", err)
+			} else {
+				return err
+			}
+		}
+		// Handle all configurations as static for now. That
+		// is, the list of block devices matching Devices
+		// wildcards will not be updated without new
+		// configNotify(). class.DynamicDevices not supported
+		// yet.
+		staticOciBlockIO[class] = ociBlockIO
+	}
+	return nil
+}
+
+// SetContainerClass assigns the pod in a container to a blockio class.
+func SetContainerClass(c cache.Container, class string) error {
+	pod, ok := c.GetPod()
+	if !ok {
+		return blockioError("failed to get Pod for %s", c.PrettyName())
+	}
+
+	ociBlockIO, classIsStatic := staticOciBlockIO[class]
+	if !classIsStatic {
+		return blockioError("no OCI BlockIO parameters for class %#v", class)
+	}
+
+	blkioCgroupPodDir := cgroups.GetBlkioDir() + "/" + pod.GetCgroupParentDir()
+	containerCgroupDir := utils.GetContainerCgroupDir(blkioCgroupPodDir, c.GetID())
+	if containerCgroupDir == "" {
+		return blockioError("failed to find cgroup directory for container %s under %#v, container id %#v", c.PrettyName(), blkioCgroupPodDir, c.GetID())
+	}
+
+	err := cgroups.SetBlkioParameters(containerCgroupDir, ociBlockIO)
+	if err != nil {
+		return blockioError("assigning container %v to class %#v failed: %w", c.PrettyName(), class, err)
+	}
+
+	return nil
+}
+
+// getCurrentIOSchedulers returns currently active io-scheduler used for each block device in the system.
+func getCurrentIOSchedulers() (map[string]string, error) {
+	var ios = map[string]string{}
+	schedulerFiles, err := filepath.Glob(sysfsBlockDeviceIOSchedulerPaths)
+	if err != nil {
+		return ios, blockioError("error in IO scheduler wildcards %#v: %w", sysfsBlockDeviceIOSchedulerPaths, err)
+	}
+	for _, schedulerFile := range schedulerFiles {
+		devName := strings.SplitN(schedulerFile, "/", 5)[3]
+		schedulerDataB, err := ioutil.ReadFile(schedulerFile)
+		if err != nil {
+			// A block device may be disconnected. Continue without error.
+			log.Error("failed to read current IO scheduler %#v: %v\n", schedulerFile, err)
+			continue
+		}
+		schedulerData := string(schedulerDataB)
+		openB := strings.Index(schedulerData, "[")
+		closeB := strings.Index(schedulerData, "]")
+		if -1 < openB && openB < closeB {
+			currentScheduler := schedulerData[openB+1 : closeB]
+			ios["/dev/"+devName] = currentScheduler
+		} else {
+			return ios, blockioError("could not parse current scheduler in %#v\n", schedulerFile)
+		}
+	}
+	return ios, nil
+}
+
+// deviceParametersToOci converts single blockio class parameters into OCI BlockIO structure.
+func devicesParametersToOci(dps []DevicesParameters, currentIOSchedulers map[string]string) (cgroups.OciBlockIOParameters, error) {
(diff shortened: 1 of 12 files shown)


pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -133,25 +164,282 @@ func (ctl *blockio) BlockIOClass(c cache.Container) string {
...
-
-	bioclass, ok := opt.Classes[cclass]
+	_, ok := opt.Classes[cclass]
 	if !ok {
-		if bioclass, ok = opt.Classes["*"]; !ok {
-			bioclass = cclass
+		if _, ok := opt.Classes["*"]; !ok {
+			return ""
+		} else {
+			return "*"
```
klihub: Hmm, what is the logic/reasoning here ? Do we want to support a verbatim class name of "*" ? The original code was using that as a fallback/wildcard for class name mapping like this:
  - start from block I/O class set for the container, falling back to QoS class is none is set
  - if the class is (explicitly) mapped to something use that, otherwise
  - if the class is wildcard-mapped to something use that, otherwise
  - use the class set or derived from QoS class

pkg/cgroups/cgroupblkio.go
```diff
@@ -0,0 +1,188 @@
...
+// NewOciBlockIOParameters creates new OciBlockIOParameters instance.
+func NewOciBlockIOParameters() OciBlockIOParameters {
+	return OciBlockIOParameters{
+		Weight:                  -1,
+		LeafWeight:              -1,
+		WeightDevice:            nil,
+		ThrottleReadBpsDevice:   nil,
+		ThrottleWriteBpsDevice:  nil,
+		ThrottleReadIOPSDevice:  nil,
+		ThrottleWriteIOPSDevice: nil,
```
klihub: nitpick: golang is actively discouraging labelled initialization of composite members to zero value. IOW, you should not set anything else than those -1's above.

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -133,25 +164,282 @@ func (ctl *blockio) BlockIOClass(c cache.Container) string {
...
+type BlockDeviceInfo struct {
+	Major  int64
+	Minor  int64
+	Origin string // for traceability: why this device is configured
+}
+
+func configurableBlockDevices(devWildcards []string) (map[string]BlockDeviceInfo, error) {
+	// Return map {devNode: BlockDeviceInfo}
+	// Example: {"/dev/sda": {Major:8, Minor:0, Origin:"from symlink /dev/disk/by-id/ata-VendorXSSD from wildcard /dev/disk/by-id/*SSD*"}}
+	errors := []error{}
```
klihub: nitpick: why do we collect these errors if we never do anything with them beyond collecting ?

pkg/utils/errors.go
```diff
@@ -0,0 +1,45 @@
...
+
+package utils
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+func JoinErrors(errors []error) error {
```
klihub: This is more of a comment about our whole code base and not this PR... I think we should eventually switch to using the stock golang errors package and use wrapping(%w)/errors.Unwrap wherever necessary/reasonable. If/When we do that, maybe we should also take a look at  [multierror](https://github.com/hashicorp/go-multierror) for handling multiple errors.

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -133,25 +164,282 @@ func (ctl *blockio) BlockIOClass(c cache.Container) string {
...
+			log.Warn("skipping %#v: cannot weight/throttle partitions%s", devRealpath, origin)
+			continue
+		}
+		blockDevices[devRealpath] = BlockDeviceInfo{
+			Major:  int64(major),
+			Minor:  int64(minor),
+			Origin: strings.TrimSpace(fmt.Sprintf("%v %v", blockDevices[devRealpath].Origin, devOrigin)),
+		}
+	}
+	return blockDevices, nil
```
klihub: Was there supposed to be an error Join/test in the end before that `return` like above:
```
    err = utils.JoinErrors(errors)
￼    if err != nil {
        return nil, err
￼    }
    return blockDevices, nil
```

klihub:
@askervin Apart a few small things this looks overall good to me.

pkg/cgroups/cgroupblkio.go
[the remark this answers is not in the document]
askervin: Ok, changed this to initialize only -1's.

pkg/cri/resource-manager/control/blockio/blockio.go
askervin: My initial thought with the "*" class was to support setting weights/throttling for all pods that do not belong to any other BlockIO or QOS class that is mentioned in the configuration.

The benefit would be that the same configuration would not need to be repeated. For example, one could set lower weight for Burstable and BestEffort classes by using class "*" and then define only Guaranteed class to assign higher weights.

The big drawback would be that the class configuration logic would then differ from what is already in RDT.

What do you think, would it be good to support a "default" class in configuration(s)? Or will I take this back to the original logic (copy from RDT), and we can possibly reconsider class matching for all control blocks later, if needed?

pkg/cri/resource-manager/control/blockio/blockio.go
askervin: Great finding. :) Signs of evolution already in the first commit... I'll recheck what needs to be done.

pkg/cri/resource-manager/control/blockio/blockio.go
klihub: Yes. Defaults definitely make sense. That's what '*' was meant for, but it was imagined to be used in a slightly different way, like in the RDT controller (IIRC). There the mapping happens in the the controller level while the classes are defined in low-level RDT implementation. So for instance you would define 'default' and 'Guaranteed' classes in the actual block I/O implementation (pkg/cgroups blockio) and map BestEffort and Burstable to 'default' in the controller.

I am not sure if this separation of the mapping from the class definitions really makes sense. It might be actually more confusing this way than if we had it in a single place. So I'm fine with either way (or anything else if it makes sense), but generally speaking I think we should try to keep the configuration approach to the RDT and block I/O controllers consistent unless there is a pressing reason to diverge.

pkg/cgroups/cgroupblkio.go
```diff
@@ -0,0 +1,188 @@
...
+		rateDeviceString := fmt.Sprintf("%d:%d %d", rateDevice.Major, rateDevice.Minor, rateDevice.Rate)
+		errors = append(errors, writeToFilesInDir(cgroupsDir, blkioThrottleReadIOPSFiles, rateDeviceString)...)
+	}
+
+	for _, rateDevice := range blockIO.ThrottleWriteIOPSDevice {
+		rateDeviceString := fmt.Sprintf("%d:%d %d", rateDevice.Major, rateDevice.Minor, rateDevice.Rate)
+		errors = append(errors, writeToFilesInDir(cgroupsDir, blkioThrottleWriteIOPSFiles, rateDeviceString)...)
+	}
+
+	err := utils.JoinErrors(errors)
```
klihub: nitpick: why not just `return util.JoinErrors(errors)` ?

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -133,25 +164,282 @@ func (ctl *blockio) BlockIOClass(c cache.Container) string {
...
+	// Example: devMatches["/dev/disk/by-id/ata-VendorSSD"] == "from wildcard \"dev/disk/by-id/*SSD*\""
+	devMatches := map[string]string{} // {devNodeOrSymlink: origin}
+	for _, devWildcard := range devWildcards {
+		devWildcardMatches, err := filepath.Glob(devWildcard)
+		if err != nil {
+			errors = append(errors, fmt.Errorf("bad device wildcard %#v: %+v", devWildcard, err))
+			continue
+		}
+		if len(devWildcardMatches) == 0 {
+			log.Warn("deviceParametersToOci: device wildcard %#v does not match any device nodes", devWildcard)
```
klihub: nitpick: isn't the 'function name prefix:' in the warning message incorrect (shouldn't it be `configurableBlockDevices`) ?

pkg/cri/resource-manager/control/blockio/blockio.go
askervin: Agreed with next steps: move logic from control/blockio to pkg/blockio and create similar class mapping to control/blockio as there is in contorl/rdt

pkg/cri/resource-manager/control/blockio/blockio.go
askervin: Thanks! Rewritten so that all parameter sets are handled even in case of error.

pkg/cgroups/cgroupblkio.go
askervin: Converted to multierror and implemented as you suggest.

pkg/cri/resource-manager/control/blockio/blockio.go
askervin: Yep that was wrong. Removed the prefix completely for better error messages.

askervin:
This is still WIP, blockio will be restructured:
- logic and tests from .../control/blockio will be moved to pkg/blockio
- control/blockio will be left as thin as block control/rdt
- rdt class handling will be copied to control/blockio

pkg/cgroups/cgroupblkio.go
```diff
@@ -0,0 +1,168 @@
...
+	content := fmt.Sprintf("%d:%d %d", major, minor, value)
+	return writeToFileInDir(baseDir, filenames, content)
+}
+
+// writeToFileInDir writes content to the first existing file in the list under baseDir.
+func writeToFileInDir(baseDir string, filenames []string, content string) error {
+	var errors *multierror.Error
+	// Returns list of errors from writes, list of single error due to all filenames missing or nil on success.
+	for _, filename := range filenames {
+		filepath := baseDir + "/" + filename
```
kad: https://golang.org/pkg/path/filepath/#Join

pkg/cgroups/cgroupblkio.go
```diff
@@ -0,0 +1,168 @@
...
+		err := writeToFile(filepath, content)
+		if err == nil {
+			return nil
+		}
+		errors = multierror.Append(errors, err)
+	}
+	err := errors.ErrorOrNil()
+	if err != nil {
+		return fmt.Errorf("could not write content %#v to any of files \"%s\": %w", content, strings.Join(filenames, "\", \""), err)
```
kad: instead of `\"%s\"` you can use %q. Also, instead of `strings.Join(filenames)` you can just print `%v` of filenames.

pkg/cgroups/cgroupblkio.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	for _, rateDevice := range blockIO.ThrottleWriteIOPSDevice {
+		errors = multierror.Append(errors, writeDevValueToFileInDir(cgroupsDir, blkioThrottleWriteIOPSFiles, rateDevice.Major, rateDevice.Minor, rateDevice.Rate))
+	}
+	return errors.ErrorOrNil()
+}
+
+// writeDevValueToFileInDir writes MAJOR:MINOR VALUE to the first existing file under baseDir
+func writeDevValueToFileInDir(baseDir string, filenames []string, major, minor, value int64) error {
+	content := fmt.Sprintf("%d:%d %d", major, minor, value)
```
kad: why not to merge that with next function? in practice it is just one line with `fmt.Sprintf()` difference.

pkg/cgroups/cgroupblkio.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// writeToFile writes content to an existing file.
+func writeToFile(filename string, content string) error {
+	f, err := os.OpenFile(filename, os.O_WRONLY, 0666)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	_, fperr := fmt.Fprintf(f, content)
```
kad: `_, err = fmt.Fprintf`
no need to allocate one more variable if you already have one `err`.

pkg/cgroups/cgroupblkio.go
askervin: Thanks, done in next push.

pkg/cgroups/cgroupblkio.go
askervin: For readability of CRI-RM output, I'd prefer current error message formatting:
```
could not write content "1:0 100" to any of files "blkio.bfq.weight_device", "blkio.weight_device": ...
```

over `%q` + `filenames`:
```
could not write content "1:0 100" to any of files ["blkio.bfq.weight_device" "blkio.weight_device"]: ...
```

or `%v` + `filenames`:
```
could not write content "1:0 100" to any of files [blkio.bfq.weight_device blkio.weight_device]: ...
```

Please let me know if you'd still like to see either of the latter messages rather than the first, and I'll simplify the code.

pkg/cgroups/cgroupblkio.go
askervin: When you reviewed the code, this and the unit tests for `cgroupsblkio.go` were not in place yet:

```go
type platformInterface interface {
    writeToFile(filename string, content string) error
}
```

The goal here is to keep the function that has to interact with the platform as simple as possible. This enables unit testing as much as possible, including this string formatting, just by mocking only the simple platform functions.

Currently `cgroupsblkio_test.go` verifies that correct string has been written to correct file.

pkg/cgroups/cgroupblkio.go
askervin: Thanks, changed that to what you're suggesting. I often like single-assignment-style, but here with errors it doesn't really help anybody. :)

pkg/cgroups/cgroupblkio.go
kad: doing manual quoting `\"%s\"` instead of `%q` from system library doesn't make any sense.
Having brackets around list of files it might be a bit less readable for end-user, but a lot more readable for developer of the code, with less chances to miss escaping of quote characters.

`
return fmt.Errorf("could not write content %#v to any of files %q: %w", content, filenames, err)
`
is more readable than
`
return fmt.Errorf("could not write content %#v to any of files \"%s\": %w", content, strings.Join(filenames, "\", \""), err)
`

askervin: Ok, I switched to `%q` and removed `Join()`. PR updated.

pkg/blockio/blockio.go
```diff
@@ -0,0 +1,407 @@
...
+func Initialize() error {
+	pkgcfg.GetModule("blockio").AddNotify(configNotify)
+	applyConfig(true)
+	return nil
+}
+
+// configNotify is blockio class definition configuration notification callback.
+func configNotify(event pkgcfg.Event, source pkgcfg.Source) error {
+	log.Info("blockio I/O class configuration updated")
+	ignoreErrors := (event != "update")
```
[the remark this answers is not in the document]
askervin: Thanks, definitely. Fixed.

pkg/blockio/blockio.go
```diff
@@ -0,0 +1,407 @@
...
+		log.Warn("configuration validation partly disabled due to IO scheduler detection error %#v", ioSchedulerDetectionError.Error())
+	}
+
+	staticOciBlockIO = map[string]cgroups.OciBlockIOParameters{}
+	// Create static OCI BlockIO structures for each blockio class
+	for class := range opt.Classes {
+		ociBlockIO, err := devicesParametersToOci(opt.Classes[class], currentIOSchedulers)
+		if err != nil {
+			if ignoreErrors {
+				log.Warn("ignoring: %v", err)
```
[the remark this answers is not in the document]
askervin: Switched to `log.Error()` here and changed another `Warn` to `Error`, too.

PR #208, opened 2020-03-17
rdt: un-export Control interface

marquiz:
The resctrl API is essentially a singleton on the system level, thus,
there should be no reason to have multiple RDT separate control
interfaces in CRI-RM, either. This drops the 'Control' interface and
replaces that with "singleton" functions.

Also, add minimal unit test for the public interface, testing that an
uninitialized rdt interface works as expected.

### DIFF
--- a/pkg/rdt/rdt.go
+++ b/pkg/rdt/rdt.go
@@ -37,57 +37,64 @@ const (
 	rootClassName = "SYSTEM_DEFAULT"
 )
 
-// Control is the interface managing Intel RDT resources
-type Control interface {
-	// GetClasses returns the names of RDT classes (or resctrl control groups)
-	// available
-	GetClasses() []string
-
-	// SetProcessClass assigns a set of processes to a RDT class
-	SetProcessClass(string, ...string) error
-}
-
-var rdtInfo Info
-
-var log logger.Logger = logger.NewLogger("rdt")
-
 type control struct {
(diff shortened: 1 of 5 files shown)


pkg/rdt/rdt.go
```diff
@@ -95,7 +102,7 @@ func (r *control) GetClasses() []string {
 	return ret
 }
 
-func (r *control) SetProcessClass(class string, pids ...string) error {
+func (r control) setProcessClass(class string, pids ...string) error {
```
klihub: Hmm... are you sure it is a good idea to change the signature of *all* these functions to have a *copied receiver* instead of a *pointer* ? including the ones which are essentially *setters/mutators* ?

Now those work because you only touch `map` members which are passed *by reference*. However, this might bite us in the arse later, if/when someone extends `control` with something else than a `map` or a `slice` and then carelessly copies the idea of using non-pointer receivers for setters.

I'd rather have them consistently as *pointer receivers* already now, even if a non-pointer would work as a side-effect of the underlying data type.

klihub:
Apart from my signature-related concerns, I think this looks otherwise good.

pkg/rdt/rdt.go
marquiz: OH my, you're very much right. Thanks for staying sharp!! A total brain fart from me. Now fixed.

PR #210, opened 2020-03-19
resource-manager: reload forced configuration on SIGHUP.

klihub:
Add support for reloading forced configuration upon a signal for simplifying development and testing of configuration handling in sub-components. If forced configuration is in use, by default it can be reloaded using SIGHUP. Setting the reload signal name to empty disables this altogether.

### DIFF
--- a/pkg/cri/resource-manager/flags.go
+++ b/pkg/cri/resource-manager/flags.go
@@ -23,17 +23,18 @@ import (
 
 // Options captures our command line parameters.
 type options struct {
-	ImageSocket    string
-	RuntimeSocket  string
-	RelaySocket    string
-	RelayDir       string
-	AgentSocket    string
-	ConfigSocket   string
-	ResctrlPath    string
-	FallbackConfig string
-	ForceConfig    string
-	MetricsTimer   time.Duration
-	RebalanceTimer time.Duration
+	ImageSocket       string
+	RuntimeSocket     string
+	RelaySocket       string
+	RelayDir          string
+	AgentSocket       string
+	ConfigSocket      string
+	ResctrlPath       string
+	FallbackConfig    string
+	ForceConfig       string
+	ForceConfigSignal string
+	MetricsTimer      time.Duration
+	RebalanceTimer    time.Duration
 }
 
 // Relay command line options.
@@ -58,6 +59,8 @@ func init() {
 		"Fallback configuration to use unless/until one is available from the cache or agent.")
 	flag.StringVar(&opt.ForceConfig, "force-config", "",
 		"Configuration used to override the one stored in the cache. Does not override the agent.")
+	flag.StringVar(&opt.ForceConfigSignal, "force-config-signal", "SIGHUP",
+		"Signal used to reload forced configuration.")
 
 	flag.DurationVar(&opt.MetricsTimer, "metrics-interval", 30*time.Second,
 		"Interval for polling/gathering runtime metrics data. Use 'disable' for disabling.")
--- a/pkg/cri/resource-manager/resource-manager.go
+++ b/pkg/cri/resource-manager/resource-manager.go
@@ -15,6 +15,10 @@
 package resmgr
 
 import (
+	"golang.org/x/sys/unix"
+	"os"
+	"os/signal"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/cri/resource-manager/flags.go
```diff
@@ -58,6 +59,8 @@ func init() {
 		"Fallback configuration to use unless/until one is available from the cache or agent.")
 	flag.StringVar(&opt.ForceConfig, "force-config", "",
 		"Configuration used to override the one stored in the cache. Does not override the agent.")
+	flag.StringVar(&opt.ForceConfigSignal, "force-config-signal", "SIGHUP",
```
ipuustin: Why would anyone use anything other than `SIGHUP`? I think this config option might not be needed.

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -289,6 +298,45 @@ func (m *resmgr) loadConfig() error {
...
+	}
+
+	m.signals = make(chan os.Signal, 1)
+	signal.Notify(m.signals, sig)
+
+	go func(signals <-chan os.Signal) {
+		for {
+			select {
+			case _, ok := <-signals:
+				if !ok {
```
ipuustin: Do I understand right that this happens when the signal channel is closed?

pkg/cri/resource-manager/flags.go
marquiz: SIGUSR might be used, too

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -289,6 +298,45 @@ func (m *resmgr) loadConfig() error {
...
+					return
+				}
+			}
+
+			m.Info("reloading forced configuration %s...", opt.ForceConfig)
+
+			m.Lock()
+			if err := pkgcfg.SetConfigFromFile(opt.ForceConfig); err != nil {
+				m.Error("failed to reload forced configuration %s: %v",
+					opt.ForceConfig, err)
```
marquiz: Is this enough? Should we do something similar to what [SetConfig in resource-manager.go](https://github.com/intel/cri-resource-manager/blob/670360625f165da227dfcf1654f37dc17cf3b80e/pkg/cri/resource-manager/resource-manager.go#L163) does? E.g. call `m.control.StartStopControllers()`

pkg/cri/resource-manager/flags.go
marquiz: But, I'm not sure if this really needs to be configurable

pkg/cri/resource-manager/resource-manager.go
klihub: Yes, you're right. I'll fix that.

pkg/cri/resource-manager/resource-manager.go
klihub: > Do I understand right that this happens when the signal channel is closed?

Yes.

pkg/cri/resource-manager/flags.go
klihub: > Why would anyone use anything other than `SIGHUP`? I think this config option might not be needed.

I fully agree with your question. But I can't agree with your conclusion.

*Trust is good, control is better.* I would trust you or Markus any day to not do utterly stupid things. But we pull in an increasing number of packages written by all kinds of folks. We don't know them, we can't trust them. Hence we need control. The only thing we can really control in is our own code base.

Plus, with this being configurable you get for free the possibility to disable it.

pkg/cri/resource-manager/resource-manager.go
klihub: Fixed.

klihub:
@marquiz How about this ?

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -154,33 +159,66 @@ func (m *resmgr) Stop() {
...
+func (m *resmgr) setConfig(src interface{}) error {
+	var pending *config.RawConfig
+	var err error
+
+	switch cfg := src.(type) {
+	case *config.RawConfig:
+		pending = cfg
+		err = pkgcfg.SetConfig(cfg.Data)
+	case string:
+		err = pkgcfg.SetConfigFromFile(cfg)
```
marquiz: I'd somehow document that it's intentional that we're not caching config that is read from a file. It's not outright obvious and at least I tend to forget it

klihub: Added comments about never storing forced configuration in the cache. One in the switch pending is not set for forced configuration, and another where a pending config is stored.

klihub:
I think I'll just merge this. Based on our today's discussion on the call regarding the github workflow, I understood so that @ipuustin did not consider the extra configurability to be a merging show-stopper.

ipuustin:
Yes, I'm fine with the change as it is!

PR #212, opened 2020-03-20
ebpf: switch to cilium/ebpf from iovisor/gobpf.

mythi:
A POC to move away from `iovisor/gobpf` to `cilium/ebpf`.

The latter provides nicer API (less worries about binary
encoding and map utilities). However, it DOES NOT support
enabling tracepoints so this commits reverse engineers from
`iovisor/gobpf` how it's done (a good learning to see under the
hood).

This POC also builds the eBPF program inside the collector.

TODO:
- kernel version checking
- rlimits
- ...


pkg/avx/collector.go
```diff
@@ -1,22 +1,247 @@
...
+	bpf "github.com/cilium/ebpf"
 	"github.com/intel/cri-resource-manager/pkg/cgroups"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
-	bpf "github.com/iovisor/gobpf/elf"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sys/unix"
 )
 
+var program = [...]byte{
```
klihub: Wow. How was this generated and where is its source ?

mythi: it's from: `xxd -i libexec/avx512.o`

klihub: Ah, ok, it's just that. I thought it was something generated by some Cilium-specific toolchain.

mythi: not yet but there's one and I'm planning to see if we could move to that.

mythi:
v2: now completely without clang/ELF dependency.

pkg/avx/collector.go
```diff
@@ -70,118 +74,229 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
+	// Increase `ulimit -l` limit to avoid BPF_PROG_LOAD error (runc #2167).
+	memlockLimit := &unix.Rlimit{
+		Cur: unix.RLIM_INFINITY,
+		Max: unix.RLIM_INFINITY,
 	}
-
-	allSwitchCounters := bpfModule.Map("all_context_switch_count")
-	if allSwitchCounters == nil {
-		return nil, errors.New("map all_context_switch_count not found")
+	_ = unix.Setrlimit(unix.RLIMIT_MEMLOCK, memlockLimit)
```
ipuustin: Why no error checking?

pkg/avx/collector.go
```diff
@@ -70,118 +74,229 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
-		return nil, err
-	}
-
-	bpfModule := bpf.NewModule(elfFilepath)
-
-	sectionParams := make(map[string]bpf.SectionParams)
-	if err := bpfModule.Load(sectionParams); err != nil {
-		return nil, errors.Wrap(err, "unable to load eBPF ELF file")
+	// TODO: get rid of this
+	// Increase `ulimit -l` limit to avoid BPF_PROG_LOAD error (runc #2167).
```
ipuustin: Could this be some Systemd directive such as `LimitMEMLOCK=`? Maybe just check that the limit is large enough and error out if it isn't.

pkg/avx/collector.go
```diff
@@ -70,118 +74,229 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
-
-	allSwitchCounters := bpfModule.Map("all_context_switch_count")
-	if allSwitchCounters == nil {
-		return nil, errors.New("map all_context_switch_count not found")
+	_ = unix.Setrlimit(unix.RLIMIT_MEMLOCK, memlockLimit)
+
+	mapSpec := map[string]*bpf.MapSpec{
+		"avx_timestamp": {
+			Type:       bpf.Hash,
+			KeySize:    8,
```
ipuustin: Hmm, where do these sizes come from? Could they be read out from the original source?

pkg/avx/collector.go
```diff
@@ -70,118 +74,229 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
+		asm.Return(),
 	}
 
-	lastCPUCounters := bpfModule.Map("cpu")
-	if lastCPUCounters == nil {
-		return nil, errors.New("map cpu not found")
+	prog, err := bpf.NewProgram(&bpf.ProgramSpec{
+		Name:         "avx_collector",
+		Type:         bpf.TracePoint,
+		License:      "GPL",
```
ipuustin: Was the original code GPL? You should indicate the licensing in the file header / `COPYING`...

pkg/avx/collector.go
```diff
@@ -234,94 +362,73 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
...
 
 	// We need to wait so that the response channel doesn't get closed.
 	wg.Wait()
 
-	_, _, err = c.moveAllElements(c.allContextSwitchCounters, unsafe.Sizeof(uint64(0)), unsafe.Sizeof(uint32(0)))
-	if err != nil {
-		log.Error("unable to delete elements of all_context_switch_count: %+v", err)
+	key = 0
+	iter = m.Iterate()
+	for iter.Next(&key, &val) {
```
ipuustin: I get some C++ vibes here. :-)  Is it so that the deletion doesn't happen automatically when `m` goes out of scope?

pkg/avx/collector.go
mythi: this piece is still TODO. I want to understand better why it's needed in the first place (we didn't need it earlier)

pkg/avx/collector.go
mythi: the unit is bytes. keysize 8 is u64 etc from the original .c etc. this PR is basically the original source rewritten and if this approach is acceptable the original source could be dropped.

pkg/avx/collector.go
mythi: I wanted to have Apache here but that triggers:
`cannot call GPL-restricted function from non-GPL compatible program` from the kernel validator.

pkg/avx/collector.go
mythi: this deletes the map elements from the kernel memory, i.e., resets the counter values.

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
+	}
 
-	if err := checkElfKernelVersion(elfFilepath); err != nil {
-		return nil, err
+	str := make([]byte, len(uts.Release))
+	for _, val := range uts.Release[:] {
+		if val == 0 {
+			break
+		}
+		str = append(str, byte(val))
```
marquiz, line 145: I don't know if it makes it any easier to read, but, you could replace this with a one-liner something like:
```
str := string(bytes.SplitN(uts.Release[:], []byte{0}, 2)[0])
```

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
 	}
 
-	bpfModule := bpf.NewModule(elfFilepath)
+	ver := strings.SplitN(strings.Trim(string(str), "\x00"), ".", 3)
```
marquiz, line 148: With the above `str` should be guarenteed to not have any null characters so you could drop `Trim()` for readability?

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
-	sectionParams := make(map[string]bpf.SectionParams)
-	if err := bpfModule.Load(sectionParams); err != nil {
-		return nil, errors.Wrap(err, "unable to load eBPF ELF file")
+	major, err := strconv.ParseUint(ver[0], 10, 8)
+	if err != nil {
+		return 0
+	}
+	minor, err := strconv.ParseUint(ver[1], 10, 8)
+	if err != nil {
+		return 0
```
marquiz, line 156: We should still return major because we were able to parse that?

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to set rlimit")
 	}
 
-	avxSwitchCounters := bpfModule.Map("avx_context_switch_count")
-	if avxSwitchCounters == nil {
-		return nil, errors.New("map avx_context_switch_count not found")
+	spec, err := bpf.LoadCollectionSpec(filepath.Join(bpfInstallpath, bpfBinaryName))
+	if err != nil {
+		log.Info("User eBPF not found. Using default CollectionSpec from ELF program bytes")
```
marquiz, line 178: Should we print the error (with `%v`) just to be more informative about the error condition?

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
 
 var (
 	bpfBinaryName  = "avx512.o"
-	bpfInstallpath = "/usr/libexec/bpf"
+	bpfInstallpath = "/usr/libexec/bpf/"
```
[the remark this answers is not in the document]
mythi: typo :) dropped

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
 
-	sec := elfFile.Section("version")
-	if sec == nil {
-		return 0, 0, 0, errors.New("unable to find 'version' section")
-	}
-	data, err := sec.Data()
+	tid, err := strconv.Atoi(strings.TrimSpace(string(id)))
 	if err != nil {
-		return 0, 0, 0, errors.Wrap(err, "unable to get version data")
+		return -1, errors.New("unable to convert tracepoint ID")
```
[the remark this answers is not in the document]
mythi: it's `syscall.Errno` Perhaps I should add the actual errno as a string..

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
-	lastCPUCounters          *bpf.Map
-}
+func getKernelVersion() uint32 {
 
-// NewCollector creates new Prometheus collector for AVX metrics
-func NewCollector() (prometheus.Collector, error) {
+	var uts syscall.Utsname
 
-	elfFilepath := path.Join(bpfInstallpath, bpfBinaryName)
+	err := syscall.Uname(&uts)
```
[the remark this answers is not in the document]
mythi: `syscall` is still used below but I moved this to `unix`

pkg/avx/collector.go
mythi: great comment, I like your proposal a lot.

pkg/avx/collector.go
mythi: added.

pkg/avx/collector.go
mythi: makes sense

pkg/avx/collector.go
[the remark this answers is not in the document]
marquiz: Yes, you're correct. I didn't look at the api

pkg/avx/collector.go
```diff
@@ -71,117 +80,135 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
 	if err != nil {
-		return err
+		return -1, errors.Wrap(err, "unable open perf events")
 	}
 
-	currentCode, err := bpf.CurrentKernelVersion()
-	if err != nil {
-		return errors.Wrap(err, "unable to get current kernel version")
+	if _, _, errno := unix.Syscall(unix.SYS_IOCTL, uintptr(pfd), unix.PERF_EVENT_IOC_ENABLE, 0); errno != 0 {
+		return -1, errors.New("unable to set up perf events")
```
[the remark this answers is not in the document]
marquiz: My bad here, too. `syscall.Errno` is already returned so
```
return -1, errors.New("unable to set up perf events: %s", errno)
```
should be enough

pkg/avx/collector.go
```diff
@@ -77,111 +102,123 @@ var (
...
-// getElfKernelVersion returns major, minor and patch parts of kernel version
-// compiled into eBPF ELF file.
-func getElfKernelVersion(path string) (uint8, uint8, uint8, error) {
-	elfFile, err := elf.Open(path)
+func enablePerfTracepoint(prog *bpf.Program, tracepoint string) (int, error) {
+
+	id, err := ioutil.ReadFile(filepath.Join(kernelTracepointPath, tracepoint, "id"))
 	if err != nil {
-		return 0, 0, 0, errors.Wrapf(err, "unable to open ELF file %s", path)
+		return -1, errors.Wrap(err, "unable read tracepoint ID")
```
klihub, line 115: nittypo: "unable to read"

pkg/avx/collector.go
```diff
@@ -77,111 +102,123 @@ var (
...
 	}
 
-	lastUpdateNs := bpfModule.Map("last_update_ns")
-	if lastUpdateNs == nil {
-		return nil, errors.New("map last_update_ns not found")
+	ver := getKernelVersion()
+
+	for _, prog := range spec.Programs {
+		if ver < prog.KernelVersion {
+			return nil, errors.Wrap(err, "host kernel too old to run the program")
```
klihub, line 199: nit: I'd suggest also printing that minimal (prog.KernelVersion) acceptable version to let the user know.

Actually, and this is purely by looking at the code and without really understanding whether it can happen in practice that different spec.Programs have different kernel version requirements, I'd change that loop to not break out on the firsts version check failure but instead keep going and report all. You know, just to avoid hoards of angry sysadmins putting out a contract on our head once, after the 10th consecutive kernel upgrade, cri-resmgr still refuses to start up demanding one more newer kernel. I'd probably just turn that loop into a dedicated kernel version check function that return nil on success and an error on failure which would include the *highest* required kernel version among spec.Programs...

mythi: there's only one version specified in the elf so with the avx512 program(s), we can exit immediately. I made it in a loop to avoid the need to hard code map keys...

klihub: Hmm... I meant potentially multiple loaded ELF binaries in the AVX collector in the future, which `ebpf.CollectionSpec` seems to allow. But if we never going to have/need multiple programs then breaking out is fine. But could you add that required `prog.KernelVersion` to the returned error ?

mythi: > which `ebpf.CollectionSpec` seems to allow

AFAICS the elf reader gives one `CollectionSpec` per elf file. An elf can specify multiple programs. 

I've added the version info to the error for both minimum version and host kernel version .

klihub: This LGTM enough to merge and then address any remaining issues with further patches/PRs.

pkg/avx/collector.go
klihub: Maybe that should be changed to `KeySize: unsafe.Sizeof(uint64(0))` as a semi-self-documenting compromise. Then if @ipuustin  is okay with the remaining TODO items getting addressed in subsequent PRs, I would be ready to LGTM this for merging.

mythi: this does not exist in this PR anymore. let me close this conversation.

pkg/avx/elfdump.go
```diff
@@ -0,0 +1,84 @@
...
+const (
+	blocksPerRow = 12
+)
+
+type Program struct {
+	ProgramLines []string
+}
+
+func main() {
+	f, err := ioutil.ReadFile("../../libexec/avx512.o")
```
ipuustin, line 39: Maybe use the binary directory instead of working directory?

pkg/avx/elfdump.go
```diff
@@ -0,0 +1,84 @@
...
+`))
+
+	outfile, err := os.Create("programbytes_gendata.go")
+	if err != nil {
+		fmt.Println("elfdump:", err)
+		os.Exit(1)
+	}
+	defer outfile.Close()
+
+	template.Execute(outfile, p)
```
ipuustin, line 83: Might make sense to exit with an error value if generating output file fails.

pkg/avx/collector.go
```diff
@@ -77,111 +102,126 @@ var (
...
+
+	pfd, err := unix.PerfEventOpen(&attr, -1, 0, -1, unix.PERF_FLAG_FD_CLOEXEC)
 	if err != nil {
-		return 0, 0, 0, errors.Wrap(err, "unable to get version data")
+		return -1, errors.Wrap(err, "unable to open perf events")
 	}
 
-	// Least Significant Byte first
-	return data[2], data[1], data[0], nil
+	if _, _, errno := unix.Syscall(unix.SYS_IOCTL, uintptr(pfd), unix.PERF_EVENT_IOC_ENABLE, 0); errno != 0 {
```
ipuustin, line 136: Hmm, is it defined what `errno` is if the call succeeds (_"On error, -1 is returned, and errno is set appropriately."_)? Shouldn't you check the syscall return value?

pkg/avx/collector.go
```diff
@@ -77,111 +102,126 @@ var (
...
 
 // NewCollector creates new Prometheus collector for AVX metrics
 func NewCollector() (prometheus.Collector, error) {
 
-	elfFilepath := path.Join(bpfInstallpath, bpfBinaryName)
-
-	if err := checkElfKernelVersion(elfFilepath); err != nil {
-		return nil, err
+	// Set rlimit to be able to lock map values in memory
+	memlockLimit := &unix.Rlimit{
```
ipuustin, line 181: I think we can do this here for now, but in the long run I think the process limits, caps, etc. should come from systemd service files or similar. This might be easier said than done though. :-)

mythi: Right. This used to be "unlimited" but I changed it to minimum needed.

pkg/avx/elfdump.go
mythi: the eBPF ELF is generated to `libexec` directory. Are you asking to change the `Makefile` to output to `bin` directory?

pkg/avx/collector.go
mythi: `errno` is a `Syscall.Errno`: "the zero Errno is by convention a non-error" and I'm checking `!= 0`. Do you think the variable name (`errno`) is misleading?

ipuustin: I think I'm still stuck to the syscall return value of thinking. :-) If you know that errno 0 means that the call is successful, then that's the way to go.

pkg/avx/elfdump.go
ipuustin: No, I just meant that the file path is now relative to the working directory, meaning that you can run this binary only from the same directory where the binary is located. I was thinking that it should maybe be relative to the directory where the binary is. But this is more of a matter of taste I think, so no worries.

pkg/avx/elfdump.go
mythi: added error checking

klihub:
@kad, @marquiz, @mythi: Shall we merge this then ?

marquiz:
Same here 👍 Should we hold over the release, though?

klihub:
> Same here  Should we hold over the release, though?

Yes, let's do so.

mythi:
I'm OK with that. It's not currently enable so it should not have much effect but the change is a moving part anyway.

klihub:
> I'm OK with that. It's not currently enable so it should not have much effect but the change is a moving part anyway.

We can merge this immediately once (#285,) #288 and #289 are merged and we tag a new release.

PR #214, opened 2020-03-23
rdt: drop 'resctrl-path' config option and cmdline flag

marquiz:
The option is rather useless because in practice we depend on
(auto-)detecting the resctrl mount (options and mount point), anyway,
for determining the mode (percentage or MBPs) of MB allocation.
Simplifies the code and hopefully reduces confusion of the end users.

### DIFF
--- a/pkg/cri/resource-manager/control/rdt/doc.go
+++ b/pkg/cri/resource-manager/control/rdt/doc.go
@@ -30,7 +30,6 @@ Here is a sample configuration fragment for this controller which sets up
 mappings for the 3 Kubernetes QoS classes and defines also a default class.
 
   rdt:
-    ResctrlPath: /sys/fs/resctrl
     Classes:
       Guaranteed: BestRDT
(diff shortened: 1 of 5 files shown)


klihub:
Would this be also totally useless for checking RDT functionality in potential end-to-end tests ? Dima's PoC e2e tests (mis)uses other command line options to set up a fake CRI and create a resource-manager per test case. Of course if we add a way to override /proc/mounts then we could still achieve something to the same effect for RDT.

marquiz:
I'm writing functional tests that do this kind mocking – these will be part of the `pkg/rdt` package.

For real end-to-end testing I'd prefer to do it on a real system, without (or as little as possible) mocking. If we need to fool/misuse cri-rm, I'd do it by using fake `/proc/mounts´ (because we'd need to fake it anyway to "force" required mount options)

marquiz:
Thx! Let's add an option to force `proc/mounts` path later if really needed for testing

PR #220, opened 2020-03-25
Makefile: pass 'test' build tag to all tests.

klihub:
Pass the 'test' build tag to all tests. This can/will be used by some e2e test cases to expose extra interfaces/functionality for testing/verification.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,12 @@ GO_CILINT := golangci-lint
 
 # TEST_TAGS is the set of extra build tags passed for tests.
 # We disable AVX collector for tests by default.
-TEST_TAGS := -tags noavx
-GO_TEST   := $(GO_CMD) test $(TEST_TAGS)
+TEST_TAGS := noavx,test
+GO_TEST   := $(GO_CMD) test -tags $(TEST_TAGS)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,11 +8,12 @@ GO_CILINT := golangci-lint
 
 # TEST_TAGS is the set of extra build tags passed for tests.
 # We disable AVX collector for tests by default.
-TEST_TAGS := -tags noavx
-GO_TEST   := $(GO_CMD) test $(TEST_TAGS)
+TEST_TAGS := noavx,test
```
ipuustin: I think passing the `test` flag is fine as such but we need to be extra careful when actually doing something based on it. Having tests run with different options as the regular code can lead to a lot of head-scratching when debugging...

klihub: True and I agree. Carefulness is always warranted. We'll need extra scrutiny wjen reviewing patches which rely on this ro male sure no runtime functional changes are introduced.

I have one PR relying on this in the pipeline which simply adds an extra interface to expose some internals for extra verification during tests.

klihub:
@marquiz , @kad Any chance of getting #220 and #221 reviewed and merged ? Those should be really straightforward ones.

PR #221, opened 2020-03-25
Makefile: add race-test target for race detection in a single dir.

klihub:
Add `race-test` target to run non-verbose tests with race detection enabled in
a single directory. Use as
```
    make WHAT=dir/to/test race-test
```
If `WHAT` is ommitted `race-test` is identical to `test`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,6 +236,19 @@ else
             exit $$rc
 endif
 
+race-test racetest:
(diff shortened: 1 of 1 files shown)


klihub:
@marquiz , @kad Any chance of getting #220 and #221 reviewed and merged ? Those should be really straightforward ones.

PR #222, opened 2020-03-25
dump: fix dump/config race conditions, preserve message order, add unit tests

klihub:
There are a number of race conditions in pkg/dump. The most prominent one is related to concurrent message dumping and dynamic reconfiguration. This one is relatively easy to trigger on  a busy node. It also gets immediately triggered when one starts adding other than trivial end-to-end tests.

This PR reworks pkg/dump to fix these race conditions and a number of its other problems. In particular the PR
  - reworks message dumping to be race-free
  - preserves the order of messages
  - simplifies configuration by dropping command line support
  - adds unit tests

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Intel Corporation. All Rights Reserved.
+// Copyright 2019-2020 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import (
 	"github.com/ghodss/yaml"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	logger "github.com/intel/cri-resource-manager/pkg/log"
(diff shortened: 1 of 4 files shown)


pkg/dump/dump.go
```diff
@@ -45,175 +46,303 @@ const (
...
+		d.methods[idx] = method
+		d.details[method] = detail
+	}
+}
+
+// name does a name-only dump of the given message.
+func (d *dumper) name(dir direction, kind, method string, msg interface{}, latency time.Duration) {
+	switch dir {
+	case request:
+		return
```
marquiz, line 238: We don't dump requests?

pkg/dump/dump.go
```diff
@@ -45,175 +46,303 @@ const (
...
-func dumpToFile(dir string, latency time.Duration, format string, args ...interface{}) {
-	if !checkDumpFile() {
+// tofile dumps a single line to a file.
+func (d *dumper) tofile(dir direction, latency time.Duration, format string, args ...interface{}) {
+	if d.file == nil {
 		return
 	}
-
-	fmt.Fprintf(file, "["+stamp(dir, latency)+"] "+format+"\n", args...)
+	fmt.Fprintf(d.file, "["+stamp(dir, latency)+"] "+format+"\n", args...)
```
marquiz, line 315: nit: minuscule simplification
```
	if d.file != nil {
		fmt.Fprintf(d.file, "["+stamp(dir, latency)+"] "+format+"\n", args...)
	}
```

pkg/dump/dump.go
klihub: With name-only dumping we dump request/reply-pairs as a single line, and only once the reply has come in, so we know whether the request succeeded or failed.

pkg/dump/dump.go
```diff
@@ -45,175 +46,303 @@ const (
...
+	defer dump.Unlock()
+	dump.train(methods)
+}
+
+// RequestMessage dumps a CRI request.
+func RequestMessage(kind, name string, req interface{}, sync bool) {
+	if !dump.disabled {
+		var ch chan struct{}
+		if sync {
+			ch = make(chan struct{})
```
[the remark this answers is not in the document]
klihub: No, this channel is only used for signalling. I should fix the other instances which apparently I forgot in a rush to change to get correctly created with 0 length.

PR #223, opened 2020-03-26
pkg/blockio: fix unit test for single-scheduler-device case.

klihub:
If a block I/O devices supports only one scheduler (for instance a software raid or DM device), then the sysfs/.../queue/scheduler simply contains the name of the only supported scheduler without any brackets.


askervin:
Good finding & fix!
LGTM.

klihub:
@marquiz , @kad Could either of you check/ACK this ? Antti's approval is not enough to make it mergeable.

PR #224, opened 2020-03-26
RDT metrics

marquiz:
This PR adds support for RDT monitoring data both as internal metrics and exported as Prometheus metrics. It also adds simple unit tests for improving the test coverage of RDT.

Support for Prometheus metrics is implemented so that it should be easy to move to another package, if that was seen architecturally more justifiable.


pkg/rdt/rdt.go
```diff
@@ -80,54 +105,28 @@ func Initialize() error {
...
-				c.Debug("no task %s", pid)
-			} else {
-				return rdtError("failed to assign processes %v to class %q: %v", pids, class, c.cmdError(err))
-			}
-		}
+	for k, v := range c.classes {
+		ret[k] = v
 	}
-	return nil
+	return ret
```
klihub: Is it deliberate that, while we used to, we don't any more return classes in a stable (IOW sorted) order ?

pkg/rdt/rdt.go
```diff
@@ -284,27 +308,39 @@ func (c *control) getClassTasks(name string) ([]string, error) {
...
+		if _, err := f.WriteString(pid + "\n"); err != nil {
+			unwrapped := err
+			if pathError, ok := err.(*os.PathError); ok {
+				unwrapped = pathError.Unwrap()
+			}
+			if unwrapped == syscall.ESRCH {
+				log.Debug("no task %s", pid)
+			} else {
+				return rdtError("failed to assign processes %v to class %q: %v", pids, r.name, rdt.cmdError(err))
+			}
```
klihub: Isn't the standard `errors` package meant to simplify this kind of error checks, with something like
```
    if errors.Is(err, syscall.ESRCH) {
        log.Debug(...)
    } else {
        return rdtError(...)
    }
```

pkg/rdt/rdt.go
```diff
@@ -87,6 +90,30 @@ type MonGroup interface {
...
+// MonLeafData represents the raw numerical stats from one RDT monitor data leaf
+type MonLeafData map[string]uint64
+
+// MonResource is the type of RDT monitoring resource
+type MonResource string
+
+const (
+	// MonResourceL3 is the RDT L3 cache monitor resource
+	MonResourceL3 MonResource = "l3"
+)
```
klihub: Will there be more (for memory bandwidth?), or what is the reason for defining a type with/for a single resource ?

pkg/rdt/rdt.go
```diff
@@ -87,6 +90,30 @@ type MonGroup interface {
 	Parent() CtrlGroup
 }
 
+// Monitor is the interface for retrieving RDT monitoring data
+type Monitor interface {
+	GetMonData() MonData
+}
```
klihub: This is just a nitpicking question out of curiosity: the `Monitor` interface is never used anywhere as such, not even in test cases. IOW, practically every `ResctrlGroup` is `Monitor`able and currently everything/the only thing `Monitor`able is a `ResctrlGroup`. Is there any real/particular reason for splitting `Monitor` out as a separate interface, maybe some future changes in mind ?

pkg/rdt/rdt.go
```diff
@@ -87,6 +90,30 @@ type MonGroup interface {
...
+
+// MonL3Data contains L3 monitoring stats of one monitoring group
+type MonL3Data map[uint64]MonLeafData
+
+// MonLeafData represents the raw numerical stats from one RDT monitor data leaf
+type MonLeafData map[string]uint64
+
+// MonResource is the type of RDT monitoring resource
+type MonResource string
```
klihub: nitpick: this is a general note, much more related to my own code than this...

I've noticed lately that I tend to often do the same thing you do here. Define a set of `const`s with `string`s as the underlying type, while the consts are also used to index `map`s. With this setup, the only thing you gain/save is typing: you don't have to write a trivial `String()` function with a `receiver` of your defined type. The price you pay for that laziness is much less efficient `map lookup`s. Sometimes, maybe even often, this does not matter at all. However...

In recently written code of my own, I prefer to use `int` as the underlying type of the `const`s with a `String()` defined for it with the right effect. IOW, I prefer to spend extra time (typically) in `formatted printing`/string operations, which are supposed to be rare and heavy anyway, then in `map lookup`s which tend to be more frequent and therefore should be more efficient.

pkg/rdt/rdt.go
marquiz: Hmm, good catch and good question. Yes, it was a deliberate change, but, in some ancient version of the patchset where it was convenient to have the quick lookup from map. However, that property is not used, anymore, and, I think it is probably better to stay in the sorted slice format. I'll change this accordingly

pkg/rdt/rdt.go
marquiz: Thanks for pointing this out. I just blindly copy-pasted the old badly-written code without thinking. Will fix this...

pkg/rdt/rdt.go
marquiz: I really don't know. This construct was just to make the code more future-proof, as the kernel interface could easily be extended. We might get L2 monitoring in the future, for example. This is pure speculation, though.

Should I change this, WDYT?

pkg/rdt/rdt.go
marquiz: Ach, another good catch 👍 A stranded leftover interface from some earlier version. I'll drop it

pkg/rdt/rdt.go
klihub: No, don't change it.

pkg/rdt/rdt.go
marquiz: Changed back to sorted slice.

I also change `CtrlGroup.GetMonGroups()` similarly

pkg/rdt/rdt.go
marquiz: Fixed

pkg/rdt/rdt.go
marquiz: Fixed: I dropped the Monitor interface.

pkg/rdt/rdt.go
marquiz: Good point. And, yes, as an old C developer I've noticed this pattern and it's implications, too 😄 I've adapted this habit from K8s et al I guess. But, I think it's handy in some cases by reducing the LOC.

In this particular case the string type is used in a very non-performance-critical code path (and, I don't see it ending up in such a code path). Thus, I let this be as it is, for now.

marquiz:
> Additional note, which AFAICT can't be reported in any intelligent manner with github reviews is a typo in the commit message of [38c6879](https://github.com/intel/cri-resource-manager/commit/38c68793c2a3ead3fbcb42cbb856c22080e1a1b1): 'functionalitu' should be 'functionality'...

Almost forgot, but, now fixed

mythi:
the registration now happens by importing `pkg/rdt/` so is there a dependency in order the of when the controllers and events are getting registered?

marquiz:
Umm, the registration happens in `init()` so it's registered before `main()` starts. Why would the otder matter??

mythi:
Ah, right. I asked before thinking through the init. No issues.

PR #225, opened 2020-03-27
instrumentation: create new Jaeger Exporter on (re-start)

marquiz:
Otherwise the Jaeger related config changes have no effect.


klihub:
Hmm... is that the case event with #211 applied ?

marquiz:
Actually prolly not. Closing this one

Issue #227, opened 2020-03-27
BlockIO: Update weight/throttling on running containers on blockio class and class mapping config changes


askervin:
Currently new configuration is applied only on new containers created after the configuration has become effective.

msivosuo:
implemented by #235

PR #235, opened 2020-04-08
blockio: apply new and updated configs to running containers

askervin:
- Add cgroups file-to-oci blockio conversion.
- Add cgroups configuration transformation from whatever existing one
  to wanted configuration by adding/modifying/removing parameters.
- Hook resetting blockio parameters to CRI-RM launch and config change.
- Add unit tests.

### DIFF
--- a/pkg/cgroups/cgroupblkio.go
+++ b/pkg/cgroups/cgroupblkio.go
@@ -16,9 +16,11 @@ package cgroups
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
 
@@ -50,30 +52,74 @@ var blkioThrottleWriteIOPSFiles = []string{"blkio.throttle.write_iops_device"}
 //  other |  Write to cgroups, sets the value
 type OciBlockIOParameters struct {
 	Weight                  int64
-	WeightDevice            []OciWeightDeviceParameters
-	ThrottleReadBpsDevice   []OciRateDeviceParameters
-	ThrottleWriteBpsDevice  []OciRateDeviceParameters
-	ThrottleReadIOPSDevice  []OciRateDeviceParameters
-	ThrottleWriteIOPSDevice []OciRateDeviceParameters
+	WeightDevice            OciDeviceWeights
+	ThrottleReadBpsDevice   OciDeviceRates
+	ThrottleWriteBpsDevice  OciDeviceRates
+	ThrottleReadIOPSDevice  OciDeviceRates
+	ThrottleWriteIOPSDevice OciDeviceRates
 }
 
-// OciWeightDeviceParameters contains values for
+// OciDeviceWeight contains values for
 // - blkio.[io-scheduler].weight
-type OciWeightDeviceParameters struct {
+type OciDeviceWeight struct {
 	Major  int64
 	Minor  int64
 	Weight int64
 }
 
-// OciRateDeviceParameters contains values for
+// OciDeviceRate contains values for
 // - blkio.throttle.read_bps_device
 // - blkio.throttle.write_bps_device
 // - blkio.throttle.read_iops_device
 // - blkio.throttle.write_iops_device
-type OciRateDeviceParameters struct {
+type OciDeviceRate struct {
 	Major int64
 	Minor int64
 	Rate  int64
+}
+
+// OciDeviceWeights contains weights for devices
+type OciDeviceWeights []OciDeviceWeight
+
+// OciDeviceRates contains throttling rates for devices
+type OciDeviceRates []OciDeviceRate
+
+// OciDeviceParameters interface provides functions common to OciDeviceWeights and OciDeviceRates
+type OciDeviceParameters interface {
+	Append(maj, min, val int64)
+	Update(maj, min, val int64)
+}
+
+// Append appends (major, minor, value) to OciDeviceWeights slice.
+func (w *OciDeviceWeights) Append(maj, min, val int64) {
+	*w = append(*w, OciDeviceWeight{Major: maj, Minor: min, Weight: val})
+}
+
+// Append appends (major, minor, value) to OciDeviceRates slice.
+func (r *OciDeviceRates) Append(maj, min, val int64) {
+	*r = append(*r, OciDeviceRate{Major: maj, Minor: min, Rate: val})
+}
+
+// Update updates device weight in OciDeviceWeights slice, or appends it if not found.
+func (w *OciDeviceWeights) Update(maj, min, val int64) {
+	for index, devWeight := range *w {
+		if devWeight.Major == maj && devWeight.Minor == min {
+			(*w)[index].Weight = val
+			return
+		}
+	}
+	w.Append(maj, min, val)
+}
+
+// Update updates device rate in OciDeviceRates slice, or appends it if not found.
+func (r *OciDeviceRates) Update(maj, min, val int64) {
+	for index, devRate := range *r {
+		if devRate.Major == maj && devRate.Minor == min {
+			(*r)[index].Rate = val
+			return
+		}
+	}
+	r.Append(maj, min, val)
 }
 
 // NewOciBlockIOParameters creates new OciBlockIOParameters instance.
@@ -83,18 +129,18 @@ func NewOciBlockIOParameters() OciBlockIOParameters {
 	}
 }
 
-// NewOciWeightDeviceParametes creates new OciWeightDeviceParameters instance.
-func NewOciWeightDeviceParametes() OciWeightDeviceParameters {
-	return OciWeightDeviceParameters{
+// NewOciDeviceWeight creates new OciDeviceWeight instance.
+func NewOciDeviceWeight() OciDeviceWeight {
+	return OciDeviceWeight{
 		Major:  -1,
 		Minor:  -1,
 		Weight: -1,
 	}
 }
 
-// NewOciRateDeviceParameters creates new OciRateDeviceParameters instance.
-func NewOciRateDeviceParameters() OciRateDeviceParameters {
-	return OciRateDeviceParameters{
+// NewOciDeviceRate creates new OciDeviceRate instance.
+func NewOciDeviceRate() OciDeviceRate {
+	return OciDeviceRate{
 		Major: -1,
 		Minor: -1,
 		Rate:  -1,
@@ -104,6 +150,131 @@ func NewOciRateDeviceParameters() OciRateDeviceParameters {
 // GetBlkioDir returns the cgroups blkio controller directory.
 func GetBlkioDir() string {
 	return blkioCgroupDir
+}
+
+type devMajMin struct {
+	Major int64
+	Minor int64
+}
+
+// ResetBlkioParameters adds new, changes existing and removes missing blockIO parameters in cgroupsDir
+func ResetBlkioParameters(cgroupsDir string, blockIO OciBlockIOParameters) error {
+	var errors *multierror.Error
+	oldBlockIO, getErr := GetBlkioParameters(cgroupsDir)
+	errors = multierror.Append(errors, getErr)
+	newBlockIO := NewOciBlockIOParameters()
+	newBlockIO.Weight = blockIO.Weight
+	// Set new device weights
+	seenDev := map[devMajMin]bool{}
+	for _, ociWDP := range blockIO.WeightDevice {
+		seenDev[devMajMin{ociWDP.Major, ociWDP.Minor}] = true
+		newBlockIO.WeightDevice = append(newBlockIO.WeightDevice, ociWDP)
+	}
+	// Reset old device weights that were missing from blockIO.WeightDevice
+	for _, ociWDP := range oldBlockIO.WeightDevice {
+		if !seenDev[devMajMin{ociWDP.Major, ociWDP.Minor}] {
+			newBlockIO.WeightDevice = append(newBlockIO.WeightDevice, OciDeviceWeight{ociWDP.Major, ociWDP.Minor, 0})
+		}
+	}
+	newBlockIO.ThrottleReadBpsDevice = resetDevRates(oldBlockIO.ThrottleReadBpsDevice, blockIO.ThrottleReadBpsDevice)
+	newBlockIO.ThrottleWriteBpsDevice = resetDevRates(oldBlockIO.ThrottleWriteBpsDevice, blockIO.ThrottleWriteBpsDevice)
+	newBlockIO.ThrottleReadIOPSDevice = resetDevRates(oldBlockIO.ThrottleReadIOPSDevice, blockIO.ThrottleReadIOPSDevice)
+	newBlockIO.ThrottleWriteIOPSDevice = resetDevRates(oldBlockIO.ThrottleWriteIOPSDevice, blockIO.ThrottleWriteIOPSDevice)
+	errors = multierror.Append(errors, SetBlkioParameters(cgroupsDir, newBlockIO))
+	return errors.ErrorOrNil()
+}
+
+// resetDevRates adds wanted rate parameters to new and resets unwated rates
+func resetDevRates(old, wanted []OciDeviceRate) []OciDeviceRate {
+	new := []OciDeviceRate{}
+	seenDev := map[devMajMin]bool{}
+	for _, rdp := range wanted {
+		new = append(new, rdp)
+		seenDev[devMajMin{rdp.Major, rdp.Minor}] = true
+	}
+	for _, rdp := range old {
+		if !seenDev[devMajMin{rdp.Major, rdp.Minor}] {
+			new = append(new, OciDeviceRate{rdp.Major, rdp.Minor, 0})
+		}
+	}
+	return new
+}
+
+// GetBlkioParameters returns OCI BlockIO parameters from files in cgroups blkio controller directory.
+func GetBlkioParameters(cgroupsDir string) (OciBlockIOParameters, error) {
+	var errors *multierror.Error
+	blockIO := NewOciBlockIOParameters()
+	content, err := readFromFileInDir(cgroupsDir, blkioWeightFiles)
+	if err == nil {
+		weight, err := strconv.ParseInt(strings.TrimSuffix(content, "\n"), 10, 64)
+		if err == nil {
+			blockIO.Weight = weight
+		} else {
+			errors = multierror.Append(errors, fmt.Errorf("parsing weight from %#v failed: %w", content, err))
+		}
+	} else {
+		errors = multierror.Append(errors, err)
+	}
+	errors = multierror.Append(errors, readOciDeviceParameters(cgroupsDir, blkioWeightDeviceFiles, &blockIO.WeightDevice))
+	errors = multierror.Append(errors, readOciDeviceParameters(cgroupsDir, blkioThrottleReadBpsFiles, &blockIO.ThrottleReadBpsDevice))
+	errors = multierror.Append(errors, readOciDeviceParameters(cgroupsDir, blkioThrottleWriteBpsFiles, &blockIO.ThrottleWriteBpsDevice))
+	errors = multierror.Append(errors, readOciDeviceParameters(cgroupsDir, blkioThrottleReadIOPSFiles, &blockIO.ThrottleReadIOPSDevice))
+	errors = multierror.Append(errors, readOciDeviceParameters(cgroupsDir, blkioThrottleWriteIOPSFiles, &blockIO.ThrottleWriteIOPSDevice))
+	return blockIO, errors.ErrorOrNil()
+}
+
+// readOciDeviceParameters parses device lines used for weights and throttling rates
+func readOciDeviceParameters(baseDir string, filenames []string, params OciDeviceParameters) error {
+	var errors *multierror.Error
+	contents, err := readFromFileInDir(baseDir, filenames)
+	if err != nil {
+		return err
+	}
+	for _, line := range strings.Split(contents, "\n") {
+		// Device weight files may have "default NNN" line at the beginning. Skip it.
+		if line == "" || strings.HasPrefix(line, "default ") {
+			continue
+		}
+		// Expect syntax MAJOR:MINOR VALUE
+		devVal := strings.Split(line, " ")
+		if len(devVal) != 2 {
+			errors = multierror.Append(errors, fmt.Errorf("invalid line %q, single space expected", line))
+			continue
+		}
+		majMin := strings.Split(devVal[0], ":")
+		if len(majMin) != 2 {
+			errors = multierror.Append(errors, fmt.Errorf("invalid line %q, single colon expected before space", line))
+			continue
+		}
+		major, majErr := strconv.ParseInt(majMin[0], 10, 64)
+		minor, minErr := strconv.ParseInt(majMin[1], 10, 64)
+		value, valErr := strconv.ParseInt(devVal[1], 10, 64)
+		if majErr != nil || minErr != nil || valErr != nil {
+			errors = multierror.Append(errors, fmt.Errorf("invalid number when parsing \"major:minor value\" from \"%s:%s %s\"", majMin[0], majMin[1], devVal[1]))
+			continue
+		}
+		params.Append(major, minor, value)
+	}
+	return errors.ErrorOrNil()
+}
+
+// readFromFileInDir returns content from the first successfully read file.
+func readFromFileInDir(baseDir string, filenames []string) (string, error) {
+	var errors *multierror.Error
+	// If reading all the files fails, return list of read errors.
+	for _, filename := range filenames {
+		filepath := filepath.Join(baseDir, filename)
+		content, err := currentPlatform.readFromFile(filepath)
+		if err == nil {
+			return content, nil
+		}
+		errors = multierror.Append(errors, err)
+	}
+	err := errors.ErrorOrNil()
+	if err != nil {
+		return "", fmt.Errorf("could not read any of files %q: %w", filenames, err)
+	}
+	return "", nil
 }
 
 // SetBlkioParameters writes OCI BlockIO parameters to files in cgroups blkio contoller directory.
@@ -158,6 +329,7 @@ func writeToFileInDir(baseDir string, filenames []string, content string) error 
 
 // platformInterface includes functions that access the system. Enables mocking the platform.
 type platformInterface interface {
+	readFromFile(filename string) (string, error)
 	writeToFile(filename string, content string) error
 }
 
@@ -166,6 +338,12 @@ type defaultPlatform struct{}
 
 // currentPlatform defines which platformInterface is used: defaultPlatform or a mock, for instance.
 var currentPlatform platformInterface = defaultPlatform{}
+
+// readFromFile returns file contents as a string.
+func (dpm defaultPlatform) readFromFile(filename string) (string, error) {
+	content, err := ioutil.ReadFile(filename)
+	return string(content), err
+}
 
 // writeToFile writes content to an existing file.
 func (dpm defaultPlatform) writeToFile(filename string, content string) error {
--- a/pkg/cri/resource-manager/control/blockio/blockio.go
+++ b/pkg/cri/resource-manager/control/blockio/blockio.go
@@ -17,6 +17,8 @@ package blockio
 import (
 	"fmt"
 
+	"github.com/hashicorp/go-multierror"
+
 	"github.com/intel/cri-resource-manager/pkg/blockio"
 	"github.com/intel/cri-resource-manager/pkg/config"
 	"github.com/intel/cri-resource-manager/pkg/cri/client"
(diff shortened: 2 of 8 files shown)


askervin:
@klihub, this patch implements #227, and would be ready for review. There's one thing that I feel pretty clumsy: how pkg/blockio.go is `Initialized()` in order to get access cache and call `BlockIOClass()` from the real controller (`control/blockio`) later on. I wonder if you'd have good ideas to make it smoother.

klihub:
> @klihub, this patch implements #227, and would be ready for review. There's one thing that I feel pretty clumsy: how pkg/blockio.go is `Initialized()` in order to get access cache and call `BlockIOClass()` from the real controller (`control/blockio`) later on. I wonder if you'd have good ideas to make it smoother.

I haven't read through the other parts yet, but I took a sneak peek at that piece. And I see what you mean... You basically 'push down' the cache (and the controller) during initialization to the low-level block I/O code so that you can implement `reconfigureRunningContainers()` there. That one then  needs to call also back to the controller to map the class so it can do its deed. It doesn't look like the right way to me either.

I think this should be the other way around, like during the normal control flow: the controller calling into the low-level block I/O code. Looking at the code, it looks like the best place to do that is the existing configuration notification handler of block I/O controller itself. Currently it is an effective no-op, simply logging the fact of a configuration update and then always accepting it by returning nil.

You can instead hook it up to the low-level block I/O code's configuration update event, to make sure that it has validated the configuration, and then just do the same thing you now do in `reconfigureRunningControllers()` but without jumping through the extra hoops to get the containers and the right class.

You need to change the block I/O controllers config registration to something like:

```
// in pkg/blockio:
const (
  ConfigModuleName = "blockio" // and also use this instead of config.Register("blockio"...)
)

// in the block I/O controller
func init() {
  config.Register("resource-manager.blockio", configHelp, opt, defaultOptions)
  config.GetModule(blockio.ConfigModuleName).AddNotify(getBlockIOController().(*blockioctl).configNotify)
}
```

Then you can for instance move `reconfigureRunningContaines()` from the low-level package to the controller and call it from `configNotify()`.

And BTW, *I think* if you change the return type of `getBlockIOController()` to `*blockioctl` from `control.Controller`, then you can get rid of the type assertion above.

As a final note, if we run into problems with the approach described above, or if this turns out to be something which is needed by most controllers then it's a better idea to add a new explicit controller hook (e.g. `Reconfigure()`) to `control.Controller` to handle controller-specific post-configuration actions.

askervin:
@klihub , that was a very good idea, thanks! That part looks good now, I think.

When I tested this patch (`-force-config` + `kill -HUP ...`), I noticed two behaviors that I'm still not perfectly happy about:
1. At start `configNotify` gets called before controller's `Start()`. Therefore there's no access to `cache` yet through the `ctl` singleton, and therefore parameters for running containers are not reset at launch.
2.  At start and at `SIGHUP`, `configNotify` gets called twice, as it is the callback for both class mapping and class definition callbacks. It makes sense to reconfigure running containers whenever either of these configurations gets updated. But now we get one extra call in these cases.

I'm wondering if `configNotify` callbacks should be set only at `Start()` instead of `init()`, and maybe `Start()` should call `reconfigureRunningContainers()` anyway without configuration notifications...

sample-configs/blockio.cfg
```diff
@@ -12,6 +12,10 @@
 policy:
   Active: none
 
+logger:
+  Debug: on:all
+  Enable: on:all
+
```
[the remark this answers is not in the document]
askervin: This is deliberate. I was thinking that this helps the user, who tries out the raw example config (`cri-resmgr -force-config blockio.cfg`), to immediately see how the configuration is handled and how it goes to down to cgroups.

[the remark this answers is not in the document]
askervin: (I'm not marking this as resolved, as I'm not making any change yet. Let me know if you'd prefer cleaner `cri-resmgr` output - with the drawback that you cannot see which cgroups container got updated and how.)

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -134,26 +134,51 @@ func (ctl *blockioctl) BlockIOClass(c cache.Container) string {
...
 }
 
-// Register us as a controller.
+// init registers this controller and sets configuration change handling.
 func init() {
 	control.Register(BlockIOController, "Block I/O controller", getBlockIOController())
+	// high-level configuration: pod class to blockio class mapping
+	config.GetModule(ConfigModuleName).AddNotify(getBlockIOController().configNotify)
+	// low-level configuration: blockio class definitions
+	config.GetModule(blockio.ConfigModuleName).AddNotify(getBlockIOController().configNotify)
```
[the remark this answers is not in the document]
askervin: I felt that important-to-know functionality got scattered and hidden with that split in this new arrangement. With the original split, single callback function (`configNotify`) in one file (`blockio.go`) would be defined as a callback for one configuration in `flags.go` and for another configuration in another file (`blockio.go`). As a developer I'd prefer reading code where all callbacks to a function are set in one place. And it doesn't hurt if that place is close to the function.

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -134,26 +134,51 @@ func (ctl *blockioctl) BlockIOClass(c cache.Container) string {
...
+	// the updated configuration, therefore silently ignored.
+	ctl.reconfigureRunningContainers()
 	return nil
 }
 
+// reconfigureRunningContainers force setting current blockio configuration to all containers running on the node
+func (ctl *blockioctl) reconfigureRunningContainers() error {
+	var errors *multierror.Error
+	if ctl.cache == nil {
+		return nil
```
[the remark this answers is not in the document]
askervin: I think that would make the control flow more readable.

So the order in the controller point of view would be:

1. `init()` by Go, no cache available
2. `Create(cache.Cache)`
3. All configuration notifications that are bound at `init` or possibly in `Create`. In this case `reconfigureRunningContainers()` could not called between `Create()` and `Start()`, because not all (both) necessary configurations are there.
4. `Start()` - here we know that all configurations are in place. In this case this would be a good spot to call `reconfigureRunningContainers()`.

So I think that I should still add `reconfigureRunningContainers()` to `Start()` anyway now, because currently `ctl.cache == nil` indicates that the controller has not been `Start()`ed, so `configNotify()` has not reconfigured anything yet.

klihub:
> @klihub , that was a very good idea, thanks! That part looks good now, I think.
> 
> When I tested this patch (`-force-config` + `kill -HUP ...`), I noticed two behaviors that I'm still not perfectly happy about:
> 
> 1. At start `configNotify` gets called before controller's `Start()`. Therefore there's no access to `cache` yet through the `ctl` singleton, and therefore parameters for running containers are not reset at launch.

Yes. This is something that would be easy to fix in the infra by adding an explicit `Create(cache.Cache)` to the controller interface. That would get called in the startup sequence before we acquire the first configuration.

> 2. At start and at `SIGHUP`, `configNotify` gets called twice, as it is the callback for both class mapping and class definition callbacks. It makes sense to reconfigure running containers whenever either of these configurations gets updated. But now we get one extra call in these cases.

Ah, I missed this in my initial review...

So why do you need now two notifications now ? Or to put it another way: why do you hook up that single function twice, to two different config modules ? After all, in your initial PR the controller-level notification was a no-op: it only logged that the configuration was updated.

The idea was that instead of being a no-op you hook that as a 'foreign' callback to the low-level blockio package's configuration update event (plus and move the code that fiddles with containers and the cache out of the low level package (which you did).

Everything should work correctly IMO, if you simply drop the extra notification hookup. If you need to do data validation, then you hook up a real config notify callback for just that... a different function. But I thought you didn't need it since you did not do it in the original PR with the no-op update callback.

> 
> I'm wondering if `configNotify` callbacks should be set only at `Start()` instead of `init()`, and maybe `Start()` should call `reconfigureRunningContainers()` anyway without configuration notifications...

Some of the policies do that, but I'm not sure if that's really necessary/the best way go here.

askervin:
@klihub , I'm hooking up to both configuration changes as  a change in both QoS-to-blockio class mapping and blockio class definitions is a good reason to reconfigure running containers. Example mapping change: switch "BestEffort" pods from "io-prio-low" to "raid-broken-throttle-writes". You're correct, original patch did not trigger reconfiguration on this change. It was a bug.

sample-configs/blockio.cfg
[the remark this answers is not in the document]
klihub: Hmm... that `Enable: on:all` should be the default even if you leave that line out.
And about the changes being visible to the user... shouldn't the configuration changes simply be logged with `Info()` so they get always logged without any changes to the configuration ?

pkg/cri/resource-manager/control/blockio/blockio.go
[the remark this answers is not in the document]
klihub: Yes. Having it in `Start()` is a good idea.

sample-configs/blockio.cfg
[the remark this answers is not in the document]
askervin: `Info()` level tells that new blockio configuration has been applied.

`Debug()` level tells the pretty name of each configured container down to the cgroups directory and applied configuration values:

```
D: [     blockio    ] configure container "bb-burst:busybox" blockio class "Default"
D: [   cgroupblkio  ] configuring cgroups blkio controller in directory "/sys/fs/cgroup/blkio/kubepods/burstable/podc919e8bb-7b17-4fce-ab36-182a3d1e4793/7839a9d9101d8
59792eadebfb049086bbcb208a30794fdc68dee0bbdc42a665d" with parameters {Weight:100 WeightDevice:[] ThrottleReadBpsDevice:[] ThrottleWriteBpsDevice:[] ThrottleReadIOPSDe
vice:[] ThrottleWriteIOPSDevice:[]}
```

Yep, Enable was not needed => removed.

klihub:
> @klihub , I'm hooking up to both configuration changes as a change in both QoS-to-blockio class mapping and blockio class definitions is a good reason to reconfigure running containers. Example mapping change: switch "BestEffort" pods from "io-prio-low" to "raid-broken-throttle-writes". You're correct, original patch did not trigger reconfiguration on this change. It was a bug.

About the configuration and the update callbacks.

 1. If *any* of the configuration changes, *all* config update notification callbacks will be invoked.
 2. Callbacks are always invoked in the *same order*.
 3. By the time the *first* callback is invoked, the (primary) configuration data has been *physically updated in all* configuration modules.

Also by design we never try to hide the fact of reconfiguration from a module. Actually very much the opposite: the invocation of a config callback does not mean that the related data has changed. It simply means that new configuration data has been acquired, with or without changes to that particular component.

So then the idea/duty of a config callback is
 1. validate current (potentially updated) given data (as much as possible)
 2. if the user config data is not the runtime config data, update the latter form the former. For instance, RDT goes out of its way to make configuration user friendly, so it needs to do quite a bit of heavy lifting/chewing on the data when it is reconfigured. Therefore its configuration and runtime configuration data are not the same.

So in our case here, from the primary data point of view it does not matter where you hook yourself up: when the callback comes in, all primary data has been updated. So you can safely hook yourself up as a foreign notification callback to the low-level blockio package configuration and do everything from that, including verification of your own data.

sample-configs/blockio.cfg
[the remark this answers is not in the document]
klihub: Ok, fair enough.

askervin:
Oh, my non-verified assumption on config callbacks was wrong. Thanks for the clarification!

I dropped the unnecessary config callback registration. And I'm starting to have a good feeling about this patch. :)

Btw, one more thing that works somehow counter-intuitively. The for loop in `reconfigureRunningContainers()` is supposed to visit every container in every pod once:

```go
for _, pod := range ctl.cache.GetPods() {
	for _, c := range pod.GetContainers() {
		class := ctl.BlockIOClass(c)
		err := blockio.SetContainerClass(c, class)
...
```

However, the debug output shows that the same cgroups directory is visited twice every time:

```
D: [     blockio    ] configure container "etcd-single:etcd" blockio class "LowPrioThrottled"
D: [   cgroupblkio  ] configuring cgroups blkio controller in directory "/sys/fs/cgroup/blkio/kubepods/besteffort/pod02f2e7755a9ca0289de18c1a050efac9/f02dda572553d5adf8e5637aa6561ac92c523c5f687e57b3d4c3369ff58f9453" with parameters {Weight:80 WeightDevice:[{Major:253 Minor:0 Weight:50}] ThrottleReadBpsDevice:[{Major:253 Minor:0 Rate:50000000}] ThrottleWriteBpsDevice:[{Major:253 Minor:0 Rate:10000000}] ThrottleReadIOPSDevice:[{Major:253 Minor:0 Rate:10000}] ThrottleWriteIOPSDevice:[{Major:253 Minor:0 Rate:5000}]}
D: [     blockio    ] configure container "etcd-single:etcd" blockio class "LowPrioThrottled"
D: [   cgroupblkio  ] configuring cgroups blkio controller in directory "/sys/fs/cgroup/blkio/kubepods/besteffort/pod02f2e7755a9ca0289de18c1a050efac9/f02dda572553d5adf8e5637aa6561ac92c523c5f687e57b3d4c3369ff58f9453" with parameters {Weight:80 WeightDevice:[{Major:253 Minor:0 Weight:50}] ThrottleReadBpsDevice:[{Major:253 Minor:0 Rate:50000000}] ThrottleWriteBpsDevice:[{Major:253 Minor:0 Rate:10000000}] ThrottleReadIOPSDevice:[{Major:253 Minor:0 Rate:10000}] ThrottleWriteIOPSDevice:[{Major:253 Minor:0 Rate:5000}]}
```

Debugging shows that `pod.GetContainers()` indeed returns slice of two exactly-the-same looking container pointers:

```
*github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache.container {
cache: *(*"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache.cache")(0xc00038c090),
ID: "1eea46acbdf1720a2563edd202186baefc2cace74fd254b901a002d0cadcc0a1",
PodID: "bd298b0de6369bd2b4c678897b7c05b308ba924aa7396b1ddfeb5303a1bb662f",
CacheID: "4a0be0d8-352a-4f4e-a82a-d8cce071284a:coredns",
...
```

klihub:
> Oh, my non-verified assumption on config callbacks was wrong. Thanks for the clarification!

No prob, I guessed it from your justification for the double registration ;-)

> 
> I dropped the unnecessary config callback registration. And I'm starting to have a good feeling about this patch. :)
> 
> Btw, one more thing that works somehow counter-intuitively. The for loop in `reconfigureRunningContainers()` is supposed to visit every container in every pod once:
> 
> ```go
> for _, pod := range ctl.cache.GetPods() {
> 	for _, c := range pod.GetContainers() {
> 		class := ctl.BlockIOClass(c)
> 		err := blockio.SetContainerClass(c, class)
> ...
> ```
> 
> However, the debug output shows that the same cgroups directory is visited twice every time:
> 
> ```
> D: [     blockio    ] configure container "etcd-single:etcd" blockio class "LowPrioThrottled"
> D: [   cgroupblkio  ] configuring cgroups blkio controller in directory "/sys/fs/cgroup/blkio/kubepods/besteffort/pod02f2e7755a9ca0289de18c1a050efac9/f02dda572553d5adf8e5637aa6561ac92c523c5f687e57b3d4c3369ff58f9453" with parameters {Weight:80 WeightDevice:[{Major:253 Minor:0 Weight:50}] ThrottleReadBpsDevice:[{Major:253 Minor:0 Rate:50000000}] ThrottleWriteBpsDevice:[{Major:253 Minor:0 Rate:10000000}] ThrottleReadIOPSDevice:[{Major:253 Minor:0 Rate:10000}] ThrottleWriteIOPSDevice:[{Major:253 Minor:0 Rate:5000}]}
> D: [     blockio    ] configure container "etcd-single:etcd" blockio class "LowPrioThrottled"
> D: [   cgroupblkio  ] configuring cgroups blkio controller in directory "/sys/fs/cgroup/blkio/kubepods/besteffort/pod02f2e7755a9ca0289de18c1a050efac9/f02dda572553d5adf8e5637aa6561ac92c523c5f687e57b3d4c3369ff58f9453" with parameters {Weight:80 WeightDevice:[{Major:253 Minor:0 Weight:50}] ThrottleReadBpsDevice:[{Major:253 Minor:0 Rate:50000000}] ThrottleWriteBpsDevice:[{Major:253 Minor:0 Rate:10000000}] ThrottleReadIOPSDevice:[{Major:253 Minor:0 Rate:10000}] ThrottleWriteIOPSDevice:[{Major:253 Minor:0 Rate:5000}]}
> ```
> 
> Debugging shows that `pod.GetContainers()` indeed returns slice of two exactly-the-same looking container pointers:
> 
> ```
> *github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache.container {
> cache: *(*"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache.cache")(0xc00038c090),
> ID: "1eea46acbdf1720a2563edd202186baefc2cace74fd254b901a002d0cadcc0a1",
> PodID: "bd298b0de6369bd2b4c678897b7c05b308ba924aa7396b1ddfeb5303a1bb662f",
> CacheID: "4a0be0d8-352a-4f4e-a82a-d8cce071284a:coredns",
> ...
> ```

The containers are hashed twice, once by CRI ID and once by our internal cache ID. Looks like pod.GetContainers() fails to properly filter duplicates from the returned slice. I'll take a look at it.

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -134,26 +135,48 @@ func (ctl *blockioctl) BlockIOClass(c cache.Container) string {
...
+	}
+	for _, pod := range ctl.cache.GetPods() {
+		for _, c := range pod.GetContainers() {
```
klihub, line 161: Can you switch this loop to use `cache.GetContainers()`. As you pointed out `pod.GetContainers()` fails to properly filter duplicates from the returned slice. I'll fix that but `cache.GetContainers()` does not have the same problem and since you don't do anything with pods it looks more appropriate anyway for this case...

askervin: Yeah... after having a look inside `pod.GetContainers`, it is definitely better to iterate directly on cache level. Fixed.

pkg/cgroups/cgroupblkio.go
```diff
@@ -106,6 +109,143 @@ func GetBlkioDir() string {
...
+	}
+	errors = multierror.Append(errors, readDevWeightsRates(cgroupsDir, blkioWeightDeviceFiles, &blockIO.WeightDevice, nil))
+	errors = multierror.Append(errors, readDevWeightsRates(cgroupsDir, blkioThrottleReadBpsFiles, nil, &blockIO.ThrottleReadBpsDevice))
+	errors = multierror.Append(errors, readDevWeightsRates(cgroupsDir, blkioThrottleWriteBpsFiles, nil, &blockIO.ThrottleWriteBpsDevice))
+	errors = multierror.Append(errors, readDevWeightsRates(cgroupsDir, blkioThrottleReadIOPSFiles, nil, &blockIO.ThrottleReadIOPSDevice))
+	errors = multierror.Append(errors, readDevWeightsRates(cgroupsDir, blkioThrottleWriteIOPSFiles, nil, &blockIO.ThrottleWriteIOPSDevice))
+	return blockIO, errors.ErrorOrNil()
+}
+
+func readDevWeightsRates(baseDir string, filenames []string, outWeights *[]OciWeightDeviceParameters, outRates *[]OciRateDeviceParameters) error {
```
marquiz, line 183: This interface looks a bit awkward. How about replacing `outWeights` and `outRates` with one empty interface parameter and doing a type switch in the function.

Or, is there really need to have two separate types `OciWeightDeviceParameters` and `OciRateDeviceParameters`? At a quick glance they could be replaced with one `OciDeviceParameters`, simplifying the code quite a bit.

WDYT?

pkg/cgroups/cgroupblkio.go
```diff
@@ -106,6 +109,143 @@ func GetBlkioDir() string {
...
+			major, _ := strconv.ParseInt(submatches[1], 10, 64)
+			minor, _ := strconv.ParseInt(submatches[2], 10, 64)
+			value, _ := strconv.ParseInt(submatches[3], 10, 64)
+			mmvs = append(mmvs, majorMinorValue{
+				major: major,
+				minor: minor,
+				value: value,
+			})
+		}
+	}
```
marquiz, line 226: Maybe it's just me, but, I dislike regexes in simple cases like these. SImple strings split should be good, and, I'd print an error/warning if we encounter unparsable input.

WDYT?

pkg/cri/resource-manager/control/blockio/blockio.go
```diff
@@ -134,26 +135,46 @@ func (ctl *blockioctl) BlockIOClass(c cache.Container) string {
...
 func (ctl *blockioctl) configNotify(event config.Event, source config.Source) error {
-	// BlockIO class mapping in opt (in flags.go) has changed.
-	// It will affect only new pods. cgroupsblkio parameters of running pods
-	// is not changed.
-	log.Info("class mapping configuration updated")
+	ignoreErrors := (event == config.RevertEvent)
+	err := blockio.UpdateOciConfig(ignoreErrors)
+	if err != nil && !ignoreErrors {
+		return err
+	}
```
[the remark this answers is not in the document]
askervin: Good point. When `UpdateOciConfig` returns an error, it has not updated the configuration, so applying the configuration doesn't make sense.

=> removed `&& !ignoreErrors`

sample-configs/blockio.cfg
```diff
@@ -12,6 +12,9 @@
 policy:
   Active: none
 
+logger:
+  Debug: on:all
+
```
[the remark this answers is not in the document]
askervin: This was also asked by @klihub. Reason for this is to show what is actually happening to users who try `cri-resmgr -force-config blockio.cfg`.

https://github.com/intel/cri-resource-manager/pull/235#discussion_r408708188

pkg/cgroups/cgroupblkio.go
```diff
@@ -106,6 +109,143 @@ func GetBlkioDir() string {
 	return blkioCgroupDir
 }
 
+type devMM struct {
```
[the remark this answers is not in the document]
askervin: For consitency, substituted:
`devMM` -> `devMajMin`
`majorMinorValue` -> `majMinVal` (struct name)
`MajorMinorValue` -> `MajMinVal` (function name, regexp var name)

sample-configs/blockio.cfg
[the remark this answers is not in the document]
askervin: (I'm not marking this resolved as I'm not taking any action - yet.)

pkg/cgroups/cgroupblkio.go
askervin: I agree. However, I'm strictly following here structures in [OCI blockio spec](https://github.com/tianon/oci-runtime-spec/blob/master/config-linux.md#block-io) in order to keep the cgroups interface OCI-like. OCI has different structs for blkioWeightDevice and blkioThrottle*. The price is extra code complexity.

This particular interface is an alternative to my first solution that had separate `readDevWeights` and `readDevRates` functions that were practically the same, as you can imagine.

askervin: Please mark conversation resolved if this explanation makes this code bearable in your eyes.

Now that we're not really interested on `leafWeight`, we really could get rid of separate `rate` and `weight` structures and simplify the code. But maybe that's another patch?

pkg/cgroups/cgroupblkio.go
askervin: Ok, rewrote the parser without regexps and with error messages on unexpected lines.

pkg/cgroups/cgroupblkio.go
marquiz: What do you think about the empty interface parameter I suggested?

Or then drop `readDevWeightsRates` for good and do smth like:
```
type OciDeviceParameters interface {
        append(maj, min, val,int64)
}

type OciWeightDeviceParameter struct {
        Major  int64
        Minor  int64
        Weight int64
}

/*
* Add corresponding definitions for OciRateDeviceParameters
*/

type OciWeightDeviceParameters []OciWeightDeviceParameter

func (p *OciWeightDeviceParameters) append(maj, min, val int64) { *p = append(*p, OciWeightDeviceParameter{Major: maj, Minor: min, Weight: val} }

func readMajorMinorValues(baseDir string, filenames []string, OciDeviceParameters) error {
...
}

I think this would be a bit more go-like
```

askervin:
Added unit tests to parse some negative cases, and verified this is still working. :)
Ready to go.

pkg/cgroups/cgroupblkio.go
klihub: If we start walking down that path, then I'd rather also rename the current `OciWeightDeviceParameters` and `OciRateDeviceParameters` to `OciDeviceWeight` and `OciDeviceRate` correspondingly. Then define plural-named slices of these, `OciDeviceWeights` and `OciDeviceRates`, and have them support the interface @marquiz  suggested above. Finally, I'd drop/fold `readMajMinVals()` into `readDevWeightsRates()` and rename it to `readOciDeviceParameters()`. Something like this (sans the worst typos and thinkos, of course):

```
type OciDeviceWeight struct {
    Major  int64
    Minor  int64
    Weight int64
}

type OciDeviceRate struct {
    Major  int64
    Minor  int64
    Rate   int64
}

type OciDeviceParameters interface {
    append(maj, min, val int64)
}

type OciDeviceWeights []OciDeviceWeight
type OviDeviceRates []OciDeviceRate

func (w *OciDeviceWeights) append(maj, min, val int64) {
    *w = append(*w, OciDeviceWeight{Major: maj, Minor: min, Weight: val})
}

func (r *OciDeviceRates) append(maj, min, val int64) {
    *r = append(*r, OciDeviceRate{Major: maj, Minor: min, Rate: val})
}

func readOciDeviceParameters(baseDir string, filenames []string, params OciDeviceParameters) error {
    var errors *multierror.Error
    ... // loop of reading and parsing the files
        params.append(major, minor, value)
    ...
    return errors.ErrorOrNil()
}
```

marquiz: 👍exactly, more thought out naming (and expression) than what I had

klihub: @askervin Are there any plans to address this last review comment ? Or would you prefer not to ?

askervin: @klihub , @marquiz , I think this was very good improvement to naming and structuring the code. I walked a bit further down the path in order to clean up `updateOrAppendDevice{Weight,Rate}` functions similarly. So the interface above got the form
```go
type OciDeviceParameters interface {
	Append(maj, min, val int64)
	Update(maj, min, val int64)
}
```

The patch is now updated and tested (old and new unit tests + system test with SIGHUP-config update).

sample-configs/blockio.cfg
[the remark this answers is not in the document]
kad: should it be debug only for "blockio" loggers?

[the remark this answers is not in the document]
askervin: Yes, that's better. Fixed.

pkg/cgroups/cgroupblkio.go
```diff
@@ -50,51 +52,99 @@ var blkioThrottleWriteIOPSFiles = []string{"blkio.throttle.write_iops_device"}
...
+func (r *OciDeviceRates) Update(maj, min, val int64) {
+	if r != nil {
+		for index, devRate := range *r {
+			if devRate.Major == maj && devRate.Minor == min {
+				(*r)[index].Rate = val
+				return
+			}
+		}
+	}
+	r.Append(maj, min, val)
```
marquiz, line 126: Another segfault

pkg/cgroups/cgroupblkio.go
```diff
@@ -50,51 +52,99 @@ var blkioThrottleWriteIOPSFiles = []string{"blkio.throttle.write_iops_device"}
...
+func (w *OciDeviceWeights) Update(maj, min, val int64) {
+	if w != nil {
+		for index, devWeight := range *w {
+			if devWeight.Major == maj && devWeight.Minor == min {
+				(*w)[index].Weight = val
+				return
+			}
+		}
+	}
+	w.Append(maj, min, val)
```
marquiz, line 113: This will segfault if w is nil.

Do we even want to check for `w != nil`? And thus, silently ignore some possible bugs.

askervin: Good finding, thanks!

I removed the unnecessary `w != nil` check. Now if `.Append` or `.Update` are called with `nil` slice pointer, it'll cause a runtime error in the `for ... range *list` loop.

pkg/cgroups/cgroupblkio.go
askervin: Fixed like the next point.

PR #236, opened 2020-04-14
config: always call registration-time notifiers first.

klihub:
Make sure that 'owning' configuration notification callbacks (attached during module registration as a `config.WithNotify()` option passed to `config.Register()`), are always invoked *before* 'foreign' notification callbacks (attached using `config.AddNotify()`). This establishes a notification invocation order that is defined well enough to be relied on.

Without this patch, notification callbacks were internally stored and therefore invoked in the temporal order of registration. If a foreign callback was attached to a yet to be created module, then eventually the module's own configuration callback(s) ended up registered and called after any such foreign one.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -122,9 +122,12 @@ func Register(path, description string, ptr interface{}, getfn GetConfigFn, opts
 
 	m.check()
 
+	foreign := m.notifiers
+	m.notifiers = nil
 	for _, opt := range opts {
 		opt.apply(m)
 	}
+	m.notifiers = append(m.notifiers, foreign...)
 
 	return m
 }


ipuustin:
Another way to do this might be to use a priority queue (https://golang.org/pkg/container/heap/#example__priorityQueue) to hold the notifiers. This would have the benefit of keeping the list automatically in order (to avoid bugs) and having support of adding various notifiers dynamically later. The drawback would be increased code and complexity. What do you think?

klihub:
> Another way to do this might be to use a priority queue (https://golang.org/pkg/container/heap/#example__priorityQueue) to hold the notifiers. This would have the benefit of keeping the list automatically in order (to avoid bugs) and having support of adding various notifiers dynamically later. The drawback would be increased code and complexity. What do you think?

We do have support for attaching config update notification callbacks after (and even before) a 'config module' has been created. It's just that the order of invocation between the 'owner's' notifications (attached at registration time) vs. 'foreign' notifications (attached before or after the module was registered) was not well defined. This PR addresses that.

At this point I would not go further than this. We do have the means to force notification/configuration order in both ways (once this patch is in place): before and after a module and we don't seem to need more at this point.

ipuustin:
I'm fine with this approach, too.

PR #239, opened 2020-04-15
fix: cache: don't return duplicates by pod.Get[Init]Containers().

klihub:
Make sure the slices returned by pod.GetContainers() and pod.GetInitContainers() contain every container only once.


askervin:
LGTM.

Tested that the fix really works with loops:
```go
        for _, pod := range cache.GetPods() {
                for _, cont := range pod.GetContainers() {
                        log.Info("container %q", cont.PrettyName())
                }
        }
```

Without this fix `pod.GetContainers()` return duplicate containers in controller `Start(cache, client)`.

PR #249, opened 2020-04-23
cpuallocator: don't crash/try ever to index into an empty slice.

klihub:
Without this patch, it is possible to trick the CPU allocator to crash, with enough/the right CPUs offline, as it ends up trying to index the first element of an empty slice when looking for idle cores.


klihub:
> Uh-oh, was it me(?) 
> 
> lgtm

No, I think that was me. I have a faint memory of that 'I really shouldn't be doing this' ting about indexing that returned slice without checking its size. Yup, git blame agrees. Shame on me...

PR #250, opened 2020-04-23
memtier: fix memory type checking, a few inherited scoring bugs. 

klihub:
This patch series
  - fixes memory type subset check to use proper sub-masking instead of comparison
  - fixes a few scoring bugs inherited from the original topology-aware policy code
  - adds detailed logging of the score sorting/comparison algorithm
  - adds logging of the pod's/container's memory type preferences

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/memtier/pools.go
+++ b/pkg/cri/resource-manager/policy/builtin/memtier/pools.go
@@ -601,11 +601,11 @@ func (p *policy) compareScores(request Request, scores map[int]Score,
 	}
 
 	// 3) matching memory type wins
-	if request.MemoryType() != memoryUnspec {
-		// see if the nodes have different memory types and whether they match the request
-		if node1.GetMemoryType() == request.MemoryType() && node2.GetMemoryType() != request.MemoryType() {
+	if reqType := request.MemoryType(); reqType != memoryUnspec {
+		if node1.HasMemoryType(reqType) && !node2.HasMemoryType(reqType) {
 			return true
-		} else if node1.GetMemoryType() != request.MemoryType() && node2.GetMemoryType() == request.MemoryType() {
+		}
+		if !node1.HasMemoryType(reqType) && node2.HasMemoryType(reqType) {
 			return false
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -586,28 +590,40 @@ func (p *policy) compareScores(request Request, scores map[int]Score,
 
 	// 1) a node with insufficient isolated or shared capacity loses
 	switch {
-	case isolated2 < 0 || shared2 < 0:
```
ipuustin, line 589: This is the same pattern as below with the memory (see if node1 is "good", and then if node2 isn't good, node1 wins. If not, check for the opposite case). While this is right, we should maybe think how this could be done in a more readable and easy-to-validate way for all the cases we have to check.

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -586,28 +590,40 @@ func (p *policy) compareScores(request Request, scores map[int]Score,
...
 	}
 
+	log.Debug("  - affinity is a TIE")
+
 	// 3) matching memory type wins
-	if request.MemoryType() != memoryUnspec {
-		// see if the nodes have different memory types and whether they match the request
-		if node1.GetMemoryType() == request.MemoryType() && node2.GetMemoryType() != request.MemoryType() {
+	if reqType := request.MemoryType(); reqType != memoryUnspec {
+		if node1.HasMemoryType(reqType) && !node2.HasMemoryType(reqType) {
```
ipuustin, line 617: Yes, it is better this way.

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
klihub: Hmm... I think it's not the same. The real problem with the above code snippet was that it never considered the case that both nodes fail to fullfill the preference, so it failed to handle that correctly (always blindly chose the first node). On a node without any isolated CPUs this basically screwed the whole pool sorting algorithm for all workloads eligible or annotated to prefer isolated CPUs, by practically always chosing the first node in every comparison.

PR #253, opened 2020-04-29
fixes/cache: add test cases for making sure CPU estimates are properly rounded for full CPUs.

klihub:
Add test cases for making sure the estimates of CPU requests/limits worth of full CPUs are accurate. Such containers are treated specially by stock kubernetes/kubelet as the 'static' policy considers Guaranteed QoS class containers with such requests eligible for exclusive CPU allocation.

Also, since the relevant related constants and functions are now exported from kubelet, pull in those ones and use them instead of duplication, whenever this is possible.

### DIFF
--- a/pkg/cri/resource-manager/cache/utils.go
+++ b/pkg/cri/resource-manager/cache/utils.go
@@ -23,16 +23,7 @@ import (
 	resapi "k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
 	cri "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
-)
-
-// Constants/variables needed for converting between milliCPU, CFS shares, quota and period.
-const (
-	// CFS CPU shares, quota and period to/from milliCPU conversion
-	minShares      = 2
-	sharesPerCPU   = 1024
-	milliCPUToCPU  = 1000
-	QuotaPeriod    = 100000
-	minQuotaPeriod = 1000
+	kubecm "k8s.io/kubernetes/pkg/kubelet/cm"
 )
 
 var memoryCapacity int64
@@ -92,11 +83,10 @@ func QuotaToMilliCPU(quota, period int64) int64 {
 
 // sharesToMilliCPU converts CFS CPU shares to milliCPU.
 func sharesToMilliCPU(shares int64) int64 {
-	if shares == minShares {
+	if shares == kubecm.MinShares {
 		return 0
 	}
-	//return int64(float64(shares*milliCPUToCPU) / float64(sharesPerCPU) + 0.5)
-	return int64((float64(shares*milliCPUToCPU) / float64(sharesPerCPU)) + 0.5)
+	return int64(float64(shares*kubecm.MilliCPUToCPU)/float64(kubecm.SharesPerCPU) + 0.5)
 }
 
 // quotaToMilliCPU converts CFS quota and period to milliCPU.
@@ -104,21 +94,12 @@ func quotaToMilliCPU(quota, period int64) int64 {
 	if quota == 0 || period == 0 {
 		return 0
 	}
-	return int64(float64(quota*milliCPUToCPU)/float64(period) + 0.5)
+	return int64(float64(quota*kubecm.MilliCPUToCPU)/float64(period) + 0.5)
 }
 
 // MilliCPUToShares converts milliCPU to CFS CPU shares.
 func MilliCPUToShares(milliCPU int) int64 {
-	if milliCPU == 0 {
-		return minShares
-	}
-
-	shares := (milliCPU * sharesPerCPU) / milliCPUToCPU
-	if shares < minShares {
-		return minShares
-	}
-
-	return int64(shares)
+	return int64(kubecm.MilliCPUToShares(int64(milliCPU)))
 }
 
 // MilliCPUToQuota converts milliCPU to CFS quota and period values.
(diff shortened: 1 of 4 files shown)


ipuustin:
Maybe we should always only round small numbers up, in case someone tries to make a burstable container by giving it a request of 1 mCPU? Also, why don't we make 1000, 2000, etc. "snappy" so that everything in (for example) range 901 - 1099 would become 1000 to make sure we don't mess guaranteed containers?

grahamwhaley:
A couple of thoughts...
- never give folks less than they asked for :-) Well, OK, either way, if you don't give them what they asked for, expect them to come and ask why they got what they did (they do go look sometimes...)... so, to alleviate that....
- I'd strongly recommend you clearly document the algorithm in the docs of the repo, to reduce the number of direct questions.

klihub:
> A couple of thoughts...
> 
> * never give folks less than they asked for :-) Well, OK, either way, if you don't give them what they asked for, expect them to come and ask why they got what they did (they do go look sometimes...)... so, to alleviate that....

I agree. The problem here is we don't know what they asked for.

> * I'd strongly recommend you clearly document the algorithm in the docs of the repo, to reduce the number of direct questions.

Agree and will do so. And the most important point in that will be : "If you want to be absolutely sure that CRI Resource Manager has an exact understanding of the resource requests and limits of your pods, use the annotating webhook designed for that very purpose."

klihub:
> Also, why don't we make 1000, 2000, etc. "snappy" so that everything in (for example) range 901 - 1099 would become 1000 to make sure we don't mess guaranteed containers?

The reverse calculation is not that inaccurate for CPU. For requests, the inaccuracy is below 2 milli-CPUs and limit reverse-calculation is accurate above a 10 milliCPU cutoff.

Hmm... actually now that you asked, even the current rounding is too aggressive. I'll change it to be in line with the achievable accuracy and update the tests.

pkg/cri/resource-manager/cache/utils.go
```diff
@@ -82,13 +86,46 @@ func estimateComputeResources(lnx *cri.LinuxContainerResources, cgroupParent str
...
+}
+
+// roundMilliCPU rounds values within milliCPUThreshold to milliCPUResolution.
+func roundMilliCPU(value int64) int64 {
+	diff := value % milliCPUResolution
+	if diff == 0 {
+		return value
+	}
+
+	if diff != 0 && diff < milliCPUThreshold {
```
ipuustin, line 120: No need to check if `diff != 0` since it was made certain in the above `if` statement.

pkg/cri/resource-manager/cache/utils.go
```diff
@@ -82,13 +86,46 @@ func estimateComputeResources(lnx *cri.LinuxContainerResources, cgroupParent str
...
+		return value - diff
```
ipuustin, line 121: Hmm, so this means that we round down if `diff == 1`? Does this cause trouble if `value == 1`?

pkg/cri/resource-manager/cache/utils.go
```diff
@@ -82,13 +86,46 @@ func estimateComputeResources(lnx *cri.LinuxContainerResources, cgroupParent str
...
+	}
+	if diff > (milliCPUResolution - milliCPUThreshold) {
```
ipuustin, line 123: And here, if `diff > 48`, we round by adding `50 - diff` to the value. Since `diff` can only be `49`, this means that we add one to the value. You should add comments explaining the round-up and round-down cases so that the reader doesn't need to run the code in their head. :-)

pkg/cri/resource-manager/cache/utils.go
klihub: Decided to remove the rounding code altogether for now. Instead, we now simply check that for full CPU's our estimates are accurate and otherwise the estimates are within our expected accuracy. Also, we switched to using the relevant constants (and also functions whenever possible) directly from kubelet as those are now exported.

pkg/cri/resource-manager/cache/utils.go
```diff
@@ -127,10 +108,10 @@ func MilliCPUToQuota(milliCPU int64) (int64, int64) {
 		return 0, 0
 	}
 
-	period := int64(QuotaPeriod)
-	quota := (milliCPU * period) / milliCPUToCPU
-	if quota < minQuotaPeriod {
-		quota = minQuotaPeriod
+	period := int64(kubecm.QuotaPeriod)
```
ipuustin, line 111: This may be a silly question, but is this a constant or does it change with Kubelet's `CPUCFSQuotaPeriod` setting?

pkg/cri/resource-manager/cache/utils.go
```diff
@@ -92,33 +83,23 @@ func QuotaToMilliCPU(quota, period int64) int64 {
...
+	return int64(float64(shares*kubecm.MilliCPUToCPU)/float64(kubecm.SharesPerCPU) + 0.5)
 }
 
 // quotaToMilliCPU converts CFS quota and period to milliCPU.
 func quotaToMilliCPU(quota, period int64) int64 {
 	if quota == 0 || period == 0 {
 		return 0
 	}
-	return int64(float64(quota*milliCPUToCPU)/float64(period) + 0.5)
+	return int64(float64(quota*kubecm.MilliCPUToCPU)/float64(period) + 0.5)
```
ipuustin, line 97: Is `+ 0.5` meant to work as `math.Ceil()` replacement?

klihub: No, `math.Ceil()` always rounds up, IOW results in the smallest integer larger than or equal to the input. This rounds up or down to the nearest integer instead.

ipuustin: Yes, sure, but I'm just concerned about the `0` case. Since we know here that `quota` and `period` are bigger than zero, should we aim to have the milliCPUs to be at least `1`? Other than that the difference of 1 milliCPU is not much, if we are not using this function to handle the "integer CPU" case of figuring out whether or not `milliCPUs % 1000 == 0`.

pkg/cri/resource-manager/cache/utils.go
klihub: Not silly, it's a very valid question. The answer is that this is a constant but it is only used if the CPUCFSQuotaPeriod feature gate is disabled (the default). Otherwise the period is taken from the configured CPUCFSQuotaPeriod value.

This code assumed we're running with the default settings. I haven't taken a look at whether and how to do the reverse calculation in the other case. I can give it a try.

PR #254, opened 2020-04-30
webhook: fix segfault in handler

marquiz:
Do not attempt to UID from request if its deserialization failed.


marquiz:
Fixed a typo in commit message and rebased.

askervin:
Suggesting reordering words in the comment line:
```
// Use same the UID in response that ...
->
// Use the same UID in response that ...
```

Otherwise LGTM.

marquiz:
> Suggesting reordering words in the comment line

Haha, I was copy-pasteing bad wording 😄 Thanks for spotting this, fixed now.

PR #263, opened 2020-05-05
 webhook: run as non-root user inside container image

marquiz:


marquiz:
Please hold, not fully tested yet

marquiz:
OK, now tested, and, fixed

cmd/cri-resmgr-agent/Dockerfile
```diff
@@ -35,6 +35,7 @@ COPY . .
 RUN make BUILD_DIRS=cri-resmgr-agent
 
 FROM scratch as final
+USER 1000:1000
```
klihub, line 38: Does this really work ? I mean after this change the agent can still successfully create its socket in /var/run/cri-resmgr. How about any user 1000 on the host (the first user created with default id, at least on Fedora, CentOS, etc.)... it has no way of easily impersonating the agent by creating a socket with the same path ?

marquiz: > Does this really work ? I mean can the agent after this change still successfully create its socket in /var/run/cri-resmgr ? How about any user 1000 on the host (the first user created with default id, at least on Fedora, CentOS, etc.)... it has no way of easily impersonating the agent by creating a socket with the same path ?

Argh, I must have actually tested a *wrong* image from a patchset building on top of this one. You're right, `1000:1000` doesn't work (at it's most probably not able to create the socket file) and is a bad idea because of the user mapping. I changed this to `1:0` but that is also probably worse than just running as root.

I'll drop this patch for now

ipuustin: If we want, we could just use a username (`USER cri-resmgr` or similar) there and create the user on the host side?

marquiz:
New version:
- only webhook is touched
- UID:GID changed to `65534:65534`

cmd/cri-resmgr-agent/Dockerfile
klihub: Thanks, I also think dropping this for now is better.

If we want to walk down this path, we can take a look at it separately later. Then I would probably take an approach where the agent has a well-known dedicated user which gets created when cri-resmgr is installed. We'd also need to drop the agent's socket under a dedicated directory (maybe under /var/run/cri-resmgr/agent), which probably would be pre-created with the right ownership and permissions during boot using systemd-tmpfiles or something similar. Anyway, we can sort out the details later if/when we decide to do try moving in this direction.

klihub:
> New version:
> 
> * only webhook is touched
> * UID:GID changed to `65534:65534`

Yes, there it should be fine, together with the change away from the sub-1024 port you made.

cmd/cri-resmgr-agent/Dockerfile
marquiz: > If we want, we could just use a username (`USER cri-resmgr` or similar) there and create the user on the host side?

But that same user/uid must exist in the container, too, with matching uid, afaiu

PR #264, opened 2020-05-06
memtier: improve policy documentation.

ipuustin:


ipuustin:
@ss7pro : we can change HBMEM -> HBM pretty easily at this point, even though it requires a breaking API change. The reason for the memory request inaccuracy is the reverse calculation if the webhook doesn't set the K8s request to the annotations: currently we don't even try to estimate the memory request for burstable pods. See this function: https://github.com/intel/cri-resource-manager/blob/master/pkg/cri/resource-manager/cache/utils.go#L41

PR #266, opened 2020-05-06
fix potential nil pointer dereference in CRI controller.

askervin:
Sometimes create.Config.Linux can be nil. Revealed by critest when
running against memtier and topology-aware policies.

### DIFF
--- a/pkg/cri/resource-manager/control/cri/cri.go
+++ b/pkg/cri/resource-manager/control/cri/cri.go
@@ -86,6 +86,9 @@ func (ctl *crictl) PreCreateHook(c cache.Container) error {
 	create.Config.Envs = c.GetCRIEnvs()
 	create.Config.Mounts = c.GetCRIMounts()
 	create.Config.Devices = c.GetCRIDevices()
+	if create.Config.Linux == nil {
+		create.Config.Linux = &criapi.LinuxContainerConfig{}
+	}
 	create.Config.Linux.Resources = c.GetLinuxResources()
 
 	c.ClearPending(CRIController)


askervin:
After this change [`critest`](https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/validation.md) gives the same pass/fail ratio to memtier and topology-aware policies as the null policy and running the test directly against `containerd` in my setup (70 pass, 2 fail, 7 skipped).

critests against cri-resmgr are [under development in `test/critest`](https://github.com/askervin/cri-resource-manager/tree/56S_critest/test/critest).

pkg/cri/resource-manager/control/cri/cri.go
```diff
@@ -86,7 +86,9 @@ func (ctl *crictl) PreCreateHook(c cache.Container) error {
 	create.Config.Envs = c.GetCRIEnvs()
 	create.Config.Mounts = c.GetCRIMounts()
 	create.Config.Devices = c.GetCRIDevices()
-	create.Config.Linux.Resources = c.GetLinuxResources()
+	if create.Config.Linux != nil {
+		create.Config.Linux.Resources = c.GetLinuxResources()
+	}
```
klihub, line 91: Hmm.... although this should never happen if the request comes from kubelet, I still think the proper fix would be to do something like this:

```
if create.Config.Linux == nil {
    create.Config.Linux = &cri.LinuxContainerConfig{}
} 
create.Config.Linux.Resources = c.GetLinuxResources()
```

Otherwise you could in principle avoid the (CRI parts of the) policies getting enforced on your container by leaving Config.Linux deliberately nil in the create request, just like above.

pkg/cri/resource-manager/control/cri/cri.go
```diff
@@ -86,7 +86,9 @@ func (ctl *crictl) PreCreateHook(c cache.Container) error {
 	create.Config.Envs = c.GetCRIEnvs()
 	create.Config.Mounts = c.GetCRIMounts()
 	create.Config.Devices = c.GetCRIDevices()
-	create.Config.Linux.Resources = c.GetLinuxResources()
+	if create.Config.Linux != nil {
+		create.Config.Linux.Resources = c.GetLinuxResources()
```
ipuustin, line 90: While at it, it might be a good idea to check if there are other similar cases in our codebase (we get a complex object, which might be only partly initialized, back from a kubernetes API).

pkg/cri/resource-manager/control/cri/cri.go
askervin: Ok, fixed.

PR #270, opened 2020-05-07
webhook: check for empty requests

marquiz:
Fixes a possible segfault.

### DIFF
--- a/cmd/cri-resmgr-webhook/handlers.go
+++ b/cmd/cri-resmgr-webhook/handlers.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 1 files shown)


cmd/cri-resmgr-webhook/handlers.go
```diff
@@ -93,6 +93,9 @@ func handle(w http.ResponseWriter, r *http.Request) {
 	if _, _, err := deserializer.Decode(body, nil, &arReq); err != nil {
 		log.Printf("ERROR: deserializing admission request failed: %v", err)
 		arRsp.Response = errResponse(err)
+	} else if arReq.Request == nil {
+		log.Printf("REQUEST empty")
+		arRsp.Response = errResponse(fmt.Errorf("Empty request"))
```
klihub, line 98: minor nitpick: since the error is not really formatted, you could use errors.New("Empty request") here. But I'm very often guilty of the same myself.

marquiz: Changed.

This just happens so easily when you copy/paste snippets around and mangle the arguments and format strings.

PR #271, opened 2020-05-07
webhook: add liveness probe to the sample deployment spec

marquiz:


marquiz:
> Why is `containerPort: 443` ?

I had just missed that. I don't know why that had been specified in the first place. Submitted #273 to drop that.

> Also, if the liveliness probe goes to `8443` with a `HTTP GET` and `ContentType JSON`, isn't that ending up in the same `handlers.go:handle()` as the annotation requests ? And if it does end up there, how come it succeeds ? To me it looks like the handler would return an `"Empty request"` error response, because I guess the liveliness probe cannot be deserialized as a valid annotation request.

Deserialization succeeds because the helper libraries automatically "fill in" the api version and kind, based on what was registered there. It sends an "admission error" but that's not an HTTP error so it works.

 

> Or is it so, the handler responses are extra headers and even an error response returns with a `200 OK`, while other HTTP-level responses are reserved for transport-level errors ?

Basically, yes

> If it is so, then I see how this works but can I ask, just out of curiosity, why did you chose this approach instead of more explicit support for health check,

Probably because this was the quickest way 😄  And, it really checks if the admission handler responds.

marquiz:
...I can add something like you suggested, though

klihub:
> ...I can add something like you suggested, though

I guess there is no need for that as long as the webhook stays so simple that this 'it replies therefore it is healthy' check is reliable.

klihub:
> 
> Probably because this was the quickest way  And, it really checks if the admission handler responds.

Hmm... yes, handling the probes in the real admission request handler is a good point. And it'd be easy to tell probes and real admission requests apart, if necessary, by using something else than `ContentType: application/json` for the probes. But I think this fine just the way it is now.

PR #272, opened 2020-05-07
agent: add simple liveness probe to the sample deployment

marquiz:
Simple nc to check if the agent socket accepts connections.


marquiz:
Wise words. This PR is really a very simplistic health check. I had similar thoughts than your reasoning above. However, I realized that designing and writing a more comprehensive health check takes more time than what I have, and wanted to have something, for now. Anyway, we'd like to somehow not only verify the health of the gRPC server, but, the state of the config watcher and config updater, too. I created a new issue (#274) for tracking a better solution.

PR #275, opened 2020-05-08
agent: use our probe client in the liveness probe

marquiz:
Forget about nc. Makes the liveness probe a lot more reasonable, and, drops container image size to one fifth.


cmd/cri-resmgr-agent-probe/main.go
```diff
@@ -0,0 +1,59 @@
...
+		log.Fatal("failed to connect to agent: %v", err)
+	}
+	cli := agent_v1.NewAgentClient(conn)
+
+	// Do health check
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	rpl, err := cli.HealthCheck(ctx, &agent_v1.HealthCheckRequest{})
+	if err != nil {
```
klihub, line 55: So just that I understand/we explicitly spell out the conventions here:
  - a `nil` error implies an `OK` `Status` in the reply (so it is safe to leave that unchecked)
  - a non-`nil` error with a non-`nil` `Status` implies an agent-level error (possibly with further information in `Status`  about what is wrong ?)
- a non-`nil` error with a `nil` reply implies a transport/communication level error

marquiz: Good questions @klihub. The PR was quickly patched together and the status field could also be dropped. 

The status field in the reply is just for informative purposes, that could be used to carry more detailed inforation about the state. Thus, the logic would be just like you described (we just don't really use the field for anything, yet). A non-healthy state always causes an error response. In practice, a non-nil rsp error with nil status would imply a "lower-level" error as you described, and similarly, a non-nil rsp error with a non-empty status would imply that the agent was able to respond but it is not functioning properly (status would carry more detailed info about the problems). And, a nil rsp error implies an healthy state, but, even in this case we could carry some information in the status field.

marquiz:
Woops 🙄 I had a wrong understanding of gRPC error handling. You cannot carry status information in an error response. If you return an error the rsp message will be nil.

I now changed the reply message `status` field to `error`. Healthy responses carry no status information. Unhealthy reponse is an OK on gRPC level but the `error` field in the reply message is non-empty. gRPC errors indicate a lower level (communication) failure.

PR #280, opened 2020-05-12
relay: copy permissions/ownership of client socket to server.

klihub:


pkg/cri/server/server.go
```diff
@@ -199,6 +205,39 @@ func (s *server) createGrpcServer() error {
...
+// Chown changes ownership of the server's socket.
+func (s *server) Chown(uid, gid int) error {
+	usr := strconv.FormatInt(int64(uid), 10)
+	if u, err := user.LookupId(usr); u != nil && err == nil {
+		usr = u.Name
+	}
+	grp := strconv.FormatInt(int64(gid), 10)
+	if g, err := user.LookupGroupId(grp); g != nil && err == nil {
+		grp = g.Name
+	}
```
marquiz, line 229: nit: These could be moved down to the else branch where they are actually used. Would make the code easier to read. Prolly it's just me plus the late evening but it took a while for me to grasp that these are actually only for human-readability

pkg/cri/client/client.go
```diff
@@ -202,6 +214,27 @@ func (c *client) connect(kind, socket string, options ConnectOptions) (*grpc.Cli
...
 }
 
+func (c *client) dialNotify(socket string) {
+	if c.options.DialNotify == nil {
+		return
+	}
+
+	info, err := os.Stat(socket)
+	if err != nil {
+		c.options.DialNotify(socket, 0, 0, 0, err)
```
ipuustin, line 224: This works and is fine, but it just feels pretty brave to set the UID and GID to "root" when we call the function with an error value :-)

pkg/cri/relay/relay.go
```diff
@@ -134,6 +150,34 @@ func (r *relay) Server() server.Server {
...
 }
 
+func (r *relay) dialNotify(socket string, perm os.FileMode, uid int, gid int, err error) {
+	if err != nil {
+		r.Error("failed to determine permissions/ownership of client socket %q: %v",
+			socket, err)
+		return
+	}
+
+	// we ignore image socket properties, unless it is the only socket we have
```
ipuustin, line 160: Maybe explain here why this design decision?

pkg/cri/relay/relay.go
```diff
@@ -134,6 +150,34 @@ func (r *relay) Server() server.Server {
...
+		if socket != r.options.RuntimeSocket && r.options.RuntimeSocket != client.DontConnect {
+			return
+		}
+	}
+
+	r.perm = perm
+	r.uid = uid
+	r.gid = gid
+
+	if r.started {
```
ipuustin, line 171: So if the relay isn't started, we don't `chown`? Shouldn't we try to `chown` precisely when the relay is not running to avoid any race?

pkg/cri/relay/relay.go
```diff
@@ -134,6 +150,34 @@ func (r *relay) Server() server.Server {
...
+		if r.perm != 0 {
+			r.server.Chmod(r.perm)
+		}
+		if r.uid >= 0 {
```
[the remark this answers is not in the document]
klihub: But we do try then. That r.uid >=0 was supposed to make sure that if stat():ing the socket fails we don't event try to set uid/gid to -1/-1, which is what we internally used to pass in the case os.Stat() succeeded but sys.Stat() failed. But looking at the client code, I realize that I changed that in the last minute and we pass in an 0's and an error in that case. So we don't need those safeguard checks any more. I'll update.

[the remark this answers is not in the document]
marquiz: Ha, my bad, I was obviously too tired when reviewing the code. But good if you found something to fix, nevertheless 😊

pkg/cri/relay/relay.go
klihub: If you use the stock golang net.Listener/net.Dial() family of functions to create a unix-domain listening socket and then you change its ownership/permissions, then you simply cannot avoid the race condition. While the unix system call interface properly separates socket creation, binding and listening, the golang net abstraction hides creating the socket and listening on it behind a single function call.

pkg/cri/relay/relay.go
klihub: Added a comment explaining what's going on and wy.

pkg/cri/client/client.go
klihub: True, dangerous to trust the other end to not be brain-damaged. Changed those to -1's. Also changed the order of arguments to be more logical/natural.

pkg/cri/server/server.go
klihub: It's also probably partly the weird golang API: you have to pass integer arguments as strings. Anyway, the resolved usr/grp was supposed to be used in both the success and the failure branches, that's why the lookups are in the beginning. Updated the code accordingly.

pkg/cri/relay/relay.go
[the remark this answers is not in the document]
klihub: Refactored the code to have a(n IMO) more reasonable split of duties between relay, client, and the server. That did away completely with that instance of check, and having to store the properties in the relay as well.

PR #281, opened 2020-05-13
cache: create directories/files with stricter permissions, check existing permissions.

klihub:
Spell out more explicitly the expected/preferred vs. rejected permission bits for the various cache/container data files and directories.Check that files and directories have root/root as owner/group. Violation of ownership or permission check of existing files/directories results in an error/rejection (and the failure of the related operation). Non-existent files/directories are created with the expected/preferred permissions (without an extra explicit check to see if we run as root/root).

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -16,6 +16,7 @@ package cache
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -492,6 +493,20 @@ const (
 	CacheVersion = "1"
 )
 
+// permissions describe preferred/expected ownership and permissions for a file or directory.
+type permissions struct {
+	prefer os.FileMode // permissions to create file/directory with
+	reject os.FileMode // bits that cause rejection to use an existing entry
+}
+
+// permissions to create with/check against
+var (
+	cacheDirPerm  = &permissions{prefer: 0710, reject: 0022}
+	cacheFilePerm = &permissions{prefer: 0644, reject: 0022}
+	dataDirPerm   = &permissions{prefer: 0755, reject: 0022}
+	dataFilePerm  = &permissions{prefer: 0644, reject: 0022}
+)
+
 // Our cache of objects.
 type cache struct {
 	sync.Mutex    `json:"-"` // we're lockable
@@ -536,11 +551,15 @@ func NewCache(options Options) (Cache, error) {
 		implicit:   make(map[string]*ImplicitAffinity),
 	}
 
-	if err := os.MkdirAll(options.CacheDir, 0700); err != nil {
-		return nil, cacheError("failed to create cache directory %s: %v",
-			options.CacheDir, err)
+	if _, err := cch.checkPerm("cache", cch.filePath, false, cacheFilePerm); err != nil {
+		return nil, cacheError("refusing to use existing cache file: %v", err)
 	}
-
+	if err := cch.mkdirAll("cache", options.CacheDir, cacheDirPerm); err != nil {
+		return nil, err
+	}
+	if err := cch.mkdirAll("container", cch.dataDir, dataDirPerm); err != nil {
+		return nil, err
+	}
 	if err := cch.Load(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/cache/cache.go
```diff
@@ -490,6 +490,12 @@ type Cache interface {
 const (
 	// CacheVersion is the running version of the cache.
 	CacheVersion = "1"
+	// cacheFilePermissions is the unix permissions for the cache file
+	cacheFilePermissions = os.FileMode(0640)
+	// cacheDirPermissions is the unix permissions for the directory of the cache file
+	cacheDirPermissions = os.FileMode(0710)
```
ipuustin, line 496: Here the original permission was stricter... but maybe the group access is a good thing if we move later running the service as a user.

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1110,6 +1119,37 @@ func (cch *cache) setEntry(key string, ptr, obj interface{}) error {
 	return nil
 }
 
+// checkFilePermissions check and warns if an existing file has unexpected permissions.
+func (cch *cache) checkFilePermissions(path string, perm os.FileMode, what string) {
+	if info, err := os.Stat(path); err == nil && !info.IsDir() {
+		expected := perm & os.ModePerm // deliberately ignore sticky bit in comparison
+		existing := info.Mode().Perm() & os.ModePerm
+		if existing != expected && (existing|expected) != expected {
+			cch.Warn("existing %s file %q has less strict permissions %v than expected %v",
```
ipuustin, line 1128: Why not error out?

pkg/cri/resource-manager/cache/cache.go
klihub: Yes, that's deliberate. Maybe the commit message is a bit misleading.

pkg/cri/resource-manager/cache/cache.go
klihub: With a pre-created directory, the more liberal permissions might be a deliberate choice and we should be very careful when/if we try to outsmart the sysadmin. This is what we agreed in the last Fri session, but we should probably still error out in some extreme cases, for instance if a file or directory is unexpectedly world-readable. I see if I can update this to do that in some acceptable form...

klihub: Added checking of permissions bits to reject.

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1110,6 +1128,47 @@ func (cch *cache) setEntry(key string, ptr, obj interface{}) error {
 	return nil
 }
 
+// checkFilePermissions checks permission bits of an existing file.
+func (cch *cache) checkFilePermissions(path string, perm, rejected os.FileMode, what string) error {
+	if info, err := os.Stat(path); err == nil && !info.IsDir() {
+		expected := perm & os.ModePerm // deliberately ignore sticky bit in comparison
+		existing := info.Mode().Perm() & os.ModePerm
+		if rejected != 0 && (existing&rejected) != 0 {
```
marquiz, line 1136: nit: the first check is not needed, right(?) If rejected==0 then existing&rejected==0

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1110,6 +1128,47 @@ func (cch *cache) setEntry(key string, ptr, obj interface{}) error {
...
+			return cacheError("existing %s file %q has disallowed permissions set: %v",
+				what, path, existing&rejected)
+		}
+		if existing != expected && (existing|expected) != expected {
```
marquiz, line 1140: nit: I think this logic is also somehow overcomplicated. Wouldn't `(existing|expected) != expected` be enough, because if that is true then certainly `existing != expected`?

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1110,6 +1128,47 @@ func (cch *cache) setEntry(key string, ptr, obj interface{}) error {
...
+		expected := perm & os.ModePerm // deliberately ignore sticky bit in comparison
+		existing := info.Mode().Perm() & os.ModePerm
+		if rejected != 0 && (existing&rejected) != 0 {
+			return cacheError("existing %s directory %q has disallowed permissions set: %v",
+				what, path, existing&rejected)
+		}
+		if existing != expected && (existing|expected) != expected {
+			cch.Warn("existing %s directory %q has less strict permissions %v than expected %v",
+				what, path, existing, expected)
+		}
```
marquiz, line 1161: Is this duplicated from `checkFilePermissions()` if so, could it be split out in a separate helper function?

pkg/cri/resource-manager/cache/cache.go
klihub: Split out permissions check, added missing ownership tests, and removed the extra zero-test.

pkg/cri/resource-manager/cache/cache.go
klihub: Removed extra-articulated inequality check.

pkg/cri/resource-manager/cache/cache.go
klihub: Split out/combined permission checks.

pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1110,6 +1131,84 @@ func (cch *cache) setEntry(key string, ptr, obj interface{}) error {
...
+	}
+
+	// check expected root/root ownership
+	stat, ok := info.Sys().(*syscall.Stat_t)
+	if !ok || stat == nil {
+		return true, cacheError("failed to get ownership of %s %q: can't get syscall.Stat_t",
+			what, path)
+	}
+	if stat.Uid != 0 || stat.Gid != 0 {
```
marquiz, line 1165: Do we want to strictly enforce this? I think it is possible to run the container runtime as rootless, for example. Or maybe run cri-resmgr as non-root when doing introspection only, for example. WDYT?

klihub: Hmm... I considered checking against geteuid()/getegid() then went with this. Perhaps both because I was a bit too much in SDL mode. Let's drop the user/group ownership checks altogether.

PR #283, opened 2020-05-14
cache: don't segfault if opening/creating container data file fails.

klihub:
Don't crash trying to deferred-close a file that failed to open.


pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1277,12 +1277,12 @@ func (cch *cache) OpenFile(id string, name string, perm os.FileMode) (*os.File,
...
 func (cch *cache) WriteFile(id string, name string, perm os.FileMode, data []byte) error {
 	file, err := cch.OpenFile(id, name, perm)
-	defer file.Close()
-
 	if err != nil {
 		return err
 	}
+
 	_, err = file.Write(data)
+	file.Close()
```
kad, line 1285: em. Go standard paradigm is to have `open`, `if err != nil`, then `defer file.Close()` and then any other operations.

Issue #244, opened 2020-04-20
Memory Tiering: Dynamic "push-out" of DRAM pages to PMEM in userspace


msivosuo:
Dynamic "push-out" of DRAM pages to PMEM in userspace

Target: 2020-05-22

ipuustin:
The algorithm would be (roughly) this:
  1. Find out which processes belong to the container.
  2. For every process in the container, find out which pages the process uses.
  3. Using `move_pages()`, push a number of pages not in the working set, which are present in DRAM, from DRAM to PMEM. This may need to be done for many times with a delay in between, because the process will be "stuck" when the pages are moved.
  4. Goto 1.

How to figure out which pages are not part of the working set:
  1. Clear soft-dirty bits on the PTEs: https://www.kernel.org/doc/html/latest/admin-guide/mm/soft-dirty.html
  2. Wait for a while.
  3. Read out the process page maps: https://www.kernel.org/doc/html/latest/admin-guide/mm/pagemap.html
  4. The pages which don't have the soft-dirty bit are considered to be outside of the working set.

Open questions:
  1. The soft-dirty mechanism helps us find the pages that are written to. How about the read-only pages?
  2. Determine the set of parameters which affect the "push-out" and find out if they should be part of the container configuration or daemon configuration. At least the following parameters can be identified: soft-dirty timer, number of pages to move in one move_pages() call, timeout between move_pages() calls.
  3. Synchronization between the page-analysis loop and the page-moving loop. Maybe there will be a "page pool" which contains the pages which should be moved in a producer-consumer fashion. It probably doesn't make sense to have addresses to all non-dirty pages in the pool at once.

kad:
@ipuustin for open questions:
 1. Read-only pages are perfect candidates to move to PMEM. they will get almost same read performance.
  1. For such parameters we can have defaults on policy level + annotations for container level. Timer between move_pages calls is practically should be same as soft-dirty timer. We find batch, move it. If we have another non-moved candidate pages, we don't do clear refs, but just re-validate them on next cycle. And amount, in theory it should be X mb/sec, similar to autonuma kernel part. On our side the only difference should be that we don't move X mb at once, but in smaller portion and measure how fast it is done. simple TBF kind of trotting in moving.
 1. well, for large datasets, the pages considered "clean" might become dirty on next check. See previous answer.

ipuustin:
Initial work here: https://github.com/ipuustin/cri-resource-manager/tree/dynamic-demotion

[...]

PR #286, opened 2020-05-18
devel: add support for policy events, add/generate container startup event.

klihub:
This patch set extends the current resource manager event functionality in preparation for the memtier policy coldstart feature. The patch set
  - adds support for handling events in policies
  - adds support for generating events in policies
  - generates an event on successful container startup

### DIFF
--- a/pkg/cri/resource-manager/events.go
+++ b/pkg/cri/resource-manager/events.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
+	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/events"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/metrics"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
 )
@@ -109,24 +110,26 @@ func (m *resmgr) SendEvent(event interface{}) error {
 func (m *resmgr) processEvent(e interface{}) {
 	evtlog.Debug("received event of type %T...", e)
 
-	m.Lock()
-	defer m.Unlock()
-
 	switch event := e.(type) {
 	case string:
 		evtlog.Debug("'%s'...", event)
-	case *metrics.Event:
+	case *events.Metrics:
 		m.processAvx(event.Avx)
+	case *events.Policy:
+		m.DeliverPolicyEvent(event)
 	default:
 		evtlog.Warn("event of unexpected type %T...", e)
 	}
 }
 
 // processAvx processes AVX512 events.
-func (m *resmgr) processAvx(e *metrics.AvxEvent) bool {
+func (m *resmgr) processAvx(e *events.Avx) bool {
 	if e == nil {
 		return false
 	}
+
+	m.Lock()
+	defer m.Unlock()
 
 	changes := false
 	for cgroup, active := range e.Updates {
--- a/pkg/cri/resource-manager/events.go
+++ b/pkg/cri/resource-manager/events.go
@@ -92,7 +92,7 @@ func (m *resmgr) stopEventProcessing() {
 	}
 }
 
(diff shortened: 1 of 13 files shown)


ipuustin:
I think this is great work, and having the event functionality will be needed also for things like https://github.com/intel/cri-resource-manager/issues/244. Some (internal) documentation about this would be good, because this will be the mechanism for handling the "upwards" communication from the policy. Especially, since `HandleEvent()` locks the global resource manager mutex, are there things which the policy can't do from the event handler?

pkg/cri/resource-manager/policy/policy.go
```diff
@@ -111,6 +119,8 @@ type Backend interface {
 	UpdateResources(cache.Container) error
 	// Rebalance tries an optimal allocation of resources for the current container.
 	Rebalance() (bool, error)
+	// HandleEvent passes on the given event to the active policy.
+	HandleEvent(*events.Policy) (bool, error)
```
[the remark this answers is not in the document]
klihub: True, we need to document it. The boolean return value indicates whether any containers have been updated/reconfigured, and consequently if controllers should be invoked to enforce the changes.

klihub:
> I think this is great work, and having the event functionality will be needed also for things like #244. Some (internal) documentation about this would be good, because this will be the mechanism for handling the "upwards" communication from the policy. Especially, since `HandleEvent()` locks the global resource manager mutex, are there things which the policy can't do from the event handler?

Well, for the reason you mention it obviously shouldn't try doing anything which would result in acquiring the resource manager lock. But unless I forgot/overlooked something this should be now fairly difficult with the current layering so I believe it to require one to really actively look for trouble.

Apart from that, depending on the event semantics (or more accurately on the context in which the event is being delivered/processed) it is not a good idea to do something terribly time consuming directly from the event handler. If you need some really heavy processing, then it is better to launch a goroutine and have that eventually deliver another event for final processing whenever it is done with its deed.

pkg/cri/resource-manager/policy/policy.go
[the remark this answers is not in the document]
klihub: Updated the doc-string(s) to describe the semantics of the boolean return value.

pkg/cri/resource-manager/events.go
```diff
@@ -109,21 +110,22 @@ func (m *resmgr) SendEvent(event interface{}) error {
...
 
-	m.Lock()
-	defer m.Unlock()
-
 	switch event := e.(type) {
 	case string:
 		evtlog.Debug("'%s'...", event)
-	case *metrics.Event:
+	case *events.Metrics:
+		m.Lock()
```
ipuustin, line 117: We should maybe have a convention about locking: should it be done here in the generic event handler or in the functions themselves (as it is done in `DeliverPolicyEvent()`).

pkg/cri/resource-manager/requests.go
```diff
@@ -481,6 +491,36 @@ func (m *resmgr) RebalanceContainers() error {
...
+
+	method := "DeliverPolicyEvent"
+	changes, err := m.policy.HandleEvent(e)
+
+	if err != nil {
+		m.Error("%s: handling event %s.%s failed: %v", method, e.Source, e.Type, err)
+		return err
+	}
+
+	if changes {
```
ipuustin, line 513: Does `changes` mean changes to the containers? Or should it be read more like `handled`?

klihub: Yes, it's 'changes to containers'. But it would probably be better to drop that boolean return altogether. The cache can already tell if there are pending changes to containers.

pkg/cri/resource-manager/events.go
klihub: Yes. It probably will need to be 'functions lock as needed'.

klihub: Updated processAvx() to lock for consistency.

PR #292, opened 2020-05-26
resource-manager/control: remove controller-level class name transformation.

klihub:
Remove support for controller-level transformation of `RDT` and `Block-I/O` class names.

Notes: This PR also eliminates the implicit treatment of `*` as the default/fallback class for both controllers.


klihub:
> In the beginning of `pkg/cri/resource-manager/control/blockio/blockio.go`:
> 
> ```go
> ConfigModuleName = "resource-manager.blockio"
> ```
> 
> We should have just `"blockio"` here now that there's no configuration that would go under `resource-manager` anymore.

I think we should just remove that const instead. We don't register any controller configuration, now that the class map is gone.

askervin:
That's right. (Didn't notice that `ConfigModuleName` was already defined and used from `pkg/blockio/blockio.go`.)

LGTM.

PR #295, opened 2020-05-28
build: explicitly serialize go tests in CircleCI.

klihub:
If go test is given a bunch of module paths it tries to test all of those in parallel using one instance per module directory. In CircleCI this seems to easily cause the exhaustion of memory and the tests to get killed by a SIGKILL.

As a workaround run tests explicitly with the golang toolchain parallel flag `-p 1` in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
       - run: make format
       - run: make
       - run: make golangci-lint
-      - run: make test
+      - run: make GO_PARALLEL="-p 1" test
       - run:
           name: Codecov report
           command: bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - run: make format
       - run: make
       - run: make golangci-lint
-      - run: make test
```
mythi, line 17: would it for for CirleCI if we allowed to pass `-p 1` to the existing test target?

klihub: Hmm, I din't realize such a flag exists. Looks to be a top-level go toolchain flag and based on the description, it probably helps. I'll give it a try.

klihub: Seems to do the trick. Thanks for pointing out, it is much less intrusive (and more general) this way.

Issue #123, opened 2020-01-14
End-to-end testing capability/infra for CRI-RM.


msivosuo:
- [ ] Integration/e2e validation proposal
- [ ] implement the proposal

marquiz:
Had short brainstorming with @klihub. We agreed that the definition of end-to-end tests comprises running the (unmodified) cri-resmgr binary, against real endpoints wherever possible. One idea was to use QEMU for running end-to-end tests which would enable easy manipulation of hw topology, for example

kad:
We need also to run official k8s compliance/node e2e subset of tests that verify that CRI behaves properly.

mythi:
There's #185 for that. I did quickly run `cri-test` some time back but the results were not consistent (the tests were failing with vanilla containerd too).

msivosuo:
Next steps - outline what needs to be implemented and find an owner

askervin:
Proposing next steps:

- Study using `govm`-based E2E test stack. `govm` runs Qemu in Docker and already used by PMEM-CSI. Qemu should enable emulating what we need, while `govm` should enable doing so without messing up developer's host with dependencies and configurations.
- Validate selected E2E stack by adding `critest` tests. Critest documentation states: _CRI validation testing provides a test framework and a suite of tests to validate that the Container Runtime Interface (CRI) server implementation meets all the requirements. This allows the CRI runtime developers to verify that their runtime conforms to CRI, without needing to set up Kubernetes components or run Kubernetes end-to-end tests._

klihub:
> Proposing next steps:
> 
> * Study using `govm`-based E2E test stack. `govm` runs Qemu in Docker and already used by PMEM-CSI. Qemu should enable emulating what we need, while `govm` should enable doing so without messing up developer's host with dependencies and configurations.

+1

klihub:
The existing/completed govm-based exercises were good steps toward this end goal. AFAICT, what we're still lacking are:
- [ ] a way to specify the details of emulated HW for govm/tests (CPU/core to NUMA node mapping)
- [ ] a well-defined abstraction for test cases:
    - [ ] definition of emulated HW to execute test case on
    - [ ] test input artifacts: probably just a bunch of Pod YAML's
    - [ ] instructions how to execute the test: maybe an ordered set of operations on pods/YAMLs
    - [ ] expected test results: after each test step how to verify correctness of the system state

askervin:
[...]

PR #308, opened 2020-06-09
tests: add critest for CRI validation, add blockio demo test mode

askervin:
- critest validates CRI without full Kubernetes stack,
  requires a container runtime with containernetworking.
- All cri-resmgr *.cfg files in tests/critest are tested
  to produce the same result as critest without cri-resmgr.
- Refactor demo/lib to reuse procedures in tests.
- Harmonize usage of test/critest and demo/blockio run.sh.


askervin:
This patch introduces e2e testing as described in issue https://github.com/intel/cri-resource-manager/issues/123.

PR #314, opened 2020-06-12
RFC: metrics: per container metrics

mythi:
RFC to how to pull in metrics in "native" format (however, I've kept AVX coming from Prometheus but moved numa stats collection hooked to `ListContainerStats`).

The metrics storage is per container flat/map structure and the sliding window buffer can be easily fed to numerical analysis, such as Gonum's stat. A flat storage is what we'd get from cAdvisor too so consuming/adding cAdvisor data should be easy.

TODO:
- "pub&sub" or policy based filtering which metrics to collect
- configurable avx thresholds
- configurable sliding window size
- ...

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	contrib.go.opencensus.io/exporter/jaeger v0.1.0
 	contrib.go.opencensus.io/exporter/prometheus v0.1.1-0.20191218042359-6151c48ac7fa
+	github.com/VividCortex/ewma v1.1.1
 	github.com/cilium/ebpf v0.0.0-20200317182658-6f632cc9ee38
 	github.com/golang/protobuf v1.3.2
 	github.com/google/go-cmp v0.3.1
@@ -20,6 +21,7 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/sys v0.0.0-20200202164722-d101bd2416d5
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
+	gonum.org/v1/gonum v0.0.0-20190331200053-3d26580ed485
 	google.golang.org/grpc v1.26.0
 	k8s.io/api v0.17.2
 	k8s.io/apimachinery v0.17.2
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -37,6 +37,8 @@ func (m *resmgr) setupRequestProcessing() error {
 		"RemoveContainer": m.RemoveContainer,
 
 		"UpdateContainerResources": m.UpdateContainer,
+
+		"ListContainerStats": m.ListContainerStats,
 	}
 
 	if err := m.relay.Server().RegisterInterceptors(interceptors); err != nil {
@@ -457,6 +459,22 @@ func (m *resmgr) UpdateContainer(ctx context.Context, method string, request int
 	m.updateIntrospection()
 
 	return &criapi.UpdateContainerResourcesResponse{}, nil
+}
+
+// ListContainerStats feeds CRI container statistics to metrics processing.
+func (m *resmgr) ListContainerStats(ctx context.Context, method string, request interface{},
+	handler server.Handler) (interface{}, error) {
+
+	reply, rqerr := handler(ctx, request)
+
+	filter := request.(*criapi.ListContainerStatsRequest).Filter
+	if filter.Id == "" && filter.PodSandboxId == "" && len(filter.LabelSelector) == 0 {
+		if err := m.SendEvent(reply.(*criapi.ListContainerStatsResponse).Stats); err != nil {
+			m.Error("failed to inject CRI ListContainerStats for event processing: %v", err)
+		}
+	}
+
+	return reply, rqerr
 }
 
 // RebalanceContainers tries to find a more optimal container resource allocation if necessary.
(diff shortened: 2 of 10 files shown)


pkg/cri/resource-manager/requests.go
```diff
@@ -461,16 +466,77 @@ func (m *resmgr) UpdateContainer(ctx context.Context, method string, request int
...
+	Memory float64 `containermetrics:"wss"`
+}
+
+type NumaMetrics struct {
+	Total       float64 `containermetrics:"numa_total"`
+	File        float64 `containermetrics:"numa_file"`
+	Anon        float64 `containermetrics:"numa_anon"`
+	Unevictable float64 `containermetrics:"numa_unevictable"`
+}
+
```
klihub, line 480: It would be nice to see a proposed 'top-level/final' per-container data structure with all data we intend to produce/collect ATM. I think that structure should
  - have the capability to include both total and per-core CPU usage (something we can get from the `cpuacct` controller)
  - have the capability to include both total and per-node NUMA stats
  - make it possible to differentiate data not being present from metrics data being 0 (struct pointers)

I also wonder if the sliding window, EWMA, etc. implementation should be visible as a mandatory interface implemented by the various structures used to compose the 'top-level' per-container data structure...

pkg/cri/resource-manager/requests.go
```diff
@@ -461,16 +466,77 @@ func (m *resmgr) UpdateContainer(ctx context.Context, method string, request int
...
+			v.Field(0).SetFloat(float64(stat.Total.Nodes[ii]))
+			v.Field(1).SetFloat(float64(stat.File.Nodes[ii]))
+			v.Field(2).SetFloat(float64(stat.Anon.Nodes[ii]))
+			v.Field(3).SetFloat(float64(stat.Unevictable.Nodes[ii]))
+
+			ns[cgroup] = v.Addr().Interface()
+		}
+		if err := m.SendEvent(&events.Metrics{Data: ns}); err != nil {
+			m.Error("failed to send NUMA Metrics for node %s: %v", ii, err)
```
klihub, line 539: Wouldn't it be more efficient, easier to read/maintain and also more flexible (read: less inherent implementation-specific restrictions) if the extraction/collection/construction of per-container metrics would be organised into a set of functions, all operating on a pointer to an instance of the 'top-level` per container metrics data structure, which would just be called one-by-one in this loop, each filling in the parts of the structure they are responsible for extracting/collecting/constructing ?

It could also get us rid of this reflect-based code using custom struct tags which tends to be quite expensive.

pkg/cri/resource-manager/requests.go
```diff
@@ -461,16 +466,77 @@ func (m *resmgr) UpdateContainer(ctx context.Context, method string, request int
...
+			v := reflect.New(reflect.StructOf(stf)).Elem()
+
+			v.Field(0).SetFloat(float64(stat.Total.Nodes[ii]))
+			v.Field(1).SetFloat(float64(stat.File.Nodes[ii]))
+			v.Field(2).SetFloat(float64(stat.Anon.Nodes[ii]))
+			v.Field(3).SetFloat(float64(stat.Unevictable.Nodes[ii]))
+
+			ns[cgroup] = v.Addr().Interface()
+		}
+		if err := m.SendEvent(&events.Metrics{Data: ns}); err != nil {
```
klihub, line 538: I would aim for a single metrics event being generated for each metrics collection round...

pkg/cri/resource-manager/requests.go
```diff
@@ -461,16 +466,77 @@ func (m *resmgr) UpdateContainer(ctx context.Context, method string, request int
...
 func (m *resmgr) ListContainerStats(ctx context.Context, method string, request interface{},
 	handler server.Handler) (interface{}, error) {
 
 	reply, rqerr := handler(ctx, request)
 
+	numaStats := make(map[string]cgroups.NumaStat)
 	filter := request.(*criapi.ListContainerStatsRequest).Filter
 	if filter.Id == "" && filter.PodSandboxId == "" && len(filter.LabelSelector) == 0 {
-		if err := m.SendEvent(reply.(*criapi.ListContainerStatsResponse).Stats); err != nil {
-			m.Error("failed to inject CRI ListContainerStats for event processing: %v", err)
```
klihub: The reason for the original code to send practically the raw message to the event loop for processing is that we only intercept this request to use its reply internally, without any intention to modify either. Therefore it is possible to minimize the additional delay introduced by cri-resmgr being on the signalling path, if we do the local processing in the event loop which is running in a separate goroutine. So I thin all this processing should be moved to the other end of the event channel.

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/sys v0.0.0-20200202164722-d101bd2416d5
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
+	gonum.org/v1/gonum v0.0.0-20190331200053-3d26580ed485
```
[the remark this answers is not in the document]
mythi: done.

pkg/cri/resource-manager/requests.go
mythi: moved all processing in the handler

mythi:
v2: I decided to keep the flat float64 metrics in the metricsring. The ring now adds EWMA and timestamps to get the samples' time scale.

PR #325, opened 2020-06-24
cri-resmgr,grpclog: rate limit grpc logging.

klihub:
Set up a rate-limited logger for logging gRPC messages. The rate-limiting defaults to 1 unique message per every 5 minutes and can be overridden using environment variables.


klihub:
Let's fix how the agent-to-server and server-to-agent connections are handled. Currently they are unaware about each other and connections are triggered parallel. We should change that to be more like the DeviceManager and device plugins: cri-resmgr creates socket, waits for connection+Register request, connects back to socket indicated in Register request. From that on the connections should be handled in tandem: once goes down, the other should be disconnected, too, and the connection setup should start from the beginning.

PR #327, opened 2020-06-29
publish-docs.yml: use an access token for pushing to gh-pages.

klihub:
Use a github access token for publishing (pushing) to gh-pages.

Automatic publishing will not work until a github token created by a user with sufficient
write/push permissions and named ACCESS_TOKEN (or access-token) gets added to
settings/secrets for the main repository.


klihub:
Let's try getting the original GITHUB_TOKEN-based approach working instead of this...

PR #329, opened 2020-06-29
change avx and cgroupstats collectors to report container ID

mythi:
Both AVX and cgroupstats Prometheus collectors report the cgroup
path differently. What's common is the container id.

Change the collectors to report cgroup_id which is the has value
used by container runtimes.

### DIFF
--- a/pkg/avx/collector.go
+++ b/pkg/avx/collector.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -74,22 +75,21 @@ var descriptors = [numDescriptors]*prometheus.Desc{
 		AVXSwitchCountName,
 		"Number of task switches where AVX512 instructions were used in a particular cgroup.",
 		[]string{
-			"cgroup",
-			"cgroup_id",
+			"container_id",
 		}, nil,
 	),
 	allSwitchCountDesc: prometheus.NewDesc(
 		AllSwitchCountName,
 		"Total number of task switches in a particular cgroup.",
 		[]string{
-			"cgroup",
+			"container_id",
 		}, nil,
 	),
 	lastUpdateNsDesc: prometheus.NewDesc(
 		"last_update_ns",
 		"Time since last AVX512 activity in a particular cgroup.",
 		[]string{
-			"cgroup",
+			"container_id",
 		}, nil,
 	),
 }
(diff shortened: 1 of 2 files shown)


pkg/avx/collector.go
```diff
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"regexp"
```
klihub, line 27: I just wish this could be done without regexp. We have been actively trying to keep out all regexp from file path/content parsers.

That said, I realize that currently we have no real alternative considering the different cgroup drivers and related container ID to path translation. However, let's make a mental note that once we have *proper explicit support* for both the systemd and the (lib)cgroupfs drivers, and consequently the necessary functions in place to convert back and forth between container IDs and cgroup paths, we'll revisit this and seriously consider eliminating regexp usage.

pkg/avx/collector.go
```diff
@@ -74,22 +75,21 @@ var descriptors = [numDescriptors]*prometheus.Desc{
...
-			"cgroup",
 			"cgroup_id",
 		}, nil,
 	),
 	allSwitchCountDesc: prometheus.NewDesc(
 		AllSwitchCountName,
 		"Total number of task switches in a particular cgroup.",
 		[]string{
-			"cgroup",
+			"cgroup_id",
```
klihub, line 85: nitpick: isn't `cgroup_id` a bit misleading term ? IIRC, the original AVX code referred by cgroup ID to the uint64 handle/ID returned by unix.NameToHandleAt()/name_to_handle_at(2).

Shouldn't this be called `container_id` ? Or is this an established nomenclature among other existing upstream/external collectors ?

mythi: I was thinking about `containerd_id` too so I'm ok with that.

pkg/avx/collector.go
mythi: > We have been actively trying to keep out all regexp from file path/content parsers.

OK, I didn't know this. How is the basename for the container path different between the cgroup drivers?

klihub: There are several differences, IIRC. The systemd driver appends the .scope-suffix to directories, the other driver does not. The systemd driver also has an encoding/decoding scheme for certain characters in path names, the other driver does not have it. And I almost forgot: the container cgroup path also depends on the container runtime.

But since our current codebase lacks proper full support for the non-systemd cgroupfs driver and code to return a container ID for a path (exactly what you need here AFAICT), I think this regexp should not be a showstopper. I just wanted to make a note, so that we'll remember this in the future if we ever get the necessary functions to avoid relying on regexp here.

pkg/avx/collector.go
klihub: Ok. Since you later also call the regexp passed as an argument as containerIDRegexp, let's use `container_id` instead.

PR #334, opened 2020-07-01
fixes: don't consider assignment to a default class an error if no classes are defined. 

klihub:
When running without any `RDT` / `BlockIO` classes configured, do not consider failed assignments to `BestEffort`, `Burstable`, or `Guaranteed` an error. These are the default classes inherited from the container's `Pod QoS class`. Also, for both controllers give a warning the first time the controller realizes it is running without any classes configured.

### DIFF
--- a/pkg/cri/resource-manager/cache/utils.go
+++ b/pkg/cri/resource-manager/cache/utils.go
@@ -28,6 +28,15 @@ import (
 
 var memoryCapacity int64
 
+// IsPodQOSClassName returns true if the given class is one of the Pod QOS classes.
+func IsPodQOSClassName(class string) bool {
+	switch corev1.PodQOSClass(class) {
+	case corev1.PodQOSBestEffort, corev1.PodQOSBurstable, corev1.PodQOSGuaranteed:
+		return true
+	}
+	return false
+}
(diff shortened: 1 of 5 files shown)


klihub:
> lgtm.
> only question: have you tried scenario where it starts with fallback config without defined classes, and then reconfigured with agent with definitions of classes. will it work correctly?

Hmm..., I'm testing with a fallback config + agent, but I haven't nuked the cache in ages, so effectively it always starts with the last cached configuration instead of the fallback. I'll test it with a cleaned cache.

klihub:
> > lgtm.
> > only question: have you tried scenario where it starts with fallback config without defined classes, and then reconfigured with agent with definitions of classes. will it work correctly?

For external configuration updates we entirely depend on whether the active policy backend updates the containers. We can probably do better than that (some policies don't have/handle configuration updates at all) for instance like we do in the pending/draft #331 for external policy updates. There the cache is forcibly marking all containers with pending updates for all controllers and triggers a Rebalance().

Probably that would be a better model to handle configuration updates as well. It's a higher-level action though, in the resource-manager. It can be dealt with in another PR.

PR #335, opened 2020-07-05
resource-manager: add command line option for resetting cached configuration.

klihub:
Add command line option `--reset-cached-config` to reset any configuration data stored in the cache.

### DIFF
--- a/pkg/cri/resource-manager/resource-manager.go
+++ b/pkg/cri/resource-manager/resource-manager.go
@@ -78,8 +78,13 @@ func NewResourceManager() (ResourceManager, error) {
 		return nil, err
 	}
 
-	if opt.ResetPolicy {
+	switch {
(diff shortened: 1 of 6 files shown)


pkg/cri/resource-manager/resource-manager.go
```diff
@@ -78,8 +78,13 @@ func NewResourceManager() (ResourceManager, error) {
 		return nil, err
 	}
 
-	if opt.ResetPolicy {
+	switch {
+	case opt.ResetPolicy && opt.ResetConfig:
+		os.Exit(m.resetCachedPolicy() + m.resetCachedConfig())
+	case opt.ResetPolicy:
```
kad, line 84: in theory, here 'fallthrough' can be used instead of trying to do sum of `error` type.

PR #342, opened 2020-07-17
adjustments: fix sample adjustment yaml to match the schema

askervin:


sample-configs/external-adjustment.yaml
```diff
@@ -21,14 +21,13 @@ spec:
...
-        cpu: 1500m
-        memory: 750Mi
-    toptierLimit: 500Mi
+    requests:
+      cpu: 750m
+      memory: 500Mi
+    limits:
+      cpu: 1500m
+      memory: 750Mi
+  toptierLimit: 500Mi
```
kad, line 30: have you checked that? `toptierlimit` supposed to be under `resources`

askervin: In the current [schema](https://github.com/intel/cri-resource-manager/blob/master/pkg/apis/resmgr/v1alpha1/adjustment-schema.yaml#L72) it's in the same level as `resources` and `classes`.

I'll move it under `resources` and check that it'll be handled correctly.

askervin: Aha, found the reason why it's not under `resources`. Currently the [`Resources` field in the `AdjustmentSpec`](https://github.com/intel/cri-resource-manager/blob/master/pkg/apis/resmgr/v1alpha1/types.go#L49) is taken directly from the k8s.io API.

What would you prefer? Should we define our own Resources field structure. I guess it wouldn't hurt, as we already specify our own schema for its contents.

kad: no, I'd prefer to reuse k8s.io specs wherever is possible. so, we just need to make sure that this sample is matching schema as Krisztian defined.

askervin: Yep, this sample works with Krisztian's schema.

Just finished a trial where [`toptierLimit` is part of `resources`](https://github.com/askervin/cri-resource-manager/commit/8294ed8e2abdb6d4cab1c0e5211923e3fe414795) - to understand what this change would take. This trial would still need further changes to cache structure, if we'd like to keep `resources` in Adjustments and `resources` in the cache the same (as cache resources are again from `k8s.io`). I wouldn't like this either.

askervin:
> @askervin Is this ready for review/getting merged from your POV ?

Yep.

Issue #319, opened 2020-06-15
Container Affinity e2e validation


msivosuo:
Implement comprehensive set of container affinity e2e tests for improved validation

msivosuo:
Initial set of e2e validation tests implemented, requires manual validation steps. -> Closing.

PR #349, opened 2020-08-12
tests: add NUMA node testing infra and memtier CPU isolation test

askervin:
- numajson2qemuopts.py: convert NUMA node list (JSON)
  with CPUs, memory and distance definitions to Qemu options.
- demo/lib: add host-create-vm support for pass-through Qemu options
  and NUMA node list in JSON.
- test/numa: add a test/demo for CPU binding.
- Refactor common parts from new test and demo/blockio to demo/lib/vm.
- Implements infra described in Issue #123.
- Implements a part of Container Affinity E2E validation, Issue #319.

### DIFF
--- a/test/numa/README.md
+++ b/test/numa/README.md
@@ -1,0 +1,160 @@
+# CRI Resource Manager - NUMA node tests
+
+## Prerequisites
+
+Install:
+- `docker`
+- `govm`
+  In case of errors in building `govm` with `go get`, or creating a virtual machine (`Error when creating the new VM: repository name must be canonical`), these are the workarounds:
+  ```
+  cd ~/go/src && go get -d github.com/govm-project/govm && cd github.com/govm-project/govm && go install  && cd .. && docker build govm -f govm/Dockerfile -t govm/govm:latest
+  ```
+
+## Usage
+
+```
+[VAR=VALUE...] ./run.sh MODE
+```
+
+Get help on available `VAR=VALUE`'s with `./run.sh help`.
+
+## Test phases
+
+In the *setup phase* NUMA node tests create a virtual machine unless
+it already exists. When it is running, tests create a single-node
+cluster and launches `cri-resmgr` on it, unless they are already
+running.
+
+In the *test phase* NUMA node tests run the default test script, a
+custom script, or give a prompt (`run.sh> `) asking a user to run test
+script commands in the `interactive` mode. *Test scripts* are `bash`
+scripts that can use helper functions for running commands and
+observing the status of the virtual machine and software running on
+it.
+
+In the *tear down phase* NUMA node tests copy logs from the virtual
+machine and finally stop or delete the virtual machine, if that is
+wanted.
+
+## Test modes
+
+- `test` mode runs fast and reports `Test verdict: PASS` or
+  `FAIL`. The exit status is zero if and only if a test passed.
+
+- `play` mode runs the same phases and scripts as the `test` mode, but
+  slower. This is good for following and demonstrating what is
+  happening.
+
+- `interactive` mode runs the setup and tear down phases, but instead
+  of executing a test script it gives an interactive prompt.
+
+Print help to see clean up, execution speed and other options for all
+modes.
+
+## Running from scratch and quick rerun in existing virtual machine
+
+The test will use `govm`-managed virtual machine named in the `vm`
+environment variable. The default is `crirm-test-numa`. If a virtual
+machine with that name exists, the test will be run on it. Otherwise
+the test will create a virtual machine with that name from
+scratch. You can delete a virtual machine with `govm delete NAME`.
+
+If you want rerun the test many times, possibly with different test
+inputs or against different versions of `cri-resmgr`, either use the
+`play` mode or set `cleanup=0` in order to keep the virtual machine
+after each run. Then tests will run in the same single node cluster,
+and the test script will only delete running pods before launching new
+ones.
+
+## Testing locally built cri-resmgr and cri-resmgr from github
+
+If you make changes to `cri-resmgr` sources and rebuild it, you can
+force the test script to reinstall newly built `cri-resmgr` to
+existing virtual machine before rerunning the test:
+
+```
+cri-resource-manager$ make
+cri-resource-manager$ cd test/numa
+numa$ reinstall_cri_resmgr=1 speed=1000 ./run.sh play
+```
+
+You can also let the test script build `cri-resmgr` from the github
+master branch. This takes place inside the virtual machine, so your
+local git sources will not be affected:
+
+```
+numa$ reinstall_cri_resmgr=1 binsrc=github ./run.sh play
+```
+
+## Custom tests
+
+You can run a custom test script in a virtual machine with NUMA
+setup and single-node Kubernetes cluster is running. Example:
+
+```
+$ cat > myscript.sh << EOF
+# create two pods, each requesting two CPUs
+CPU=2 n=2 create guaranteed
+# create four pods, no resource requests
+n=4 create besteffort
+# show pods
+vm-command "kubectl get pods"
+# check that
+#   1. the first two pods are not allowed to use the same CPUs
+#   2. the first two and the last four pods are not allowed to run on
+#      the same CPUs
+#   3. the first two pods get four CPUs in total.
+verify 'cpus[0] & cpus[1] == 0' \
+       '(cpus[2] | cpus[3] | cpus[4] | cpus[5]) & (cpus[0] | cpus[1]) == 0' \
+       'bin(cpus[0] | cpus[1]).count("1") == 4'
+EOF
+$ ./run.sh play myscript.sh
+```
+
+Interactive mode helps developing and debugging custom scripts:
+
+```
+$ ./run.sh interactive
+...
+run.sh> CPU=2 n=2 create guaranteed
+```
+
+You can get help on functions available in test scripts with `./run.sh
+help script`, or with `help` when in the interactive mode.
+
+## Testing different NUMA node configurations
+
+If you change NUMA node topology of an existing virtual machine, you
+must delete the virtual machine first. Otherwise `numanodes` variable
+is ignored and the test will run in the existing NUMA
+configuration. Example:
+
+```
+numa$ govm delete my2x4
+```
+
+Run the test in a VM with two NUMA nodes, 4 CPUs and 4G RAM in each node
+```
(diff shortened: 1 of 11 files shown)


askervin:
This replaces PR #343.

Major differences:
- Old: Pods to be created were given in JSON
  New: Pods to be created are part of test script that can be replaced with a custom script.
- Add `test/numa/run.sh interactive` mode to interactive development of test script with all helper functions available.
- Add `test/numa/run.sh help script` to get help on functions available in test scripts.
- Update `README.md`: remove customizing tests with `pods` JSON, add Custom tests section.

askervin:
Carrying over ideas that should be implemented from the old PR:

- Add a simple helper script `numactlH2numajson` which would take as input output from numactl -H and convert it to the format accepted by this script.

test/numa/run.sh
```diff
@@ -0,0 +1,479 @@
...
+DEMO_LIB_DIR=$(realpath "$SCRIPT_DIR/../../demo/lib")
+BIN_DIR=${bindir-$(realpath "$SCRIPT_DIR/../../bin")}
+OUTPUT_DIR=${outdir-$SCRIPT_DIR/output}
+COMMAND_OUTPUT_DIR=$OUTPUT_DIR/commands
+
+source $DEMO_LIB_DIR/command.bash
+source $DEMO_LIB_DIR/host.bash
+source $DEMO_LIB_DIR/vm.bash
+
+script_source="$(< "$0") $(< "$DEMO_LIB_DIR/vm.bash") $(< "$DEMO_LIB_DIR/command.bash")"
```
klihub, line 19: nitpick: although host.bash does not seem to have functions with help ATM, wouldn't including it here anyway be safer in case we ever add such functions there ?

askervin: Good one, fixed!

askervin:
> This pretty much LGTM.
> 
> Have you tried to see what our `make shellcheck` says about these ? Eventually we should fix/annotate all scripts to get clean shellcheck runs.

Useful points. shellcheck errors in `test/numa/run.sh` fixed.

Documented test phases and the interactive mode to `README`.

Changed the dangerous `cleanup=1` default in the test mode. Now `cleanup` default is `0` for all modes.

test/numa/run.sh
```diff
@@ -0,0 +1,482 @@
...
+    # Usage: verify [EXPR...]
+    #
+    # Run python3 -c "assert(EXPR)" to test that every EXPR is True.
+    # Stop evaluation on the first EXPR not True and stops script.
+    # You can allow script execution to continue after failed verification
+    # by running verify in a subshell (in parenthesis):
+    #   (verify 'False') || echo '...but was expected to fail.'
+    #
+    # Variables available in expressions:
+    #   cpus: dictionary {pod_number: cpu_mask}
```
klihub, line 217: Digging out and making `PolicyJSON` from `/var/lib/cri-resmgr/cache` available would help writing additional policy-specific checks. Adding a script command, for instance `snapshot-cache`, for snapshotting the cache, and making the snapshots easily available would allow a lot of flexibility for verification.

test/numa/README.md
```diff
@@ -0,0 +1,156 @@
+# CRI Resource Manager - NUMA node tests
+
+## Prerequisites
+
+Install:
+- `docker`
+- `govm`
```
klihub, line 7: I would add here that once govm is installed tests might fail with an error like this:
```
Creating /home/kli/vms/imagestime="2020-08-13T19:49:03+03:00" level=info msg="Pulling govm/govm:latest image"
time="2020-08-13T19:49:03+03:00" level=fatal msg="Error when creating the new VM: repository name must be canonical"
```

This is a known issue and there is an [open bug](https://github.com/govm-project/govm/issues/22) about this.
Here is an easy workaround:
```
git clone https://github.com/govm/govm
docker build govm -f govm/Dockerfile -t govm/govm:latest
```

test/numa/run.sh
askervin: Sounds good. I wrote a proof-of-concept for that. I think it makes sense to have it in another PR. It's now in a [development branch](https://github.com/askervin/cri-resource-manager/commits/587_e2e_verify_json) that incudes this PR, so that it can be test.

Direct link to the commit: https://github.com/askervin/cri-resource-manager/commit/58655e98da32aeb3f1feddf1361a325d9ecaa66c

askervin:
@klihub , here's another branch for [numactlH2numajson.py](https://github.com/askervin/cri-resource-manager/commits/587_numactl-H) helper. Now we won't forget it when this PR is closed. :)

test/numa/README.md
askervin: Yeah, I'll add it.

test/numa/run.sh
```diff
@@ -0,0 +1,482 @@
...
+    ( eval "$code" ) || {
+        TEST_FAILURES="${TEST_FAILURES} test script failed"
+    }
+}
+
+# Validate parameters
+mode=$1
+user_script_file=$2
+vm=${vm-"crirm-test-numa"}
+cri_resmgr_cfg=${cri_resmgr_cfg-"cri-resmgr-memtier.cfg"}
```
klihub, line 314: Defaulting to ${SCRIPT_DIR}/cri-resmgr-memtier.cfg would be more convenient. Then one neither would need to `cd test/numa` nor specify `cri_resmgr_cfg=test/numa/cri-resmgr-memtier.cfg` when running `./test/numa/run.sh interactive` for interactive testing.

test/numa/run.sh
```diff
@@ -0,0 +1,482 @@
...
+    vm-wait-process kube-apiserver
+fi
+
+
+declare -A kind_count # associative arrays for counting created objects, like kind_count[pod]=1
+eval "${yaml_in_defaults}"
+if [ "$mode" == "interactive" ]; then
+    # shellcheck disable=SC2162
+    while read -e -p "run.sh> " -a commands; do
+        eval "${commands[@]}"
```
klihub, line 450: I think this should be `eval "${commands[@]}" || :` instead.

Otherwise (now) a `vm-command` failure is just fine, while a 'built-in' command/helper failure (for instance, using `create` with a non-existent or mistyped YAML template name) ends the interactive session. I'm guessing this is unintentional. Since the lack of command-line history forces you to retype many of your commands all the time, this will sooner or later result in a built-in helper failing and the end of the session.

test/numa/run.sh
klihub: Yes, having it on a subsequent PR is the way to go.

test/numa/run.sh
askervin: I've felt the pain of user here as well. Been thrown out of the interactive mode when I forgot to protect myself from `exit` with parenthesis.

Part of me wanted to keep the pain point here, just in case someone someday would like to run
`cat script | ./run.sh interactive` (which doesn't even work at the moment... someone eats stdin before entering interactive mode). But I agree with you: let's make this nice in the user perspective. The name of the mode is `interactive` and not `stdin`, after all.

But the bad news is: `eval ... || :` will not solve the problem because evaluated script calls `exit 1` which causes the process to exit before evaluating the right-hand-side of `||`.

On the other hand, running `eval` always in subshell `( eval ... )` prevents using variables.

Making this nice to user may need a bigger change in the script runner design.

test/numa/run.sh
askervin: Thanks, fixed!

test/numa/run.sh
klihub: Ok. As a band-aid, how about setting a variable for indicating interactive mode (INTERACTIVE_MODE=1) and doing `exit 1` in `command-error` only if that is not set ?

askervin: Ended up patching this so that points where the same script command would have caused test script to exit either directly (`exit 1`) or indirectly (`command-error`, `verify False`), the interactive mode gets reset but it does not exit. This is definitely not perfect, but I didn't want to block only some of the errors from exiting the interactive mode...

Independently of the exit-on-error behavior, this change also makes `interactive` part of the script API, which I think may be quite useful. It's nice to let the test script execute somewhere where a bug might be close, and then enter the interactive mode for debugging.

Example: reproducing (with interactive pauses) a case where the number of reserved and allowed CPUs of a guaranteed pod drops from 4 to 3 after creating another guaranteed pod reserving one CPU. Yet  there would be a free NUMA node available without any reserved CPUs:

```
vm=c4m4x2 numanodes='[{"cpu": 4, "mem": "4G", "nodes": 2}]' code="CPU=4 create guaranteed; report cpus; interactive; create guaranteed; report cpus; interactive" ./run.sh test
```

klihub: As a quick test I added this to command.bash
```
exit-if-not-interactive() {
    if [ -z "$INTERACTIVE_MODE" ]; then
        exit ${1:-1}
    fi
}
```
then replaced the most obvious offending `exit 1`'s in `command.bash:command-error()`, `run.sh:error()`, and `run.sh:verify()` with `exit-if-not-interactive 1`'s and set/cleared `INTERACTIVE_MODE=1` around the interactive mode `while`-loop.
It seemed to do the trick.

askervin: Maybe we'll take this route. I'll integrate your approach and change my self-resetting-interactive mode. The thing I especially like here is that it allows interactive mode to change variable values when it is entered inside a test script.

test/numa/run.sh
```diff
@@ -0,0 +1,499 @@
...
+    #
+    # Enter the interactive mode: read next script commands from
+    # the standard input until "exit".
+    echo "Entering the interactive mode until \"exit\"."
+    while true; do
+        # shellcheck disable=SC2162
+        ( while read -e -p "run.sh> " -a commands; do
+              eval "${commands[@]}"
+          done
+        ) && break
```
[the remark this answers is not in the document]
askervin: I tried to get `rlwrap` working so that commands could have been executed in the current shell process. Couldn't get it working. Tried something like `rlwrap -a -H /tmp/delme.history bash -c 'while read -r -e -a cmd; do echo "$cmd[@]" >> /tmp/delme.fifo; done' &`... so I'm pushing without rlwrap.

test/numa/run.sh
askervin: `exit-if-not-interactive` is now in.

test/numa/run.sh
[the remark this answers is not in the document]
klihub: You have omit the `-e` option from `read` (to disable the usage of readline), otherwise `rlwrap` will not work. But even if you do that `rlwrap` works somewhat crappily for wrapping bash, so maybe we should just forget about it.

PR #353, opened 2020-08-19
e2e tests: enable cri-rm cache reporting and verifications

askervin:
- New "cache" variable holds unmarshaled cri-resmgr cache.
- New "allocations" variable is unmarshaled cri-resmgr
  cache["PolicyJSON"]["allocations"].
- "allocations" and "cache" are available in report, verify and pyexec.
- Change verify() to use pyexec() under the hood.

### DIFF
--- a/test/numa/run.sh
+++ b/test/numa/run.sh
@@ -146,6 +146,15 @@ get-py-cpus() {
     fi
 }
 
+get-py-cache() {
+    # Fetch current cri-resmgr cache from a virtual machine.
+    speed=1000 vm-command "cat \"/var/lib/cri-resmgr/cache\"" >/dev/null 2>&1 || {
+        command-error "fetching cache file failed"
+    }
+    cat > "${OUTPUT_DIR}/cache" <<<"$COMMAND_OUTPUT"
+    py_cache="import json;cache=json.load(open(\"${OUTPUT_DIR}/cache\"));allocations=json.loads(cache['PolicyJSON']['allocations'])"
+}
+
 ### Test script helpers
 
 sleep() { # script API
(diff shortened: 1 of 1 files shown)


askervin:
Example: debugging CPU pinning in the interactive mode:
```
$ ./run.sh interactive
...
Entering the interactive mode until "exit".
run.sh> report cpus
Pod   Cpus_allowed_mask
run.sh> CPU=3 n=2 create guaranteed
...
pod/pod0 created
...
pod/pod1 created
run.sh> report cpus
Pod   Cpus_allowed_mask
pod0  1110000
pod1  1110000
run.sh> report allocations | less
...
```
Allocations report shows that both busyboxes have `'Exclusive': '4-6'`. In this case it is probably a good choice as other numanodes have less than 3 CPUs in the default test configuration.

klihub:
> Example: debugging CPU pinning in the interactive mode:
> 
> ```
> $ ./run.sh interactive
> ...
> Entering the interactive mode until "exit".
> run.sh> report cpus
> Pod   Cpus_allowed_mask
> run.sh> CPU=3 n=2 create guaranteed
> ...
> pod/pod0 created
> ...
> pod/pod1 created
> run.sh> report cpus
> Pod   Cpus_allowed_mask
> pod0  1110000
> pod1  1110000
> run.sh> report allocations | less
> ...
> ```
> 
> Allocations report shows that both busyboxes have `'Exclusive': '4-6'`. In this case it is probably a good choice as other numanodes have less than 3 CPUs in the default test configuration.

Hmm... We still shouldn't allocate exclusively the same set of cpus two or more containers. But that looks like a bug we need to nail down in the memtier policy, not in this test PR.

klihub:
> Hmm... We still shouldn't allocate exclusively the same set of cpus two or more containers. But that looks like a bug we need to nail down in the memtier policy, not in this test PR.

Ok, that's broken in interesting ways:
```
root@crirm-test-numa:/home/ubuntu# cat /var/lib/cri-resmgr/cache | jq .PolicyJSON.allocations | jq fromjson | egrep '("Pool":)|("Exclusive":)'
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "4-6",
    "Pool": "socket #6",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "4-6",
    "Pool": "numa node #3",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
    "Exclusive": "",
    "Pool": "root",
```

Maybe it is the topology pool setup which breaks badly for the default HW config. Anyway, I think this PR is ok.

klihub (approved):
LGTM

klihub (changes requested):
Uhhmkay... So the topology-based pool setup code does go bezerk so it's buggy.

But I'm guessing mostly because it does not verify one of its own implicit assumptions (should be fixed), which is: we only run on HW/CPUs with integration memory controllers. IOW, we always build the tree with a root->sockets->numa nodes->CPU cores hierarchy, never ever root->numa nodes->sockets->CPU cores.

Our default configuration violates this assumption by defining a hardware where two sockets share a single memory controller. That should be changed IMO.

klihub:
> But I'm guessing mostly because it does not verify one of its own implicit assumptions (should be fixed), which is: we only run on HW/CPUs with integrated memory controllers. IOW, we always build the tree with a root->sockets->numa nodes->CPU cores hierarchy, never ever root->numa nodes->sockets->CPU cores.

#356 implements startup-time verification of the current HW/topology assumptions for the `memtier` policy.

askervin:
Insane default HW topology fixed in #358.
In my point of view this patch is good to go.

PR #354, opened 2020-08-19
e2e: fix deprecated node distance syntax in numalist JSON.

askervin:
Distance syntax is was changed to such that all top-level keys have
fixed names instead of "dist-group-SOURCEGROUP"-like names that were
allowed in the first version. The default numanodes value was not
updated at that point, which caused regression.


klihub (approved):
LGTM.

PR #356, opened 2020-08-19
memtier: check HW topology/assumptions during startup.

klihub:
Check our otherwise implicit topology assumptions and refuse
to start if those are not valid for the current HW.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/memtier/pools.go
+++ b/pkg/cri/resource-manager/policy/builtin/memtier/pools.go
@@ -17,6 +17,8 @@ package memtier
 import (
 	"sort"
 
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -93,6 +99,35 @@ func (p *policy) buildPoolsByTopology() error {
 	return nil
 }
 
+// checkHWTopology verifies our otherwise implicit assumptions about the HW.
+func (p *policy) checkHWTopology() error {
+	// NUMA nodes (memory controllers) should not be shared by multiple sockets.
```
kad, line 104: we probably should take care about `die` part of topology reported by kernel

askervin (approved):
Looks good to me. Tested and seems to work great.

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
klihub: Yes, but I think that warrants a separate PR. This PR just adds verification of the current implementations otherwise implicit HW assumptions.

PR #362, opened 2020-08-26
rdt: fix a go1.15 test/build error.

klihub:
Fix a 'string of one rune vs. string of digits' conversion/cast
error, detected when compiling with go1.15 or later.


kad (approved):
lgtm

PR #363, opened 2020-08-26
fix: force coldstart off, disable PMEM-only allocations if we have PMEM in movable memory zones.

klihub:
We can't allocate a container exclusively to PMEM in movable memory zones. In such a case forcibly disable
coldstart and forcibly add DRAM to the memory preferences of all containers.


klihub:
@kad Logic inverted to disable PMEM-only allocations and coldstart if a NUMA node has only memory in movable zones. Also node-zone mapping is now read from /sys/devices/system/node/has*_memory.

kad (approved):
lgtm as first implementation. let's observe about other corner cases.

PR #366, opened 2020-08-31
e2e: "test/numa/run.sh debug" sets up delve in VM

askervin:
This patch makes it easy to start debugging `cri-resmgr` with delve in a fresh VM. Command

```
./run.sh debug
```

Does the following:
- Install `delve` to a VM.
- Copy `cri-resource-manager` sources to the VM.
- Configure `delve` to find the sources on the VM.
- Print help to user how to attach debugger to the running `cri-resmgr` instance, or how to relaunch `cri-resmg` inside the debugger.


klihub (approved):
LGTM

PR #368, opened 2020-09-01
static-pools: advertise cpus in the shared and infra pools

marquiz:
Advertise CPUs belonging to the 'shared' and 'infra' pools (if any) via CMK_CPUS_SHARED and CMK_CPUS_INFRA environment variables. Compatibility backport from CMK

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
@@ -53,6 +53,10 @@ const (
 	// to communicate all CPUs of the infra pool to the workload. We use the
 	// same environment variable for compatibility.
 	CmkEnvInfra = "CMK_CPUS_INFRA"
+	// CmkEnvShared is the name of the env variable that the original CMK sets
+	// to communicate all CPUs of the shared pool to the workload. We use the
+	// same environment variable for compatibility.
+	CmkEnvShared = "CMK_CPUS_SHARED"
 	// CmkEnvNumCores is the name of the env used in the original CMK to select
 	// the number of exclusive CPUs, deprecated here
 	CmkEnvNumCores = "CMK_NUM_CORES"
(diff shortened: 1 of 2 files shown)


marquiz:
Hold for a sec. Haven't tested this properly, yet 😄

klihub:
> Hold for a sec. Haven't tested this properly, yet

Ok, just ping when you consider it ready for review.

marquiz:
> Ok, just ping when you consider it ready for review

Ha, of course the PR was buggy. But, now fixed and tested ➡️ ready for review

pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
```diff
@@ -478,6 +490,18 @@ func (stp *stp) allocateStpResources(c cache.Container, cs stpContainerStatus) e
...
+	pool, ok = stp.conf.Pools[CmkPoolInfra]
+	if ok {
+		c.SetEnv(CmkEnvInfra, pool.cpuSet())
+	}
+
+	// Advertise CPUs belonging to the shared pool
+	pool, ok = stp.conf.Pools[CmkPoolShared]
+	if ok {
+		c.SetEnv(CmkEnvShared, pool.cpuSet())
+	}
```
klihub, line 503: Would it make sense to advertise these also in the sh-syntax `resources file` ?

`cri-resource-manager` injects such a file into every `CreateContainer ` request as `/.cri-resmgr/resources.sh`. To add this data to it, you just need to change the currently empty `ExportResourceData()` to return the desired key-value pairs as a `map[string]string`.

klihub (approved):
LGTM. I have one question though, regarding the possibility to advertise the same data in the stock `resources.sh` file as well.

pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
marquiz: > Would it make sense to advertise these also in the sh-syntax `resources file` ?

Good point, I really don't know 😄  Could you remind me what is the intended use for the file?

In any case I think this should be handled in a separate issue/PR, I opened #370 for discussing this.

klihub: The intended case is to provide an event-driven way (inotify) for (cri-resmgr-aware) containers to monitor and react to changes in their resource assignments.

PR #369, opened 2020-09-02
resource-manager: always setup node agent connection

marquiz:
Always setup connection to the node agent as it serves other functions
than just config handling. Formerly the agent connection was not
initialized if --force-config was used. This was problematic, as e.g.
static-pools policy requires services (other than config) from the agent
and refused to work when forced config was used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ You can do it with the following command:
 ```
 
 When started this way CRI Resource Manager reads its configuration from the
-given file. It also disables its agent interface for external configuration
-and updates.
+given file. It does not fetch external configuration from the node agent and
+also disables the config interface for receiving configuration updates.
 
 ### Using CRI Resource Manager Agent and a ConfigMap
 
--- a/pkg/cri/resource-manager/resource-manager.go
+++ b/pkg/cri/resource-manager/resource-manager.go
@@ -93,10 +93,8 @@ func NewResourceManager() (ResourceManager, error) {
 		return nil, err
 	}
 
-	if opt.ForceConfig == "" {
-		if err := m.setupConfigAgent(); err != nil {
-			return nil, err
-		}
+	if err := m.setupAgentInterface(); err != nil {
+		return nil, err
 	}
 
 	if err := m.loadConfig(); err != nil {
@@ -324,8 +322,8 @@ func (m *resmgr) setupCache() error {
 
 }
 
-// setupConfigAgent sets up the connection to the configuration agent.
-func (m *resmgr) setupConfigAgent() error {
+// setupAgentInterface sets up the connection to the node agent.
+func (m *resmgr) setupAgentInterface() error {
 	var err error
 
 	if m.agent, err = agent.NewAgentInterface(opt.AgentSocket); err != nil {


klihub:
IMO we should not do this unconditionally. I'd rather either
  - have the policy fail to start up if it expects an `agent.Interface` and does not get one, or
  - make the agent.Interface dynamically acquirable and, again, have the policy fail if it requires but cannot get one

The second option above could be implemented by
  - replacing `policy.Backend.AgentCli` by `policy.BackendOptions.GetAgentInterface() (agent.Inteface, error) {}`
  - fixing things at the `resource-manager` level so that the agent connection/interface can be dynamically set up on-demand
Then `GetAgentInterface()` would trigger an attempt to set up the agent connection, returning an `error` if this has been explicitly disabled or fails.

marquiz:
> IMO we should not do this unconditionally. I'd rather either

Why not? We're basically doing all other initializations unconditionally, too.  I think we really should not make any decisions based on `--force-config` as the agent is not only for config related stuff but all access to k8s api. I think the logic should be somewhere else and like "if force_config do not fetch or apply config".

> have the policy fail to start up if it expects an `agent.Interface` and does not get one, or

This is how it works, at the moment. If you use `--force-config` with static-pools enabled cri-resmgr just errors out.

klihub:
> > IMO we should not do this unconditionally. I'd rather either
> 
> Why not? We're basically doing all other initializations unconditionally, too.

Hmm... not all initializations.

> I think we really should not make any decisions based on `--force-config` as the agent is not only
> for config related stuff but all access to k8s api. I think the logic should be somewhere else
> and like "if force_config do not fetch or apply config".
> 
> > have the policy fail to start up if it expects an `agent.Interface` and does not get one, or
> 
> This is how it works, at the moment. If you use `--force-config` with static-pools enabled cri-resmgr just errors out.

So is the basic problem that you would like to allow running/testing full-featured CMK with `--force-config`, including node labeling, declaring extended resources (or is that done by the external CMK init utility ?), or any other potential changes CMK might do on the k8s node ?

If that is the case, then I understand the motivation for the PR. Still, I think it must be possible to disable both agent interfaces. and so far `--force-config` happened to be the way to do that. Would it be better then to
  1. change the defaults for `--force-config` to only disable the downstream agent interface, and
  2. add and explicit `--disable-agent` flag for disabling both the agent interfaces
This way `--force-config` would work the way you intend, and `--force-config` together with `--disable-agent` would revert to the current behavior.

marquiz:
> So is the basic problem that you would like to allow running/testing full-featured CMK with `--force-config`, including node labeling, declaring extended resources (or is that done by the external CMK init utility ?)

Yes, I think it is unintuitive (and makes development harder) if static-pools policy is not available with `--force-config`. It's just forced config IMO and should not "break" anything or limit the usage in any other way. I still fail to see what "harm" it does to setup the agent interface(?) In any case forced config is obeyed and cri-resmgr runs fine even if the agent daeamon is not running.

I think option #2 you outlined above would be the logical thing to do. Either specify a new option (`--disable-agent`) or then disable agent if the `--agent-socket` option is left empty or smth. I feel that the implicit logic / unexpected consequences / hidden meanings baked into `--force-config` should not be there. WDYT?

klihub:
> > So is the basic problem that you would like to allow running/testing full-featured CMK with `--force-config`, including node labeling, declaring extended resources (or is that done by the external CMK init utility ?)
> 
> Yes, I think it is unintuitive (and makes development harder) if static-pools policy is not available with `--force-config`. It's just forced config IMO and should not "break" anything or limit the usage in any other way. I still fail to see what "harm" it does to setup the agent interface(?) In any case forced config is obeyed and cri-resmgr runs fine even if the agent daeamon is not running.
> 
> I think option #2 you outlined above would be the logical thing to do. Either specify a new option (`--disable-agent`) or then disable agent if the `--agent-socket` option is left empty or smth. I feel that the implicit logic / unexpected consequences / hidden meanings baked into `--force-config` should not be there. WDYT?

Yes, I agree both with the preference for `static-pools` to be testable with `--force-config`, and therefore with explicit disabling being needed/preferred instead of the current side-effect.

With that , then this PR is IMO good to go in as such. We can handle disabling the agent with a separate PR.

klihub (approved):
LGTM. Let's handle then disabling the agent with a dedicated PR of its own.

klihub (changes requested):
Actually, could you also update the documentation accordingly ? Currently it states
```
When started this way CRI Resource Manager reads its configuration from the
given file. It also disables its agent interface for external configuration and updates.
```
This is no longer true with this PR merged.

marquiz:
> Actually, could you also update the documentation accordingly ?

K, I tried to reword it a bit

marquiz:
> explicit disabling being needed/preferred instead of the current side-effect.

I created #374 for tracking that. I'm not entirely sure if we need that, though.

klihub (approved):
LGTM.

PR #371, opened 2020-09-02
memtier: produce more correct, easier to read node supply dumps

klihub:
When dumping a node in the tree of pools, calculating free shared CPU considered only
containers directly assigned to the node. Fix this to also account for any other containers
assigned elsewhere in the subtree below the node.

Additionally, split printing memory maps into a function of its own, dump CPU resources
and memory maps in a more compact  form. Finally, provide a function for more compact
human-readable dumps of CPUSets, by combining ranges of CPUs also when the diff
between subsequent elements is other then 1. This can shorten CPUSets like
`0,2,4,6,8,10,12,14,16,18,20` to `0-20:2`.


askervin (approved):
Thanks, this is good for readability. LGTM.

PR #373, opened 2020-09-02
cache: add support for opting out from automatic topology hint generation.

klihub:
cache: add support for opting out from topology hint generation.
    
    Using the new 'topologyhints' resource manager annotation
    key it is now possible to opt out from automatic topology
    hint generation on a per pod or container basis.
    
    Use this annotation to opt out a full pod:
      annotations:
        topologyhints.cri-resource-manager.intel.com/pod: "false"
    
    Use this annotation to opt out container 'foo' in the pod:
      annotations:
        topologyhints.cri-resource-manager.intel.com/container.foo: "false"
    
    Currently topology hint generation is enabled by default, so
    using the annotation as opt in (setting it to "true") should
    have no effect on the placement of containers of a pod. This
    might change in the future however.

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -58,6 +58,9 @@ const (
 
 	// ToptierLimitUnset is the reserved value for indicating unset top tier limits.
 	ToptierLimitUnset int64 = -1
+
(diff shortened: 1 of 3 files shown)


kad (approved):
lgtm

PR #377, opened 2020-09-03
affinity fixes: fix a few doc thinkos, make simple affinities symmetric, improve scoring.

klihub:
- fix affinity full syntax examples in docs
- make simple affinities implicitly symmetric
- filter out affinities for a container on itself
- add better node affinity scoring 

Despite being explicitly documented to the contrary, based on feedback we received users expect
`simple affinity annotations` to result in implicitly symmetric `affinity expressions`.  IOW they expect
that for every pair of containers X, Y an existing affinity X: Y implies Y: X. Considering the semantics
(Y pulls X), the syntax (almost a matrix of containers) and the fact that the simple notation very much
hides the true versatility of affinity expressions, this expectation is rather understandable.
Therefore we now implicitly generate a fully symmetric set of affinities from the simple notation.

Calculate affinity for every pool node n as a combination of affinities of the nodes x along the path
from n to root and those in the subtree rooted at n. The resulting affinity is a sum of raw affinities of
these nodes with every raw node affinity diluted (divided) by 1<<(distance from n to x).

### DIFF
--- a/docs/container-affinity.md
+++ b/docs/container-affinity.md
@@ -45,7 +45,7 @@ metadata:
             - value1
             ...
             - valueN
-        - match:
+          match:
             key: key-ref
             operator: op
             values:
@@ -70,7 +70,7 @@ metadata:
             - value1
             ...
             - valueN
-        - match:
+          match:
             key: key-ref
             operator: op
             values:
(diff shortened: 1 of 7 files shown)


klihub:
Closing, split up to #378, #379, #380, and #381.

PR #379, opened 2020-09-04
docs: fix affinity full notation syntax examples.

klihub:


kad (approved):
lgtm

This makes me thing that we need to reject Pod creation if the annotations for CRI-R-M can't be properly parsed.

PR #380, opened 2020-09-04
cache: make simple affinities implicitly symmetric.

klihub:
For affinities given in the simple notation always
generate a symmetric set of affinity expressions. IOW,
for every affinity X: Y with weight W, make sure there
is an affinity Y: X with weight W. |W| is now implicitly
DefaultWeight.


kad (approved):
lgtm

PR #372, opened 2020-09-02
memtier: a bunch of resource allocation/accounting fixes.

klihub:
memtier: take parent grants into account for allocatable shared calculation
memtier: propagate exclusive allocations to parent pools/nodes
memtier: bugfixes for slicing off shared CPUs, prevent full exhaustion of shared CPU subsets of pools

PR #381, opened 2020-09-04
memtier: better affinity scoring

klihub:
Notes:
  - This PR has been updated by pre-merging pending PRs #406. That should be reviewed separately.
  - The algorithm has been slightly updated to dilute node/pool scores along the path less aggressively.

Calculate affinity for every node n based on the raw
affinities of the nodes along the path from n to root
and all the sub-nodes in the tree rooted at n.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/memtier/pools.go
+++ b/pkg/cri/resource-manager/policy/builtin/memtier/pools.go
@@ -15,6 +15,7 @@
 package memtier
 
 import (
+	"math"
 	"sort"
 
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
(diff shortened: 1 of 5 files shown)


pkg/cri/resource-manager/policy/builtin/memtier/pools_test.go
```diff
@@ -33,23 +33,37 @@ import (
...
 		if node.NodeID() == id {
-			return &node
+			return node
 		}
 	}
 	panic("No node found")
 }
 
 func setLinks(nodes []Node, tree map[int][]int) {
+	parents := map[int]int{}
```
marquiz, line 46: Same comment than in #372: how about `nonOrphans := make(map[int]struct{})` or something similar(?)

marquiz (commented):
One comment to the code. No comments to the algorithm itself, it looks great 😊

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -837,6 +838,24 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
 	return id1 < id2
 }
 
+// affinityScore calculate the 'goodness' of the affinity for a node.
```
marquiz, line 841: Hey, one more comment: you have a very good description of what this is about in the commit message. Could/should that somehow be copy-pasted or referenced here, too(?)

pkg/cri/resource-manager/policy/builtin/memtier/pools_test.go
klihub: I'll think about something. nonOrphans is not a good choice though, IMO. We do collect parents in that map, basically constructing an alternative representation of the input `tree`, using per-node parent links instead of per-node children. Parent id being collected in vain as unused is a valid comment, though.

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
klihub: I'll do that.

marquiz (dismissed):
Thanks @klihub 
lgtm

kad (approved):
lgtm

PR #382, opened 2020-09-04
Add RELEASE.md

marquiz:
Describe the release process of cri-resource-manager.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,27 @@
+# Release Process
+
+The process to release a new version of cri-resource-manager is:
+
+- [ ] File [a new issue](https://github.com/intel/cri-resource-manager/issues/new)
+  to propose a new release. Copy this checklist into the issue description.
+- [ ] In the issue description, add a changelog section, describing changes
+  since the last release.
+- Local release preparations
+  - [ ] Mandatory internal release checks and preparations.
+  - [ ] Run `make release-tests` to run an extended set of tests prior to a release
+  - [ ] Run `git tag -a -m "CRI Resource Manager release $VERSION" $VERSION`.
+  - [ ] Create src tarballs with `make vendored-dist`
+  - [ ] Build RPM packages with `make cross-rpm`
+  - [ ] Build DEB packages with `make cross-deb`
+  - [ ] Upload all artefacts to Github release page.
+  - [ ] Build container images with `make images`
+- Final verification of artefacts
+  - [ ] Verify the installation of binary packages
+  - [ ] Verify runnability of container images
+- Publishing
+  - [ ] Push the tag with `git push $VERSION`
+  - [ ] Push container images *to where*???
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,20 @@
+# Release Process
+
+The process to release a new version of cri-resource-manager is:
+
+- [ ] File [a new issue](https://github.com/intel/cri-resource-manager/issues/new)
+  to propose a new release. Copy this checklist into the issue description.
+- [ ] In the issue description, add a changelog section, describing changes
+  since the last release.
+- [ ] Run `git tag -a -m "CRI Resource Manager release $VERSION" $VERSION`.
```
klihub, line 9: I would suggest adding
```
 - [ ] run `make release-tests` to run an extended set of tests prior to a release
```
And add already now an empty `release-tests` target to our `Makefile` which we then start
populating with things like running end-to-end test, etc. as we go/get them up and running.

klihub:
Looks good, all of these make sense to me.

RELEASE.md
klihub: Should we add steps covering our internally mandated other checks prior to a release ?

klihub (dismissed):

klihub:
One more question: #386 proposes to add INSTALL.md as `docs/INSTALL.md` which is then referenced in `README.md`.
Should we put both in `docs`, or is `RELEASE.md` considered more of a self-contained entity which is not referenced from
anywhere else in the documentation ?

RELEASE.md
marquiz: Yeah, I'll add these 👍

marquiz:
> Should we put both in `docs`, or is `RELEASE.md` considered more of a self-contained entity which is not referenced from
> anywhere else in the documentation ?

I personally probably see `RELEASE.md` as maintainer aid and not so much part of the user/developer documentation. Thus, I'd just probably keep it in the root alongside `README.md`. I don't have any strong opinion on this, though, so we can move it under docs if you think that would be more appropriate.

klihub:
> > Should we put both in `docs`, or is `RELEASE.md` considered more of a self-contained entity which is not referenced from
> > anywhere else in the documentation ?
> 
> I personally probably see `RELEASE.md` as maintainer aid and not so much part of the user/developer documentation. Thus, I'd just probably keep it in the root alongside `README.md`. I don't have any strong opinion on this, though, so we can move it under docs if you think that would be more appropriate.

Ok. Makes sense. We can leave it where it is.

RELEASE.md
marquiz: Added both

marquiz:
> Ok. Makes sense. We can leave it where it is.

K, and we can always move it later if we change our mind.

Any comments on the latest (indented) version?

klihub (approved):
LGTM.

PR #383, opened 2020-09-04
e2e: integrate resource allocations topology to test scripts

askervin:
- Change variables visible in test script functions "verify",
  "report" and "pyexec". Details below.
- Add allowed_RESOURCE[POD] variables to enable verifications
  like verify 'len(allowed_nodes["pod0"]) == 1'.
- Add \"allowed\" variable to contain topology tree with pods as leaf
  nodes on branches they are allowed to use.
- "report allowed" prints full topology tree with pods allowed to
  use the resources.
- Drop "cpus" variable showing CPU masks without topology. The same
  information is now available through allowed_cpus[pod].

### DIFF
--- a/demo/lib/topology.py
+++ b/demo/lib/topology.py
@@ -6,9 +6,10 @@ Usage: topology.py [options] command
 
 Options:
   -t TOPOLOGY_DUMP    load topology_dump from TOPOLOGY_DUMP file instead of
-                      the topology_dump environment variable or local host.
+                      the "topology_dump" environment variable or local host.
   -r RES_ALLOWED      load res_allowed from RES_ALLOWED file instead of
-                      the res_allowed environment variable or local host.
+                      the "res_allowed" environment variable or local host.
+  -o OUTPUT_FORMAT    "json" or "text". The default is "text".
 
 Commands:
   help                print help
@@ -42,10 +43,10 @@ Examples:
   Watch how pod0..2 are allowed to CPUS on remote host, read topology only once
   $ export topology_dump="$(ssh remotehost "$(topology.py bash_topology_dump)")"
   $ watch 'res_allowed=$(ssh remotehost "$(topology.py bash_res_allowed pod0 pod1 pod2)") topology.py res_allowed'
-
 """
 
 import getopt
+import json
 import os
 import re
 import subprocess
@@ -53,7 +54,7 @@ import sys
 
 _bash_topology_dump = """for cpu in /sys/devices/system/cpu/cpu[0-9]*; do cpu_id=${cpu#/sys/devices/system/cpu/cpu}; echo "cpu p:$(< ${cpu}/topology/physical_package_id) d:$(< ${cpu}/topology/die_id) n:$(basename  ${cpu}/node* | sed 's:node::g') c:$(< ${cpu}/topology/core_id) t:$(< ${cpu}/topology/thread_siblings) cpu:${cpu_id}" ; done;  for node in /sys/devices/system/node/node[0-9]*; do node_id=${node#/sys/devices/system/node/node}; echo "dist n:$node_id d:$(< $node/distance)"; echo "mem n:$node_id s:$(awk '/MemTotal/{print $4/1024}' < $node/meminfo)"; done"""
 
-_bash_res_allowed = r"""for process in '%s'; do for pid in $(pgrep -f $process); do [ "$pid" != "$$" ] && [ "$pid" != "$PPID" ] && echo "${process}-${pid} $(awk '/Cpus_allowed:/{c=$2}/Mems_allowed:/{m=$2}END{print "c:"c" m:"m}' < /proc/$pid/status)"; done; done"""
+_bash_res_allowed = r"""for process in '%s'; do for pid in $(pgrep -f $process); do [ "$pid" != "$$" ] && [ "$pid" != "$PPID" ] && echo "${process}/${pid} $(awk '/Cpus_allowed:/{c=$2}/Mems_allowed:/{m=$2}END{print "c:"c" m:"m}' < /proc/$pid/status)"; done; done"""
 
 def error(msg, exit_status=1):
     """Print error message and exit."""
@@ -65,6 +66,14 @@ def error(msg, exit_status=1):
 def warning(msg):
     """Print warning."""
     sys.stderr.write('topology.py warning: %s\n' % (msg,))
+
+def output_tree(tree):
+    """Print tree to output in OUTPUT_FORMAT"""
+    if opt_output_format == "json":
+        sys.stdout.write(json.dumps(tree))
+    else:
+        sys.stdout.write(str_tree(tree) + "\n")
+    sys.stdout.flush()
 
 def add_tree(root, branch, value_dict):
     """Add key-value pairs in value_dict to given branch in the tree starting from root.
@@ -271,8 +280,8 @@ def get_topology(show_mem=True):
     if opt_topology_dump:
         topology_dump = opt_topology_dump
     else:
-        topology_dump = os.getenv("topology_dump")
-    if not topology_dump:
+        topology_dump = os.getenv("topology_dump", None)
+    if topology_dump is None:
         topology_dump = get_local_topology_dump()
     return dump_to_topology(topology_dump, show_mem=show_mem)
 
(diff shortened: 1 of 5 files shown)


askervin:
This patch enables writing tests with intermediate and final verifications on topology. Example:

```
$ cat > my.script << EOF
CPU=3 create guaranteed
CPU=4 create guaranteed
verify \
    'len(allowed_cpus["pod0"]) == 3' \
    'len(allowed_cpus["pod1"]) == 4' \
    'allowed_cpus["pod1"].intersection( allowed_cpus["pod0"] ) == set()'
$ vm=c32d4p2 ./run.sh test my.script
...
### Verifying assertion 'len(allowed_cpus["pod0"]) == 3'

### The assertion holds.

### Verifying assertion 'len(allowed_cpus["pod1"]) == 4'
### The assertion holds.

### Verifying assertion 'allowed_cpus["pod1"].intersection( allowed_cpus["pod0"] ) == set()'
### The assertion holds.
...
Test verdict: PASS
```

klihub:
> This patch enables writing tests with intermediate and final verifications on topology. Example:
> 
> ```
> $ cat > my.script << EOF
> CPU=3 create guaranteed
> CPU=4 create guaranteed
> verify \
>     'len(allowed_cpus["pod0"]) == 3' \

My most important comment is that we should have the ability to verify
allocations at the container granularity.

>     'len(allowed_cpus["pod1"]) == 4' \
>     'allowed_cpus["pod1"].intersection( allowed_cpus["pod0"] ) == set()'

Can we leave off the `allowed_` prefix in the verification context and refer
to the verified objects a `cpus`, `mems`/`memories`/`numanodes`, and
`pools` ? I think we'll only have conceptually allowed ones, so there is
no need to keep emphasizing/repeating it.

If there is a namespacing problem or fear of clash then we could expose
assignments as part of an encapsulating top-level `allocations` or `assignments`
object, using which we could go  `allocations->pods->containers->{cpus,mems,[pool]}`
> 
> ### Verifying assertion 'allowed_cpus["pod1"].intersection( allowed_cpus["pod0"] ) == set()'
> ### The assertion holds.
> ...
> Test verdict: PASS
> ```

Good ! This looks very promising ! Let's just make sure we get the necessary granularity
for verifications.

klihub:
> > My most important comment is that we should have the ability to verify
> > allocations at the container granularity.
> 
> Thanks, good point. This has definitely looked a bit confusing so far.
> 
> I'll change pod templates (like `guaranteed.yaml.in`) so that the first container will run a process that will match with a string like `pod0c0` (for pod0 container0). If we'll add multicontainer pods to tests later on, next container will be `pod0c1` and so on. Process matching is based on `pgrep -f STRING`, and now `STRING` has happened to be confusingly `pod0`. So I think `s/echo ${NAME} /echo ${NAME}c0/' on `*.yaml.in` will do the trick.
> 
> This will also solve another problem that hasn't bitten us yet: `pod10` would also match `pod1`...
> 
> > > ```
> > > 'allowed_cpus["pod1"].intersection( allowed_cpus["pod0"] ) == set()'
> > > ```
> > 
> > 
> > Can we leave off the `allowed_` prefix in the verification context and refer
> > to the verified objects a `cpus`, `mems`/`memories`/`numanodes`, and
> > `pools` ? I think we'll only have conceptually allowed ones, so there is
> > no need to keep emphasizing/repeating it.
> 
> Sure. That'd shorten up verifications nicely.
> 
> > If there is a namespacing problem or fear of clash then we could expose
> > assignments as part of an encapsulating top-level `allocations` or `assignments`
> > object, using which we could go `allocations->pods->containers->{cpus,mems,[pool]}`
> 
> Do you think test verifications should have similar shorthands for accessing cri-resmgr's view on granted cpus etc? If so, we might need `cpus` in another namespace for that. Perhaps something like `os.cpus` and `crirm.cpus`...

For verification, providing access to the `cri-rm cache` content is the most important thing (which I think we are somewhat
capable of already). That allows us to dig out what cri-rm has `decided`, verify the correctness of that, and then verify it
against what has been `enforced` in practice (content of `cgroupfs` +/or `procfs`).

About how we expose data to the tests, I think there are two reasonable approaches:
  1) complete transparency
    Provide access to `cri-rm cache content` (IOW, pods, container, decisions), plus probably some
    helper functions, basically accessors to fetch of-used data like CPU and mems, so that writing
    test cases will not increase the cost of changes to the cache linearly with the number of existing
    test cases. Provide `per-pod:container` access to `cgroupfs cpuset.cpus,mems}` content,
    `pid`-list, and per-pid access to the content of `/proc/$pid/{status+anything else necessary}`. And
    make access to these intuitive and easy to consume. Both of these are very subjective measures,
    but I think data collected to JSON with suitable designed content and indexing/keys should fulfill
    both.
  2) intuitive data model
    Basically provide all data in a way that make sense to someone who knows what cri-rm is supposed
    to do without necessarily understanding details of the underlying OS abstractions (cgroupfs-/procfs
    filesystem entry paths, etc).

I don't know which one is really better, but it is possible to implement 2 on top of 1.

askervin:
> About how we expose data to the tests, I think there are two reasonable approaches:
> 
> 1. complete transparency

This is what we have right in the form of `cache` and `allocations` (the latter being shorthand for `allocations=json.loads(cache['PolicyJSON']['allocations'])`).

> 2. intuitive data model

To ease using these structures in test scripts, I think it would be nice to access `allocations` by strings like `"podXcY"`, similarly to `packages, dies, ...`.

Do you have some other parts of the cache that we should make easy to access?

klihub:
> > About how we expose data to the tests, I think there are two reasonable approaches:
> > 
> > 1. complete transparency
> 
> This is what we have right in the form of `cache` and `allocations` (the latter being shorthand for `allocations=json.loads(cache['PolicyJSON']['allocations'])`).
> 
> > 1. intuitive data model
> 
> To ease using these structures in test scripts, I think it would be nice to access `allocations` by strings like `"podXcY"`, similarly to `packages, dies, ...`.

Yes, I also think that'd be a good idea.

> 
> Do you have some other parts of the cache that we should make easy to access?

Maybe `pods` and `containers` as shorthand notations for `cache['Pods']` and `cache['Containers']`. But IMO that only makes sense if we also provide additional data or helper functions for getting from an `allocations` entry to the corresponding `cache['Pods']` or `cache['Containers']` entry.

askervin:
> Maybe `pods` and `containers` as shorthand notations for `cache['Pods']` and `cache['Containers']`. But IMO that only makes sense if we also provide additional data or helper functions for getting from an `allocations` entry to the corresponding `cache['Pods']` or `cache['Containers']` entry.

Nice. Now there are shorthands to these pieces of cache in place, and they can be indexed with test pod/container names (like pod0 and pod0c0). Added also `kubectl` and `pp` helper functions to improve interactive mode user experience in manual testing and debugging.

```
run.sh> verify # this updates all python variables in pyexec environment (pyexec_state.py)

run.sh> pp 'allocations["pod1c0"]'
{'ColdStart': 0,
 'Container': 'c7ca3fd1-a2bf-4c23-857d-0e89fa9be596:busybox',
 'Exclusive': '2-3',
 'MemType': 'DRAM,PMEM,HBM',
 'MemoryLimit': {'0': 0, '1': 100000000, '2': 0, '4': 0, '7': 0},
 'MemoryPool': 'numa node #1',
 'Memset': '1',
 'Part': 0,
 'Pool': 'numa node #1'}

run.sh> pp 'containers["pod1c0"]["LinuxReq"]'
{'cpu_period': 100000,
 'cpu_shares': 2,
 'cpuset_cpus': '4-7',
 'cpuset_mems': '1',
 'oom_score_adj': 1000}

run.sh> pp 'pods["pod1"]'
{'Affinity': {},
 'Annotations': {'kubernetes.io/config.seen': '2020-09-07T12:50:30.255132778Z',
                 'kubernetes.io/config.source': 'api'},
 'CgroupParent': '/kubepods/besteffort/pod528934ed-5ea8-4d72-ae26-663d0dbdac9b',
 'ID': '78d31225de1afdf9503647453f1b136fe74e70c5166640a69d3b888021f9891d',
 'Labels': {'app': 'pod1',
            'io.kubernetes.pod.name': 'pod1',
            'io.kubernetes.pod.namespace': 'default',
            'io.kubernetes.pod.uid': '528934ed-5ea8-4d72-ae26-663d0dbdac9b'},
 'Name': 'pod1',
 'Namespace': 'default',
 'QOSClass': 'BestEffort',
 'Resources': {'containers': None, 'initContainers': None},
 'State': 0,
 'UID': '528934ed-5ea8-4d72-ae26-663d0dbdac9b'}
```

klihub (approved):
Really nice work !

PR #384, opened 2020-09-04
e2e: first tests for CPU pinning and a test runner

askervin:
This patch introduces initial version of E2E test directory hierarchy: `policy/topology/test-XX-...` and a very very tiny test runner that finds and runs tests from the hierarchy. (Depending on the route we take, we might implement a runner with go + some go test framework...)

Current goals of this patch are two-fold:
- Get `memtier` CPU pinning properly tested with a regression test set that we can run and extend later on.
- Try out if the test hierarchy feels flexible and good for the purpose.

Tests in this patch are using PR #383 for validating pinned resources. The PR #383 is included in this branch.

Tests FAIL for current master (commit 490f87a3f77c805a3dfba064a44bfbf497494df6) but PASS for PR #372.


klihub:
@askervin This (HEAD^) should be git rebased --onto master.

test/numa/run_tests.sh
```diff
@@ -0,0 +1,71 @@
...
+        vm="$(basename "$TOPOLOGY_DIR")"
+        export vm
+        for TEST_DIR in "$TOPOLOGY_DIR"/*; do
+            if ! [ -d "$TEST_DIR" ]; then
+                continue
+            fi
+            export_var_files "$TEST_DIR"
+            export outdir="$TEST_DIR/output"
+            mkdir -p "$outdir"
+            ./run.sh test 2>&1 | tee "$outdir/run.sh.output"
```
klihub, line 68: This is the one to get replaced with `$RUN_SH test ...` mentioned in the previous comment.

test/numa/tests/memtier/n4c16/test01-only-add-pods/code.var.sh
```diff
@@ -0,0 +1,37 @@
...
+verify \
+    'len(allowed_cpus["pod0"]) == 1'
+report allowed
+
+CPU=2 create guaranteed
+verify \
+    'len(allowed_cpus["pod0"]) == 1' \
+    'len(allowed_cpus["pod1"]) == 2' \
+    'len(allowed_cores["pod1"]) == 1' \
+    'len(allowed_cpus["pod0"].union(allowed_cpus["pod1"])) == 3'
```
klihub, line 11: After having `git rebase HEAD^ --onto master`'d this PR, I think we'll need to replace `allowed_{cpus,threads,cores,nodes}` with `{cpus,threads,cores,nodes}`.

test/numa/tests/memtier/n4c16/test01-only-add-pods/code.var.sh
```diff
@@ -0,0 +1,37 @@
...
+
+CPU=3 create guaranteed
+verify \
+    'len(allowed_cpus["pod0"]) == 1' \
+    'len(allowed_cpus["pod1"]) == 2' \
+    'len(allowed_cores["pod1"]) == 1' \
+    'len(allowed_cpus["pod2"]) == 3' \
+    'len(allowed_cores["pod2"]) == 2' \
+    'len(allowed_nodes["pod2"]) == 1' \
+    'len(allowed_cpus["pod0"].union(allowed_cpus["pod1"]).union(allowed_cpus["pod2"])) == 6'
```
klihub, line 22: Maybe it would be a good idea to run this input files through a simple pre-processing script
which performs a set of simple well defined (and documented) translations on the input and
then source the resulting output. That would make it possible to implement conventions that
would allow leaving a lot of that almost-implied repetitive stuff out which I suspect will quickly
become annoying when writing tests. 

For instance, this could then become something like this, with implied leading `verify '` and
a trailing `'` for every line starting with TAB after a single `verify`. There is probably a hugton
of other similar useful conventions that can really make writing these more convenient.
```
CPU=3 create guaranteed
verify
	len(allowed_cpus["pod0"]) == 1
	len(allowed_cpus["pod0"]) == 1
	len(allowed_cpus["pod0"]) == 1
...
```

test/numa/run_tests.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+TESTS_DIR="$1"
+
+usage() {
+    echo "Usage: run_tests.sh TESTS_DIR"
+    echo "TESTS_DIR is expected to be structured as POLICY/TOPOLOGY/TEST with files:"
+    echo "POLICY/cri-resmgr.cfg: configuration of cri-resmgr"
+    echo "POLICY/TOPOLOGY/topology.var.json: contents of the topology variable for run.sh"
+    echo "POLICY/TOPOLOGY/code.var.sh: contents of the code var (that is, test script)"
```
klihub, line 10: This is a good start. But let's try to write a bunch of test cases (somehow) split into a number
of logical groups and see if this fits the bill.

klihub:
It would be good to also add function for scriptable pod YAML generation, with configurability (but having sane defaults) for:
  - pod name
  - labels
  - annotations
  - either number or names of containers
  - per container
    - images
    - resource requests and limits (both optional for any and all resources)
    - maybe the command, but I think we could do without if it defaults to something sane

test/numa/run.sh
```diff
@@ -153,17 +153,48 @@ screen-launch-cri-resmgr-agent() {
...
+        allowed_cores[pod] = set()
+        allowed_threads[pod] = set()
+        allowed_cpus[pod] = set()
+    allowed_packages[pod].add(p)
+    allowed_dies[pod].add('%s/%s' % (p, d))
+    allowed_nodes[pod].add(n)
+    allowed_cores[pod].add('%s/%s' % (n, c))
+    allowed_threads[pod].add('%s/%s/%s' % (n, c, t))
+    allowed_cpus[pod].add(cpu)
+"
```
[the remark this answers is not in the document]
askervin: Helpers added.

test/numa/run_tests.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
askervin: Done, much better now.

test/numa/tests/memtier/n4c16/test01-only-add-pods/code.var.sh
```diff
@@ -0,0 +1,125 @@
...
+    'len(cpus["pod0c0"]) == 1'
+
+# pod1, takes full core
+CPU=2 create guaranteed
+report allowed
+verify \
+    'len(cpus["pod0c0"]) == 1' \
+    'len(cpus["pod1c0"]) == 2' \
+    'len(cores["pod1c0"]) == 1' \
+    'len(set.union(cpus["pod0c0"], cpus["pod1c0"])) == 3'
```
klihub, line 17: I understand that set theory and these 4 assertions imply that the intersection of cpus["pod0c0"] and cpus["pod1c0"] is empty, yet I would feel safer/prefer replacing the last assertion with an intersection check, because it is just much more natural considering the semantics of guaranteed pods with exclusive/full integer CPU requirements. And this applies to most of the subsequent checks to the same effect.

klihub, line 17: For the simplest basic cases, shouldn't we spell out/check the explicit ID of the cpu we expect to get allocated ? It is pretty well defined and intentional behavior, after all.

test/numa/tests/memtier/n4c16/topology.var.json
```diff
@@ -0,0 +1,5 @@
+[
+    {"mem": "1G", "cores": 2, "nodes": 2, "packages": 2},
+    {"nvmem": "8G", "node-dist": {"5": 28, "0": 17}},
```
klihub, line 3: I wouldn't include nvmem/PMEM in the basic test set, since we have neither positive nor negative checks for them. We'll probably create a dedicated test set for it later.

klihub (commented):
Looks really good IMO. I only have a few nitpicks so far.

I still need to go through the actual test cases with a fresh(er) mind tomorrow, and also check if there is anything else we should try to squeeze in to the initial PR/release.

test/numa/tests/memtier/n4c16/test01-only-add-pods/code.var.sh
askervin: Sure, I'll change the last ones to `set.intersection` and mark resolved when done.

askervin: Ok. I've tried to avoid failing a test when pinning is any of the optimal ones. But sure, if you'd prefer fixing CPU ids, then I'll change the most obvious cases to that.

test/numa/tests/memtier/n4c16/topology.var.json
askervin: Ok, I'll change topology to more basic one.

On one hand, topologies are very expensive as they require their own VMs. On the other hand, it's the separate topologies that enable cleanest cut to run tests in parallel at some point...

test/numa/run.sh
```diff
@@ -419,6 +432,17 @@ create() { # script API
...
+    # Usage: reset counters
+    #
+    # Resets counters
+    if [ "$1" == "counters" ]; then
+        kind_count[pod]=0
+    else
+        error "invalid reset \"$1\""
+    fi
+}
+
```
klihub, line 445: Could we add something like this below for `run.sh:create()` ? Otherwise it still forces folks to run it from a particular directory. As an added benefit this would allow policies, test topologies, and tests to override the default templates provided in $SCRIPT_DIR as/if necessary.

```
diff --git a/test/numa/run.sh b/test/numa/run.sh
index 7689a628..5dca9c65 100755
--- a/test/numa/run.sh
+++ b/test/numa/run.sh
@@ -388,12 +388,38 @@ delete() { # script API
     }
 }
 
+resolve-template() {
+    local name="$1" r="" d t
+    shift
+    for d in $*; do
+        if [ -z "$d" ] || ! [ -d "$d" ]; then
+            continue
+        fi
+        t=$d/$name.yaml.in
+        if ! [ -e $t ]; then
+            continue
+        fi
+        if [ -z "$r" ]; then
+            r="$t"
+        else
+            echo "WARNING: template file $r shadows $t"
+        fi
+    done
+    if [ -n "$r" ]; then
+        echo "$r"
+        return 0
+    fi
+    return 1
+}
+
 create() { # script API
     # Usage: [VAR=VALUE][n=COUNT] create TEMPLATE_NAME
     #
     # Create n instances from TEMPLATE_NAME.yaml.in, copy each of them
     # from host to vm, kubectl create -f them, and wait for them
-    # becoming Ready.
+    # becoming Ready. Templates are searched in $TEST_DIR, $TOPOLOGY_DIR,
+    # $POLICY_DIR, and $SCRIPT_DIR in this order of preference. The first
+    # template found is used.
     #
     # Parameters:
     #   TEMPLATE_NAME: the name of the template without extension (.yaml.in)
@@ -402,21 +428,23 @@ create() { # script API
     #   wait: condition to be waited for (see kubectl wait --for=condition=).
     #         If empty (""), skip waiting. The default is wait="Ready".
     #   wait_t: wait timeout. The default is wait_t=60s.
-    template_name=$1.yaml.in
+
+    template_file=$(resolve-template $1 $TEST_DIR $TOPOLOGY_DIR $POLICY_DIR $SCRIPT_DIR)
+
     local template_kind
-    template_kind=$(awk '/kind/{print tolower($2)}' < "$template_name")
+    template_kind=$(awk '/kind/{print tolower($2)}' < "$template_file")
     local wait=${wait-Ready}
     local wait_t=${wait_t-60s}
     if [ -z "$n" ]; then
         local n=1
     fi
-    if [ ! -f "$template_name" ]; then
-        error "error creating \"$1\": missing template ${template_name}"
+    if [ ! -f "$template_file" ]; then
+        error "error creating \"$1\": missing template ${template_file}"
     fi
     for _ in $(seq 1 $n); do
         kind_count[$template_kind]=$(( ${kind_count[$template_kind]} + 1 ))
         local NAME="${template_kind}$(( ${kind_count[$template_kind]} - 1 ))" # the first pod is pod0
-        eval "echo -e \"$(<"${template_name}")\"" > "$NAME.yaml"
+        eval "echo -e \"$(<"${template_file}")\"" > "$NAME.yaml"
         host-command "scp $NAME.yaml $VM_SSH_USER@$VM_IP:" || {
             command-error "copying $NAME.yaml to VM failed"
         }

```

test/numa/run.sh
```diff
@@ -419,6 +472,17 @@ create() { # script API
...
```
klihub, line 485: Can we have something like this, to automatically drop into interactive() at the end of each verify if/as the user asks for it ?
```
diff --git a/test/numa/run.sh b/test/numa/run.sh
index 966990ee..d7c0f013 100755
--- a/test/numa/run.sh
+++ b/test/numa/run.sh
@@ -54,6 +54,11 @@ usage() {
     echo "             0: leave VM running (the default)"
     echo "             1: delete VM"
     echo "             2: stop VM, but do not delete it."
+    echo "    interactive_verify: if/when to go automatically interactive in verify():"
+    echo "             always: go interactive regardless the result of verification"
+    echo "             fail: go interactive if verification fails"
+    echo "             never: never go interactive"
+    echo "             The default is never."
     echo ""
     echo "  Test input VARs:"
     echo "    topology: JSON to override NUMA node list used in tests."
@@ -361,12 +366,19 @@ verify() { # script API
     get-py-cache
     for py_assertion in "$@"; do
         speed=1000 out "### Verifying assertion '$py_assertion'"
-        ( speed=1000 pyexec "assert(${py_assertion})" ) || {
+        if ! ( speed=1000 pyexec "assert(${py_assertion})" ); then
                 out "### The assertion FAILED"
                 echo "verify: assertion '$py_assertion' failed." >> "$SUMMARY_FILE"
-                command-exit-if-not-interactive
-        }
-        speed=1000 out "### The assertion holds."
+                case "$interactive_verify" in
+                    "always"|"1"|"true"|"fail"*) interactive;;
+                    *) command-exit-if-not-interactive;;
+                esac
+        else
+            speed=1000 out "### The assertion holds."
+                case "$interactive_verify" in
+                    "always"|"1"|"true") interactive;;
+                esac
+        fi
     done
 }
 
@@ -515,6 +527,7 @@ vm=${vm-"crirm-test-numa"}
 cri_resmgr_cfg=${cri_resmgr_cfg-"${SCRIPT_DIR}/cri-resmgr-memtier.cfg"}
 cleanup=${cleanup-0}
 reinstall_cri_resmgr=${reinstall_cri_resmgr-0}
+interactive_verify=${interactive_verify-never}
 topology=${topology-'[
     {"mem": "1G", "cores": 1, "nodes": 2, "packages": 2, "node-dist": {"4": 28, "5": 28}},
     {"nvmem": "8G", "node-dist": {"5": 28, "0": 17}},
```

klihub: And while we are at it, IMO we better make it so that `create()` can also automatically drop into interactive upon failures.

askervin: `on_verify`, `on_verify_fail`, `on_create`, `on_create_fail` hooks in place.

test/numa/tests/memtier/n4c16/test01-only-add-pods/code.var.sh
klihub: Well, there is a deliberate attempt to write the scoring algorithm so that it never evaluates the score of two alternatives as equal. Everything else being equal, the CPU with the lower id wins. This was mostly put in with testing in mind, as it makes allocation runs reproducible with predictable results. Since this is deliberate behavior, I think we should have at least some cases verifying it.

test/numa/tests/memtier/n4c16/topology.var.json
klihub: True, we should prevent them from mushrooming too much.

test/numa/tests/memtier/n4c16/test01-only-add-pods/code.var.sh
klihub: We can also handle this with further PRs after this one is merged. It's been pending long enough and start to be really good for testing.

askervin: Ok, let's take further CPU id verifications later, I added a couple now.

askervin:
New directory structure as discussed off-line: this will execute policy tests:
```
cri-resource-manager/test/e2e$ ./run_tests.sh policies
```

Later on we can change `run_tests.sh` to be more flexible with the directory structure, and then add other subdirectories like `controllers` etc.

klihub (approved):
LGTM. This is really good work ! Let's merge this and handle anything else that comes up with further PRs!

marquiz:
lgtm

PR #391, opened 2020-09-10
demo/lib: add basic environment checks for govm.

klihub:


marquiz (approved):
lgtm

askervin (changes requested):
govm uses qemu from its docker image. It does not depend on qemu installed on system.

klihub:
> govm uses qemu from its docker image. It does not depend on qemu installed on system.

Ah, good point. I'll update.

askervin (approved):
LGTM

PR #393, opened 2020-09-10
e2e: Add vm-set-kernel-cmdline and vm-reboot

askervin:


demo/lib/vm.bash
```diff
@@ -62,6 +62,34 @@ vm-wait-process() {
     fi
 }
 
+vm-set-kernel-cmdline() { # script API
```
klihub, line 65: Looking good, but let's also have an append version of this. Usually it is safer than having to know, repeat and keep in sync the existing command line. Especially if just you want to enable something which is usually off, for instance something like CPU isolation, which is our primary target ATM for this.

klihub (approved):
LGTM.

PR #394, opened 2020-09-10
e2e tests: check that local/VM cri-resmgr binaries are the same.

klihub:


klihub:
Closing, #397 is doing the same.

PR #395, opened 2020-09-10
memtier: dump scores by node name.

klihub:


marquiz (approved):
Thanks @klihub, looks good!

PR #396, opened 2020-09-10
Makefile: run e2e test for release-tests.

klihub:


Makefile
```diff
@@ -358,8 +361,26 @@ else
             exit $$rc
 endif
 
-release-tests:
-	$(Q): # An empty placeholder
+release-tests: e2e-tests
+
+e2e-tests: build
+ifndef WHAT
+	$(Q)what=test/e2e/policies; \
```
marquiz, line 368: Hmm, instead of `ifndef` could you use something like this:
```
$(Q)what=$(if $(WHAT),$(WHAT),"test/e2e/policies"); \
```
so that you wouldn't need to duplicate the rest of the script

klihub: I'll change it like that. Much better.

marquiz (approved):
Looks Good To Me

askervin (approved):
LGTM

PR #397, opened 2020-09-11
e2e: add uninstall, install and launch to script API

askervin:
- launch is needed after vm-reboot, for instance after kernel command
  line change.
- uninstall and install help resetting and switching cri-resmgr
  version, for instance in interactive mode if running outdated version
  of cri-resmgr.


askervin:
This patch aims to cover PR #394.

PR #398, opened 2020-09-11
test/e2e: fix: can't exhaust shared subset of CPUs.

klihub:
Currently memtier refuses to exhaust even idle shared CPU subsets of
a pool. Therefore such attempts will try to squeeze the container to
another pool at the same level or, if none found, push the container
one level up to the parent pool.


askervin (approved):
LGTM.

Test passes in current master.

PR #403, opened 2020-09-16
e2e/demo: run.sh passes through USE_NET_BRIDGES to govm

askervin:
With USE_NET_BRIDGES=1 govm launches Qemu with bridge configuration
instead of the default macvlan. The bridge configuration works in
github actions, while the default does not.

USE_NET_BRIDGES must be 0 by default until govm PR https://github.com/govm-project/govm/pull/44 is merged.
(Update: it is merged. However, we should make sure that govm/govm:latest image gets updated too, because using USE_NET_BRIDGES=1 with an outdated image will refuse to launch Qemu from the image.)


klihub (approved):
LGTM

PR #405, opened 2020-09-17
Containerized build of documentation

marquiz:
Add makefile rules for building and serving the html documentation inside a
container.
- Running `make site-build` will build the html content unders docs.
- Running `make site-serve` builds the site content and serves it at
http://localhost:8081/

Also use containerized build in github workflows – use the same workflow as developers are likely to use. Verifies that
containerized build of documentation works.

Drops all other files than the generated documentation in `gh-pages` branch in order to make
sure that the documentation is self-contained and does not depend on the
source tree.

The updated workflow can be verified here:
https://marquiz.github.io/cri-resource-manager

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -11,11 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: Install dependencies
-      run: |
-        export PATH="$HOME/.local/bin:$PATH"
-        sudo apt-get install -y python3-setuptools
-        pip3 install --user -r docs/requirements.txt
+    - name: Build docs builder image
+      run: make site-build
     - name: Branch off base for updated gh-pages
       run: |
         git branch gh-pages
@@ -23,7 +20,6 @@ jobs:
     - name: Add latest generated docs to gh-pages
       shell: bash
       run: |
-        PATH="$HOME/.local/bin:$PATH" make html
         cp -rv _build/html/* .
         for f in $(find _build/html -type f -name '.*'); do \
             cp -v $f ./${f#_build/html/}; \
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -20,22 +20,13 @@ jobs:
     - name: Add latest generated docs to gh-pages
       shell: bash
       run: |
-        cp -rv _build/html/* .
-        for f in $(find _build/html -type f -name '.*'); do \
-            cp -v $f ./${f#_build/html/}; \
-        done
-        for src in $(find _build/html -type f); do \
-            dst="${src#_build/html/}"; \
-            echo "adding generated $dst to git repo..."; \
-            git add ./$dst; \
-        done
-        touch .nojekyll
-        git add .nojekyll
+        touch _build/html/.nojekyll
+        git --work-tree _build/html add --no-all -v .
         author=$(git log --merges --pretty=format:%an -n 1)
         email=$(git log --merges --pretty=format:%ae -n 1)
         git config --global user.name "$author"
         git config --global user.email "$email"
-        git commit -m "publish-docs.yml: generated latest html output." .
+        git commit -m "publish-docs.yml: generated latest html output."
     - name: Publish/force-push to gh-pages
       shell: bash
       env:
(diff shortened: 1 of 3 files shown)


klihub:
> Drops all other files than the generated documentation in `gh-pages` branch in order to make
> sure that the documentation is self-contained and does not depend on the
> source tree.

Personally I am not very fond of this particular idea/change. This is how it originally worked when we
got it from the docs team. It was then deliberately changed to make the generated documentation a
single commit on top of the tree from which it was generated, to make it absolutely obvious what was
the source/version the documentation came from.

I would understand nuking the sources from the `gh-pages` branch and go all in for such a change, if
we modified this to generate multiple versions of the documentation: one for master and one for each
tagged release which already included support for generating documentation.

klihub (approved):
The containerized build LGTM. I *can also live* with the doc-only commit to `gh-pages`.
See my other review comment about what would IMO make it more justifiable, and
turn it into a definite net positive change.

marquiz:
> Personally I am not very fond of this particular idea/change. This is how it originally worked when we
> got it from the docs team. It was then deliberately changed to make the generated documentation a
> single commit on top of the tree from which it was generated, to make it absolutely obvious what was
> the source/version the documentation came from.

I understand that point of view. What I most dislike in the current setup is totally "polluted tree", mxing the site content with our source tree content. Actually, I would not even keep the master branch history as part of `gh-pages` but have it as an orphan branch with just the site content and put the source branch and sha1 (from where the site was generated) just in the commit message. I think, this is obvious enough information about the source and just about as reliable as we have for the cri-resmgr version (and the container images) itself.

I don't feel very strongly about this, though, especially at this early phase of the project documentation. So I'll change it back how it was.


> I would understand nuking the sources from the `gh-pages` branch and go all in for such a change, if
> we modified this to generate multiple versions of the documentation: one for master and one for each
> tagged release which already included support for generating documentation. WDYT ? It should be
> fairly easy to add to this.

Yes, we definitely need this. But I'll add it later. I think we need to discuss the details a bit, how to really maintain the gh-pages branch. Do we really want to always (try to) re-generate documentation for all versions or more intelligently maintain the git history.

marquiz:
> The containerized build LGTM. I _can also live_ with the doc-only commit to `gh-pages`.
> See my other review comment about what would IMO make it more justifiable, and
> turn it into a definite net positive change.

I changed it back how it was. Now the source tree is preserved in the gh-pages branch. LGTM was lost, though 😄

klihub:
> > Personally I am not very fond of this particular idea/change. This is how it originally worked when we
> > got it from the docs team. It was then deliberately changed to make the generated documentation a
> > single commit on top of the tree from which it was generated, to make it absolutely obvious what was
> > the source/version the documentation came from.
> 
> I understand that point of view. What I most dislike in the current setup is totally "polluted tree", mxing the site content with our source tree content.

I also dislike that. But a more appropriate fix for that IMO would be to
  1. get rid of that `cp -rv _build/html/* .` + the related `git add/commit`
  2. instead do a `mkdir gh-pages; mv _build/html/* gh-pages` + `git add gh-pages; git commit -m xyz gh-pages`
  3. change /index.html to `<meta http-equiv="refresh" content="0; URL='gh-pages/README.html'" />`
  4. do any fixup for links resulting from the generated stuff now being located in `/gh-pages` instead of `/`

> Actually, I would not even keep the master branch history as part of `gh-pages` but have it as an orphan branch with just the site content and put the source branch and sha1 (from where the site was generated) just in the commit message. I think, this is obvious enough information about the source and just about as reliable as we have for the cri-resmgr version (and the container images) itself.

If we have multiple versions of the docs, I agree that it's easy to justify to nuke the sources.

> 
> I don't feel very strongly about this, though, especially at this early phase of the project documentation. So I'll change it back how it was.

Well, that's fine, too but it's not necessary if/since we agree where we should be heading. Nuking the sources might even increase our motivation to take the next steps for hosting multiple versions.

> 
> > I would understand nuking the sources from the `gh-pages` branch and go all in for such a change, if
> > we modified this to generate multiple versions of the documentation: one for master and one for each
> > tagged release which already included support for generating documentation. WDYT ? It should be
> > fairly easy to add to this.
> 
> Yes, we definitely need this. But I'll add it later.

Ok, that's perfectly fine. Then this is good for the time being.

> I think we need to discuss the details a bit, how to really maintain the gh-pages branch. Do we really want to always (try to) re-generate documentation for all versions or more intelligently maintain the git history.

Sure, let's have some head-scratching first then a chat about it. Here is what I currently think we should roughly do:

  1. discover tagged releases that we haven't generated documentation for
  2. generate documentation for those releases, each as a commit of its own (sorted/ordered intuitively using the release tag)
  3. rebase the commit history moving the commit generating documentation for master/HEAD to the tip
  4. git reset to HEAD^ (IOW, remove the commit for generating the rolling master documentation)
  5. generate latest documentation for master as the tip commit
  6. (could be also as part of 5) regenerate index to access all versions of the documentation

Additionally for each version in step 2, and for master in 5 above I would record the exact `tag` and the full
`SHA1` from/for the documentation was generated from, and commit those records to the `gh-pages`
branch. Otherwise eventually someone (most probably one of us) will spend a lot of time first chasing ghosts
and then chasing the guilty one(s, again most probably one/some of us)...

klihub (approved):
LGTM. Both this way and with the sources nuked.

PR #406, opened 2020-09-17
memtier: fix eligible but unsatisfied isolated allocation, add basic placement tests.

klihub:
Do not compare negative isolated capacities. In particular,
never consider equal negative capacities a tie. Do the same
for negative shared capacity. It's only for code symmetry, the
existing filtering of pools with insufficient resources should
already protect us from ever hitting that problem.

Add a minimal test set for verifying basic `memtier` container
placement.


kad (approved):
lgtm

marquiz (approved):
lgtm

PR #409, opened 2020-09-18
e2e: write files under vm-files to vm

askervin:
- run_tests.sh looks for vm-files directories under policy,
  topology and test directories and constructs vm_files
  variable recursively.
- run.sh writes files in vm_files associative array to vm.
- Files are written to vm before running/restarting cri-resmgr.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -68,6 +68,15 @@ usage() {
     echo "    cri_resmgr_cfg: configuration file forced to cri-resmgr."
     echo "    cri_resmgr_extra_args: arguments to be added on cri-resmgr"
     echo "             command line when launched"
+    echo "    vm_files: \"serialized\" associative array of files to be created on vm"
+    echo "             associative array syntax:"
+    echo "             vm_files['/path/file']=file:/path/on/host"
+    echo "                                   ='data:,plain text content'"
+    echo "                                   =data:;base64,ZGF0YQ=="
+    echo "                                   =dir: (creates only /path/file directory)"
+    echo "             vm_files['/etc/motd']='data:,hello world'"
+    echo "             How to execute run.sh with serialized array:"
+    echo "             vm_files=\$(declare -p vm_files) ./run.sh"
     echo "    code:    Variable that contains test script code to be run"
     echo "             if SCRIPT is not given."
     echo ""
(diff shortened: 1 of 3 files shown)


askervin:
This is one option for defining files to be written on virtual machines before launching tests. First I thought this would be pretty clean way to do it, but then it turned out that associative arrays are not quite first-class citizens in bash - so the result might not be that nice looking after all.

Anyway, the idea is that one can have `vm-files` directories on policy, topology and test levels. Contents of those directories are written to vm.
- Latter overrides former in policy-topology-test order.
- New files and directories will be created. Example: `vm-files/etc/cmk/pools/file.cfg` creates `/etc/cmk/pools` directory and writes `file.cfg` with the same content there.
- Content of existing files is changed. Example: `vm-files/proc/sys/vm/swappiness` writes new content to `/proc/sys/vm/swappiness`.
- Obviously there is no general clean up for this other than destroying the vm. When not used with care, this may produce unwanted side effects.

askervin:
This PR needs more beautification... serialized associative map as a parameter sounds pretty bad.

askervin:
...but also unavoidable. So I suppose we might need to accept it or take completely different route. Reopening for comments.

askervin:
A commit illustrating this kind of `vm-files` usage:

https://github.com/askervin/cri-resource-manager/commit/b6feffcd8bc4f18d785e75b437662d23af4acaf6

test/e2e/run.sh
```diff
@@ -290,6 +299,48 @@ run-hook() {
...
+            data=$(base64 -d <<< "${src#data:;base64,}")
+        elif [[ $src == "file:"* ]]; then
+            data=$(< "${src#file:}")
+        elif [[ "$src" == "dir:" ]]; then
+            echo -n "Creating on vm: $tgt/... "
+            vm-command-q "mkdir -p \"$tgt\"" || {
+                error "failed to make directory to vm \"$tgt\""
+            }
+            echo "ok."
+            continue
```
klihub, line 332: nitpick/question: why not to do this with a case/esac ?

test/e2e/run_tests.sh
```diff
@@ -93,26 +128,35 @@ for POLICY_DIR in "$TESTS_ROOT_DIR"/*; do
...
+                continue
+            fi
             (
                 export_var_files "$TOPOLOGY_DIR"
+                export_vm_files "$TOPOLOGY_DIR/vm-files"
                 vm="$(basename "$TOPOLOGY_DIR")"
                 export vm
                 for TEST_DIR in "$TOPOLOGY_DIR"/*; do
-                    if ! [ -d "$TEST_DIR" ]; then
+                    if ! [ -d "$TEST_DIR" ] ; then
```
klihub, line 148: nitpick: extra space slipped through/in

test/e2e/run.sh
```diff
@@ -290,6 +299,48 @@ run-hook() {
     eval "${hook_code}"
 }
 
+install-files() {
```
klihub, line 302: nitpick: wouldn't it be more consistent to call this `install{-_}vm{-_}files`, considering how the related/similar `export_vm_files()` and `export_var_files()`  functions and the `vm-files` variable itself called ?

bikeshed: We seem to have two naming conventions here for multipart function names. One is using underscores (`export_vm_files`), and another using hyphens (`install-files`). Is there a particular logic to this or is it related to something (eg. temporal correlation with the demo->e2e-tests evolution timeline) ?

klihub (commented):
Generally LGTM.
I had a few minor nitpicks and some general bikeshedding questions.

test/e2e/run_tests.sh
askervin: uh-oh. need to improve "git diff" self-review before pushing...
Thanks!

test/e2e/run.sh
askervin: trying to keep using `-` wherever possible to keep function naming consistent. Renamed `export_{vm,var}_files` to `export-{vm,var}-files`.

klihub (approved):
LGTM.

PR #414, opened 2020-09-23
packaging: install sysconf collateral to $(DEFAULTDIR).

klihub:
Use `$(DEFAULTDIR)` to correctly pick distro-specific location for defaults (`/etc/sysconfig` vs. `/etc/default`).
This should fill incorrect default location used and referenced in debian/ubuntu packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,12 +263,12 @@ install-systemd-%:
 
 install-sysconf-%:
 	$(Q)bin=$(patsubst install-sysconf-%,%,$@); dir=cmd/$$bin; \
-	echo "Installing sysconf collateral for $$bin..."; \
-	$(INSTALL) -d $(DESTDIR)$(SYSCONFDIR)/sysconfig && \
+	echo "Installing sysconf/default collateral for $$bin..."; \
+	$(INSTALL) -d $(DESTDIR)$(DEFAULTDIR) && \
 	for f in $$(find $$dir -name \*.sysconf); do \
-	    echo "  $$f in $(DESTDIR)$(SYSCONFDIR)/sysconfig..."; \
+	    echo "  $$f in $(DESTDIR)$(DEFAULTDIR)..."; \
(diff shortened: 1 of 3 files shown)


kad:
There is also fix needed for name of cfg
```
-rw-r--r-- root/root      1396 2020-09-22 19:42 ./etc/cri-resmgr/fallback.cfg.sample
-rw-r--r-- root/root       342 2020-09-22 19:42 ./etc/sysconfig/cri-resource-manager
```

klihub:
> There is also fix needed for name of cfg
> 
> ```
> -rw-r--r-- root/root      1396 2020-09-22 19:42 ./etc/cri-resmgr/fallback.cfg.sample
> -rw-r--r-- root/root       342 2020-09-22 19:42 ./etc/sysconfig/cri-resource-manager
> ```

Do you mean that the sample should be installed as real config file so that after package installation
the whole shebang starts up automatically ? It was deliberately not done like that. I thought that was
what we agreed to do so that it does not start without the user at least renaming the config file which
is a perfect opportunity to also take a look at its content. If my understanding is incorrect and that is
not what we want then commit 6682d24d861f120ba41181a630a2db84bc376e58 simply needs to be
reverted.

askervin (commented):
deb packages were easy to build and install, and when trying to launch cri-resmgr with systemctl, journalctl logs showed clearly that /etc/cri-resmgr/fallback.cfg could not be found. Visiting that directory immediately revealed that the sample is there and I can just rename it:
```
/etc/cri-resmgr/fallback.cfg.sample
```

So in the perspective of a user who do not read install docs, this was pretty nice. However, now INSTALL.md says that the sample should be copied from `/usr/share/doc/cri-resource-manager/fallback.cfg.sample`, but it's not there in the `deb` packages, so maybe the doc should be revisited still.

askervin (approved):
LGTM

PR #415, opened 2020-09-24
e2e: test infra and the first test for packaging (debian-10)

askervin:
This patch adds a test that verifies that cri-resmgr installed from `deb` packages and managed by `systemd` works fine as part of our e2e test cluster.

This test passes now that PR #414 is merged, but failed before that.

New Makefile target (`packaging-tests`) fails for me before actually running the test due to some error in Ubuntu build. But the test itself passes for debian-10 packages - installed on e2e ubuntu 20.04 vm.


Makefile
```diff
@@ -378,6 +378,9 @@ e2e-tests: build
 	    exit 1; \
 	fi
 
+packaging-tests: cross-deb
+	$(E2E_RUN) test/e2e/packaging/cross-deb
```
klihub, line 382: Shouldn't this be `packages/debian-10` ?

klihub: > Shouldn't this be `packages/debian-10` ?

I mean `$(E2E_RUN) test/e2e/packages/debian-10` to be exact ?

askervin: Definitely! Thanks! (Dangers of last minute renaming...)

askervin: Fixed.

test/e2e/packages/debian-10/n4c16/test01-systemd/code.var.sh
```diff
@@ -0,0 +1,10 @@
+# Clear cri-resmgr output from previous runs.
+vm-command "journalctl --vacuum-time=1s"
+
+# Create a pod.
+create besteffort
```
klihub, line 5: Maybe also splitting out pulling the image to avoid `kubectl wait` timing out due to slow network giving a false failure. With something like this:
```
...
image=$(cat <path-to->besteffort.yaml | egrep '^ *image: .*$' | sed -E 's/^ *image: *([^ ]*)$/\1/g')
vm-command crictl -i unix:///var/run/cri-resmgr/cri-resmgr.sock pull $image
create besteffort
...
```
On my first try I hit exactly such a  `kubectl wait` timeout for the busybox image.

klihub: Or to rephrase, considering the timeout snafu I hit, wouldn't it make sense to add a test API function for `pull <yaml-template-name>` which would then do that above for all images found in the template ?

```
for img in $(cat <path-to-yaml-template> | egrep '^ *image: .*$' | sed -E 's/^ *image: *([^ ]*)$/\1/g'); do
    echo "Pre-fetching image $img for <path-to-yaml-template>..." 
    vm-command crictl -i unix:///var/run/cri-resmgr/cri-resmgr.sock pull $img
done
```

askervin: We're using busybox image in all tests, so I'd be tempted to add

```
crictl -i unix:///var/run/cri-resmgr/cri-resmgr.sock pull $image
```

as part of the test setup in infra. What do you think?

klihub: I think that should be good enough for now. We can make it more flexible later if/when ever needed.

klihub (commented):
Another thing: I think for this test we should now spice up the e2e test framework so it can create also create images of other distros. I don't think generally it is a good idea to test debian packages by installing them to ubuntu, regardless of how close those are. Anyway we'll need the corresponding functionality for fedora, centos, and opensuse to test the rest. And also for testing centos+{containerd,crio} combos with e2e tests.

But not now. That's for another PR.

test/e2e/run.sh
```diff
@@ -750,7 +783,7 @@ if [ "$reinstall_cri_resmgr" == "1" ]; then
     uninstall cri-resmgr
 fi
 
-if ! vm-command-q "[ -f /usr/local/bin/cri-resmgr ]"; then
+if ! vm-command-q "[ -f /usr/local/bin/cri-resmgr ] || [ -f /usr/bin/cri-resmgr ]"; then
```
askervin, line 786: This ! and || looks buggy.

askervin (commented):
<snip>

demo/lib/vm.bash
```diff
@@ -161,7 +161,23 @@ vm-install-cri-resmgr() {
         vm-command "go get -d -v github.com/intel/cri-resource-manager"
         CRI_RESMGR_SOURCE_DIR=$(awk '/package.*cri-resource-manager/{print $NF}' <<< "$COMMAND_OUTPUT")
         vm-command "cd $CRI_RESMGR_SOURCE_DIR && make install && cd -"
-    else
+    elif [ "$binsrc" == "packages/debian-10" ]; then
```
klihub, line 164: I'd suggest already now using something like `[  "${binsrc#packaging/}" != "$binsrc" ]` as a catch-all case for all current and future packaging tests.

test/e2e/run.sh
```diff
@@ -135,7 +136,11 @@ screen-install-cri-resmgr-debugging() {
 
 screen-launch-cri-resmgr() {
     speed=60 out "### Launching cri-resmgr with config $cri_resmgr_cfg."
-    launch cri-resmgr
+    if [ "$binsrc" == "packages/debian-10" ]; then
```
klihub, line 139: Same comment here as earlier, I'd rather use a catch-all case for packaging tests:
`[ "${binsrc#packaging/}" != "$binsrc" ]`...

test/e2e/run.sh
klihub, line 786: Wouldn't a `! vm-command-q "[ type -p cri-resmgr ]"` test be better here ?

test/e2e/packages/debian-10/n4c16/test01-systemd/code.var.sh
askervin: Ended up embedding this into `create()` function. It pulls all (not-pulled-earlier) images it finds in instantiated templates before `kubectl create`.

demo/lib/vm.bash
askervin: Changed this to matching "package/{ubuntu,debian}"*, because the branch is written for debs only.

test/e2e/run.sh
askervin: Changed this to matching all package installations. It'll work as long as rpm's and deb's all install systemd service.

test/e2e/run.sh
askervin: Yes it would! Fixed.

askervin:
When adding prefetch (pull) to `create()`, I noticed that it might be nice to refactor fail hook execution. Maybe `command-error` could take a hook variable as parameter and run it in case it is set. It would simplify `create()`. But I think it's better done in another PR.

klihub (approved):
LGTM. I think this is a good basis to build more packaging tests upon.

klihub:
@marquiz Are you ok with merging this as an initial baseline ? IIUC, @askervin has some ongoing static-pools e2e test work that build on this. I have an RFC branch that tries to add multi-distro testing capability, based on an earlier version of this.

marquiz (approved):
I think we should merge this.

The only think that sticks into my eye is `debian-10` – I'm more used to the code names like `debian-buster` 😄

PR #420, opened 2020-09-29
docs: add initial architectural description.

klihub:
[Here](https://github.com/klihub/cri-resource-manager/blob/documentation/arhitecture/docs/architecture.md) is the first attempt to put something down about the architecture, both external, about how CRI-RM fits into the set of cluster/node components, and internal, about how CRI-RM is split up into its internal components.

I think we need one more figure about CRI-RM/CRI-RM agent/Control Plane, and maybe one about a rough signlling flow from kubelet through CRI-RM to CRI runtime and back. What do you guys think ?


docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+### Generic Resource Controller Layer
+
+The generic resource controller layer defines the abstract interface the rest
+of CRI-RM uses to interact with resource controller implementations and takes
+care of the details of dispatching calls to the controller implementations for
+post-policy enforcment of decisions.
+
+
+### Metrics Collector
+
```
marquiz, line 168: Some text here 😉  Should we mention tracing, too?

marquiz (commented):
Thanks @klihub for this! Looks good and definitely extensive enough for the time being. I just had a few minor comments.

If we just get the architecture diagram I think we could remove the RFC label and merge this.

I'm not sure if we need the policy descriptions here (just thinking about the maintenance of the documentation). We could just link to the documentation of individual policies. Same goes with the resource controllers.

docs/architecture.md
klihub: Sure. It's just that metrics collection is still a bit in flux/WiP. We've been a little too busy with other stuff to really push it to a level where it'd definitely start taking a final shape.

klihub: Yes, tracing is missing, as well as the introspection+visualisation.

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+CRI Resource Manager (CRI-RM) controls how much and which resources are
+assigned to containers on a node in a Kubernetes cluster. In practice,
+this control is implemented by intercepting the standard CRI communication
+protocol which is in use between kubelet, the Kubernetes node agent, and
+the CRI container runtime implementation. Intercepted CRI messages are
+potentially modified before being forwarded according to decisions made by
+a resource allocation policy algorithm local within CRI-RM. Intercepting
+the CRI protocol in turn is achieved by inserting CRI-RM as a proxy between
+the kubelet and the real CRI runtime.
```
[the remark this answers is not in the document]
klihub: Yeah... I've had that described but apparently it fell victim to a rebase/squash cycle. I'll readd it.

klihub:
> Thanks @klihub for this! Looks good and definitely extensive enough for the time being. I just had a few minor comments.
> 
> If we just get the architecture diagram I think we could remove the RFC label and merge this.
> 
> I'm not sure if we need the policy descriptions here (just thinking about the maintenance of the documentation). We could just link to the documentation of individual policies. Same goes with the resource controllers.

I agree that for most of these things we should link to more detailed descriptions elsewhere in the docs. My goal with the architectural description was twofold. The primary one is to provide an overview which is both compact and complete enough for a single read-through. If the reader is forced to bunny hop through a number of links to get a rough understanding then it is usually neither of those. The secondary is, with the correct links (and more detailed docs) in place, provide a starting point /outline for a top-down deep dive into the whole thing. Does this make sense at all ?

With that in mind, I agree that probably we should only list/link to the policies here, and describe only what a policy/the generic policy layer is supposed to do.

docs/architecture.md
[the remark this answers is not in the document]
klihub: Okay, I'll add that. It should make it clearer.

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
+# Architecture
+
+## Introduction
+
+CRI Resource Manager (CRI-RM) controls how much and which resources are
+assigned to containers on a node in a Kubernetes cluster. In practice,
```
askervin, line 6: In addition to Markus's point, I'd like to bring up quite early (maybe already here) also the fact that kubelet's CRI messaging is not the only event source. CRI-RM reacts to cluster configuration changes (cri-rm config, adjustments) and triggers actions completely independently (rebalance).

askervin (commented):
(I'll continue review a bit later...)

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+A topology-aware policy capable of handling multiple tiers/types of memory,
+typically a DRAM/PMEM combination configured in 2-layer memory mode.
+
+### Resource Controller Implementations
+
+#### RDT
+
+A resource controller implementation responsible for enforcing assignments
+of controllers to last level cache and memory bandwidth allocation classes,
+using the resctrl pseudo filesystem of the Linux kernel.
```
askervin, line 208: Could we simplify wording here... Something like

> RDT controls containers' last level cache allocations and memory bandwith using resctrl pseudo filesystem on Linux.

Then maybe a separate sentence or two on how input is structured and how it maps these resources to pods and containers.

askervin (commented):
Finished review with the above suggestions. This is good doc.

docs/architecture.md
klihub: I added a paragraph about configuration/configurability, but I think the part about how it is in practice an event source because it triggers a reconfiguration/rebalancing should only be part of the later sections. I think it'd be TMI without the necessary context to be properly understood at this point in the text.

docs/architecture.md
klihub: I tried to reformulate the descriptions of controllers, trying to avoid too long sentences.

docs/architecture.md
klihub: Wrote a paragraph about metrics collection.

marquiz:
Nice work! I think `cri-resmgr.png` should be embedded somewhere in `architecture.md`. Otherwise lgtm

klihub:
> Nice work! I think `cri-resmgr.png` should be embedded somewhere in `architecture.md`. Otherwise lgtm

Should we use that, or should I redraw it by hand using the reMarkable, similar to get something similar to the sketch we have now in place after the overview section ?

askervin (approved):
LGTM!

Maybe in the architecture diagram kubelet could be on the top and container runtime at the bottom, but I guess the audience can read it as well from left to right. :)

Anyway, I'd take this in now and update later when necessary.

klihub:
> LGTM!
> 
> Maybe in the architecture diagram kubelet could be on the top and container runtime at the bottom, but I guess the audience can read it as well from left to right. :)

Hmm... How about [this one instead](https://github.com/klihub/cri-resource-manager/tree/documentation/preview/docs/architecture.md#Overview), better or worse ?

askervin:
> > LGTM!
> > Maybe in the architecture diagram kubelet could be on the top and container runtime at the bottom, but I guess the audience can read it as well from left to right. :)
> 
> Hmm... How about [this one instead](https://github.com/klihub/cri-resource-manager/tree/documentation/preview/docs/architecture.md#Overview), better or worse ?

Dude! I think that's great!

(Almost tempted to say "nay" and start collecting CRI-RM architecture art gallery. ;D)
Maybe we'll anyway see more later on, when visualizing the agent and possibly the webhook. But that's a different story and a different figure.

marquiz:
Looks great! 👍 

Just one thing: what about the PNG? I think it would be good to show it, too.

klihub:
> Looks great! 
> 
> Just one thing: what about the PNG? I think it would be good to show it, too. If it's not "visible" it just get lost and forgotten and outdated...

Fair enough. I added it after the Node Agent section, since I don't see where else it could fit in.

marquiz (approved):
Looks great!

lgtm

kad (dismissed):
lgtm

marquiz (changes requested):
I'll take back my lgtm 😄 

There is currently a `docs/developer-guide/architecture.md`. Either we should put this new doc there or drop the other one. I don't know how relevant the architecture is for a "casual" user(?) Maybe it is and we should have the architecture doc on the main level as you did. WDYT?

klihub:
> I'll take back my lgtm 
> 
> There is currently a `docs/developer-guide/architecture.md`. Either we should put this new doc there or drop the other one. I don't know how relevant the architecture is for a "casual" user(?) Maybe it is and we should have the architecture doc on the main level as you did. WDYT?

Yes, it belongs to docs/developer-guide. I'll move it there.

marquiz (approved):
Cool!
lgtm

PR #422, opened 2020-09-30
e2e: basic plumbing for multi-distro testing, packaging tests for centos-8, ubuntu-{18,20}.04.

klihub:
This patch series
  - adds a generic interface to hide the distro-specific bits of bootstrapping test VMs
  - switches `demo/lib/*.bash` to use these generic `distro-*()` functions
  - hacks `test/e2e/run*.sh` to take new external `distro` and `cri` parameters which select the test VM distro and CRI runtime
  - updates VM naming scheme to `$topology-$distro-$cri`
  - adds top-level stub for `crio` in addition to `containerd` as a supported test CRI runtime
  - adds new packaging tests for `centos-7` (broken), `centos-8`, `ubuntu-18.04`, `ubuntu-20.04`

Bootstrapping a `centos-7`-based test VM will require additional extra calisthenics for `containerd`. The default
`overlayfs` `snapshotter` does not work on top of `xfs`, which is the filesystem of choice in the `centos-7` cloud
images. Probaly the easiest will be to go with `devmapper` and a thin pool with `loopback`-mounted file as the
block device backend.


demo/lib/vm.bash
```diff
@@ -143,25 +195,46 @@ vm-reboot() { # script API
 
 vm-networking() {
     vm-command-q "grep -q 1 /proc/sys/net/ipv4/ip_forward" || vm-command "sysctl -w net.ipv4.ip_forward=1"
-    vm-command-q "grep -q ^net.ipv4.ip_forward=1 /etc/sysctl.conf" || vm-command "sed -i 's/#net.ipv4.ip_forward=1/net.ipv4.ip_forward=1/g' /etc/sysctl.conf"
+    vm-command-q "grep -q ^net.ipv4.ip_forward=1 /etc/sysctl.conf" || vm-command "echo net.ipv4.ip_forward=1 > /etc/sysctl.conf"
```
askervin, line 198: feeling a bit uneasy on replacing sysctl.conf contents with this, yet it probably works for us in clean VMs... perhaps we could

`vm-sed-file /etc/sysctl.conf '/net.ipv4.ip_forward/d'`
and then append (`>> /etc/sysctl.conf`) here.

test/e2e/packages/ubuntu-20.04/n4c16/test01-systemd/code.var.sh
```diff
@@ -0,0 +1,10 @@
+# Clear cri-resmgr output from previous runs.
```
askervin, line 1: This starts to look like we could have `code.var.sh` on `packages` level and add `export_var_files "$TESTS_ROOT_DIR"` to `run_tests.sh`...

askervin (commented):
This will be a huge leap forward, great work - and a lot of it.

Related to this, we may need container runtime specific functions, too. Hopefully not too many. But now there's [import image](https://github.com/askervin/cri-resource-manager/blob/58h_cmk_tests/demo/lib/vm.bash#L221) in future tests. Do you have idea should we have this on vm level (like `vm-cri-import-image`) that would go via distro resolver, or should we have `cri.bash` like `distro.bash` that would try to cope with more surprising distro-runtime combos?

demo/lib/distro.bash
```diff
@@ -0,0 +1,339 @@
...
+
+distro-image-url() {
+    local fn=$(distro-resolve ${FUNCNAME#distro-})
+    $fn
+}
+
+distro-ssh-user() {
+    local fn=$(distro-resolve ${FUNCNAME#distro-})
+    $fn
+}
```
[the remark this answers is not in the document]
klihub: > To avoid repeating this code, how about defining all functions in a loop:
> 
> ```
> for func in \
>     distro-image-url \
>     distro-ssh-user \
>    ...
>     ; do eval "
> $func() {
>     local fn=\$(distro-resolve \${FUNCNAME#distro-})
>     \$fn \"\$@\"
> }
> "; done
> ```

I'm not sure about that. I think it would just make it more difficult to read, but that is admittedly a subjective preference.
Instead I've been considering changing `distro-resolve` so that it would
  - use `${FUNCNAME[1]}` (IOW the calling function's name) implicitly as the function to resolv, and
  - resolve and then invoke the resulting function passing on `"$@"` (now the templates are not identical in this sense)

With such changes the distro interface section would become a set of simple trivially copy-pastable one-liners:

```
distro-resolve() {
   local caller="${FUNCNAME[1]}" name="${caller#distro-}" fn
   ...
   # resolve name to fn
   ...
   $fn "$@"
}

distro-image-url()          { distro-resolve "$@"; }
distro-ssh-user()           { distro-resolve "$@"; }
distro-pkg-type()           { distro-resolve "$@"; }
...
distro-set-kernel-cmdline() { distro-resolve "$@"; }
```

WDYT?

demo/lib/distro.bash
```diff
@@ -0,0 +1,339 @@
...
+distro-resolve() {
+    local fn fallback
+    case $VM_DISTRO in
+        ubuntu*) fallback="debian-$1";;
+        centos*) fallback="fedora-$1 rpm-$1";;
+        fedora*) fallback="rpm-$1";;
+        *suse*)  fallback="rpm-$1";;
+    esac
+    # try version-based resolution first, then derivative fallback
+    for fn in "${VM_DISTRO/./_}-$1" "${VM_DISTRO%-*}-$1" $fallback; do
```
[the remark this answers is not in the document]
klihub: You mean to have the resolution algorithm always try that as a completely generic cross-distro fallback as a last resort ? Yes, that would probably make sense. I'll add that.

demo/lib/vm.bash
```diff
@@ -13,8 +11,21 @@ VM_GOVM_COMPOSE_TEMPLATE="vms:
       - KVM_CPU_OPTS=\$(echo "\${KVM_CPU_OPTS}")
       - EXTRA_QEMU_OPTS=\$(echo "\${EXTRA_QEMU_OPTS}")
       - USE_NET_BRIDGES=${USE_NET_BRIDGES-0}
+    user-data: |
+      #!/bin/bash
+      mkdir -p /etc/sudoers.d
+      echo 'Defaults !requiretty' > /etc/sudoers.d/10-norequiretty
```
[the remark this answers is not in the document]
klihub: Yes, definitely.

demo/lib/vm.bash
```diff
@@ -112,6 +123,53 @@ vm-wait-process() {
...
+    # Usage: vm-copy-file PATH-IN-VM
+    #
+    # Reads stdin and writes the content to PATH-IN-VM, creating any
+    # intermediate directories necessary.
+    #
+    # Example:
+    #   echo 'Ahoy, Matey...' | vm-write-file /etc/motd
+    if [ "$1" == "--cleanup" ]; then
+        shift
+        trap "rm -f $1" RETURN EXIT
```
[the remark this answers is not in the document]
klihub: First, I also had an attempt at another variant (with an IMO much nicer approach of) `vm-write-file`, which simply assumed the content would come piped through `stdin` and then was trying to write the content to the specified location on the VM, by doing something like this:
```
vm-write-file() {
    local file="$1"; shift
    case "$file" in
        */*) ;;
        *) file="./$file";;
    esac
    vm-command-q "mkdir -p ${file%/*} && cat > $file" ||
        command-error "failed to write to file $file"
}

# this was then supposed to be used like this (I was a it surprised by the correct syntax)
cat <<<EOF |
a template
file
- with $VARIOUS
- $VARIABLES of all $SORTS
- with $(date) thrown in
EOF
    vm-write-file /path/on/VM
```

I guess you guess what happened. Defeated by that  `<<<"$1"` in `vm-command-q`. For a short while I was considering as a workaround adding a `vm-pipe-input()` which would not use the `sudo bash -l <<<"$1"` idiom. But I quickly gave up the idea because I think sooner than later it would lead down the rabbit hole/rathole of getting locked up in the nuthouse because of quoting insanity.

demo/lib/vm.bash
klihub: > feeling a bit uneasy on replacing sysctl.conf contents with this, yet it probably works for us in clean VMs... perhaps we could
> 
> `vm-sed-file /etc/sysctl.conf '/net.ipv4.ip_forward/d'`
> and then append (`>> /etc/sysctl.conf`) here.

Ouch, well spotted. That was unintentional. It was supposed to be appending to and not nuking the distro-default sysctl.conf altogether. I guess `echo net.ipv4.ip_forward=1 > /etc/sysctl.d/98-ip-forward.conf` would be even better.

test/e2e/packages/ubuntu-20.04/n4c16/test01-systemd/code.var.sh
klihub: Yes, I was already long considering generalizing the current concept and changing `run_tests.sh` so that before each call to `run.sh` it would climb back the directory tree from the test directory to the test root, collect all `.sh` files, and then source and export them in reverse order (IOW top-to-bottom) that would give us basically a kind of overrideable inheritance in the test subtree.

demo/lib/distro.bash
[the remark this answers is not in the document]
askervin: Yeah, that's more friendly for eyes.

I'd probably still pass the function name as the first argument to the resolver, so it wouldn't depend on the name of the calling function.

```
distro-image-url()          { distro-resolve "${FUNCNAME}" "$@"; }
distro-ssh-user()           { distro-resolve "${FUNCNAME}" "$@"; }
...
```

demo/lib/vm.bash
askervin: Yep, that's nice.

test/e2e/packages/ubuntu-20.04/n4c16/test01-systemd/code.var.sh
askervin: Yep, I'll add it. Probably makes sense after getting this patch in.

(I'm not great fan hardcoded number and naming of policy-topology-test levels either. That could use some refactoring at the same time. Now we already know better how the tests tend to be structured.)

Currently we have `*.var.*` files that contain only variable value (without eval/echo/source etc templating). `${FILENAME%.var*}` is the name of the variable. Overrideable inheritance works for these already - except for that exporting vars from the topmost directory is missing.

Would you like to have also `source`able `*.sh` files in the tree?

demo/lib/distro.bash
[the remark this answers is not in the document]
askervin: Great, thanks!

klihub:
> This will be a huge leap forward, great work - and a lot of it.
> 
> Related to this, we may need container runtime specific functions, too. Hopefully not too many. But now there's [import image](https://github.com/askervin/cri-resource-manager/blob/58h_cmk_tests/demo/lib/vm.bash#L221) in future tests. Do you have idea should we have this on vm level (like `vm-cri-import-image`) that would go via distro resolver, or should we have `cri.bash` like `distro.bash` that would try to cope with more surprising distro-runtime combos?

Good question. At this point I would opt for trying to go with the runtime-agnostic
```
crictl -i /var/run/cri-resmgr/cri-resmgr.sock pull <image>
```
and only roll runtime-specific wrappers if there is really a need for it.

demo/lib/distro.bash
```diff
@@ -0,0 +1,322 @@
+GOLANG_VERSION=1.14.8
```
kad, line 1: 1.14.9?

demo/lib/distro.bash
```diff
@@ -0,0 +1,322 @@
...
+}
+
+fedora-install-golang() {
+    fedora-install-pkg wget tar gzip
+    from-tarball-install-golang
+}
+
+fedora-install-containerd() {
+    vm-command-q "[ -f /usr/bin/containerd ]" || {
+        distro-install-repo https://download.docker.com/linux/centos/docker-ce.repo
```
kad, line 237: for fedora, exists https://download.docker.com/linux/fedora/

Also, should we probably use containerd that shipped in distro?

demo/lib/vm.bash
```diff
@@ -88,19 +105,19 @@ vm-command() { # script API
...
     command-start "vm" "$VM_PROMPT" "$1"
     if [ "$2" == "bg" ]; then
-        ( ssh -o StrictHostKeyChecking=No ${VM_SSH_USER}@${VM_IP} sudo bash <<<"$COMMAND" 2>&1 | command-handle-output ;
+        ( ssh -o StrictHostKeyChecking=No ${VM_SSH_USER}@${VM_IP} sudo bash -l <<<"$COMMAND" 2>&1 | command-handle-output ;
           command-end ${PIPESTATUS[0]}
         ) &
         command-runs-in-bg
     else
-        ssh -o StrictHostKeyChecking=No ${VM_SSH_USER}@${VM_IP} sudo bash <<<"$COMMAND" 2>&1 | command-handle-output ;
+        ssh -o StrictHostKeyChecking=No ${VM_SSH_USER}@${VM_IP} sudo bash -l <<<"$COMMAND" 2>&1 | command-handle-output ;
```
kad, line 113: move ssh base args into variable? in case if it will be repeated in multiple places.

kad (commented):
lgtm, with few small comments.

marquiz:
Ach, I didn't find time to look at this in detail :( I'll rely on @kad - lgtm

demo/lib/distro.bash
klihub: Ok, I'll update and see if the shipped one works.

askervin (changes requested):
Suggesting modifying `blockio/run.sh` to keep the demo working:
```
diff --git a/demo/blockio/run.sh b/demo/blockio/run.sh
index 13f2f24..a495e62 100755
--- a/demo/blockio/run.sh
+++ b/demo/blockio/run.sh
@@ -68,7 +68,7 @@ screen-create-vm() {

 screen-install-k8s() {
     speed=60 out "### Installing Kubernetes to the VM."
-    vm-install-containerd
+    vm-install-cri
     vm-install-k8s
 }

@@ -145,7 +145,11 @@ demo-blockio() {

 # Validate parameters
 mode=$1
-vm=${vm-"crirm-demo-blockio"}
+distro=${distro:="ubuntu-20.04"}
+cri=${cri:="containerd"}
+vm=${vm:="blockio-$distro-$cri"}
+echo "vm is here: \"$vm\""
+host-set-vm-config "$vm" "$distro" "$cri"

 if [ "$mode" == "play" ]; then
     speed=${speed-10}
```

Similar thing for `critest/run.sh`:

```
diff --git a/test/critest/run.sh b/test/critest/run.sh
index 98f41c1..caea1c0 100755
--- a/test/critest/run.sh
+++ b/test/critest/run.sh
@@ -56,9 +56,9 @@ screen-create-vm() {
     vm-networking
 }

-screen-install-containerd() {
+screen-install-cri() {
     speed=60 out "### Installing Containerd to the VM."
-    vm-install-containerd
+    vm-install-cri
     vm-install-containernetworking
 }

@@ -119,7 +119,11 @@ require_cmd() {

 # Validate parameters
 mode=$1
-vm=${vm-"crirm-test-critest"}
+distro=${distro:="ubuntu-20.04"}
+cri=${cri:="containerd"}
+vm=${vm:="critest-$distro-$cri"}
+host-set-vm-config "$vm" "$distro" "$cri"
+
 cd "${SCRIPT_DIR}" || error "failed to cd to \"${SCRIPT_DIR}\""
 tests=${tests-*.cfg}

@@ -150,7 +154,7 @@ fi
 screen-copy-cri-resmgr

 if ! vm-command-q "dpkg -l | grep -q containerd"; then
-    screen-install-containerd
+    screen-install-cri
 fi

 if ! vm-command-q "command -v critest | grep -q critest"; then
```

...assuming `host-set-vm-config` would set `VM_SSH_USER`, too. Otherwise `run.sh`'s will need setting it explicitly.

askervin (commented):
The kernel on `e2e/run.sh` new default distro (`ubuntu-18.04`) does not support `cpuXX/topology/die_id`. `demo/lib/topology.py` did not handle missing `die_id` gracefully, which caused our `policy/memtier` e2e tests fail.

I'd vote for using the latest Ubuntu LTS (20.04) as the default distro - only because it is the latest LTS. But anyway I'd like to fix `topology.py` so that it defaults to single-die-packages. I'd prefer merging at least either of these changes as part of this patch, so that `policy/memtier` tests will keep passing before and after the merge.

```
diff --git a/demo/lib/topology.py b/demo/lib/topology.py
index e1787e0..a9f1ba9 100755
--- a/demo/lib/topology.py
+++ b/demo/lib/topology.py
@@ -168,7 +168,7 @@ def dump_to_topology(dump, show_mem=True):
     # "cpu p:0 d:1 n:3 c:2 t:00003000 cpu:13"
     # mem line:
     # "mem n:4: s:8063.83"
-    re_cpu_line = re.compile('cpu p:(?P<package>[0-9]+) d:(?P<die>[0-9]+) n:(?P<node>[0-9]+) c:(?P<core>[0-9]+) t:(?P<thread_siblings>[0-9a-f,]+) cpu:(?P<cpu_id>-9
]+)')
+    re_cpu_line = re.compile('cpu p:(?P<package>[0-9]+) d:(?P<die>[0-9]*) n:(?P<node>[0-9]+) c:(?P<core>[0-9]+) t:(?P<thread_siblings>[0-9a-f,]+) cpu:(?P<cpu_id>-9
]+)')
     re_mem_line = re.compile('mem n:(?P<node>[0-9]+) s:(?P<size>[0-9.]+)')
     re_dist_line = re.compile('dist n:(?P<node>[0-9]+) d:(?P<dist>([0-9 ]+))')
     numeric_cpu_lines = []
@@ -179,7 +179,10 @@ def dump_to_topology(dump, show_mem=True):
         if m:
             mdict = m.groupdict()
             package = int(mdict["package"])
-            die = int(mdict["die"])
+            try:
+                die = int(mdict["die"])
+            except ValueError:
+                die = 0 # handle kernels that do not provide topology/die_id
             node = int(mdict["node"])
             core = int(mdict["core"])
             thread_siblings = eval("0x" + mdict["thread_siblings"].replace(",", ""))
```

Otherwise this is looks great. Let's try to get this in after this change, and improve the rest in smaller pieces later on.

klihub:
> The kernel on `e2e/run.sh` new default distro (`ubuntu-18.04`) does not support `cpuXX/topology/die_id`. `demo/lib/topology.py` did not handle missing `die_id` gracefully, which caused our `policy/memtier` e2e tests fail.
> 
> I'd vote for using the latest Ubuntu LTS (20.04) as the default distro - only because it is the latest LTS.

Updated test/e2e/run*.sh to default to ubuntu-20.04 as the default distro.

askervin:
...still something caught my eye while running the test:

```
vm-put-file()...
    vm-command "touch $2 && cat ${1##*/} >> $2 && rm -f ${1##/*}"
```

Here `rm -f` parameter evaluates empty string. Test output:

```
root@vm> touch /etc/environment && cat tmp.cHfx2dmsBS >> /etc/environment && rm -f
```

klihub:
> ...still something caught my eye while running the test:
> 
> ```
> vm-put-file()...
>     vm-command "touch $2 && cat ${1##*/} >> $2 && rm -f ${1##/*}"
> ```
> 
> Here `rm -f` parameter evaluates empty string. Test output:
> 
> ```
> root@vm> touch /etc/environment && cat tmp.cHfx2dmsBS >> /etc/environment && rm -f
> ```

Thanks! Lysdexia... Fixed.

askervin (approved):
LGTM

PR #425, opened 2020-10-06
critest: fix memtier test and default cleanup

askervin:
- Fix critest on memtier policy, broken due to bad topology.
- Align cleanup behavior after test with e2e tests.
- Fix shellcheck errors.


klihub (approved):
LGTM

PR #427, opened 2020-10-08
docs: apply the webhook only after the webhook service is available

askervin:
If `mutating-webhook-config.yaml` is applied before `webhook-deployment.yaml`, creating and updating pods will fail. That includes creating the `cri-resmgr-webhook` pod behind the webhook service.


klihub (approved):
LGTM

marquiz:
Thanks @askervin for this detail

PR #429, opened 2020-10-09
github: add workflow for verifying docs buildability

marquiz:


klihub (commented):
Wouldn't it be better to make this part of the normal build/test workflow (which is now very misleadingly called go.yml) ? Separating this would now bootstrap yet another virtual machine and clone the repo yet another time just to test-build the docs.

If we do consider a successful build of the docs a requirement for accepting a PR, then I think this logically belongs to the current go.yml, which I'd prefer renaming to something like test-build.yml.

marquiz:
> Wouldn't it be better to make this part of the normal build/test workflow (which is now very misleadingly called go.yml) ? Separating this would now bootstrap yet another virtual machine and clone the repo yet another time just to test-build the docs.

Good point! I just super-quickly hacked this together. Now renamed "Go" to "Verify" and added docs build there.

I think we definitely want to verify the buildability of docs. Otherwise our documentation site gets out-of-sync with the codebase (in case of build errors)

klihub:
> I think we definitely want to verify the buildability of docs. Otherwise our documentation site gets out-of-sync with the codebase (in case of build errors)

Yup, agreed.

.github/workflows/verify.yml
```diff
@@ -44,3 +44,8 @@ jobs:
 
     - name: Codecov report
       run: bash <(curl -s https://codecov.io/bash)
+
+    - name: Build docs
+      run: make site-build
+name: Verify docs build
+on: [pull_request]
```
klihub, line 51: Are these needed here at the end ? Aren't these then (unintentioanlly?) changing the behavior of the workflow to be triggered on just for PRs instead of the originally inteded push or PR (vs. the on:-line  in the beginning of the file) ?

marquiz: Waat have I done?! Copy-paste remnants that really should not be there 

I was just too quick on this one, good that you reviewed it 😌  Fixed now.

kad (approved):
lgtm

PR #433, opened 2020-10-12
RFC: demonstration of some basic ideas for generating multiple versions of the documentation.

klihub:
A small hack to demo building multiple versioned docs. The workflow is not switched over, neither is the result committed to the repo. RFC. You can test this by running
```
make build-docs
make serve-docs
```
Directing your browser at http://127.0.0.1:8082. You should see a real vomit of an index page, listing all the
available documentation versions. If you don't choose any, it will automatically redirect you to the main/master
docs after 15 seconds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -640,6 +640,9 @@ SPHINXBUILD   = sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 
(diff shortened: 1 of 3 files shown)


klihub:
Closing in favor of #440 .

PR #434, opened 2020-10-12
docs: install git into the site builder container image

marquiz:
Add git so that we're able to detect the version in the container build.
And, something other than "failed to determine version from any source"
is shown in the upper left corner of the page.


klihub (approved):
LGTM

PR #435, opened 2020-10-12
docs/conf.py: use more proper URI checks for testing links, .{md,rst}-files.

klihub:
Only consider
  - links or paths with a trailing .md as a markdown file, don't get fooled by something like foo.md.bar
  - links or paths with a trailing .rst as a restructured text file, don't get fooled by something like foo.rst.bar
  - a leading http:// or https:// as a remote reference/link

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -115,12 +115,24 @@ def setup(app):
 #
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


marquiz (approved):
Thanks for the improvement!
lgtm

PR #439, opened 2020-10-12
e2e: add ubuntu-download-kernel()

askervin:
I used this to test different kernel versions on movable memory nodes when pinning container memory on those.

Do you think we could have this kind of helpers in our multi-distro support?

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -84,6 +84,30 @@ ubuntu-20_04-image-url() {
     echo "https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.img"
 }
 
+ubuntu-download-kernel() {
+    # Usage:
+    #   ubuntu-download-kernel list
+    #   ubuntu-download-kernel VERSION
+    #
+    # List or download Ubuntu kernel team kernels.
+    #
+    # Example:
+    #   ubuntu-download-kernel list | grep 5.9
+    #   ubuntu-download-kernel 5.9-rc8
+    #   vm-command "dpkg -i kernels/linux*rc8*deb"
+    #   vm-reboot
+    #   vm-command "uname -a"
+    local version=$1
+    [ -n "$version" ] ||
+        error "missing kernel version to install"
+    if [ "$version" == "list" ]; then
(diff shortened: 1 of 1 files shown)


klihub:
> I used this to test different kernel versions on movable memory nodes when pinning container memory on those.
> 
> Do you think we could have this kind of helpers in our multi-distro support?

Sure. I'm just wondering what would be the best abstraction to avoid ending up with
a round hole and a square peg for some distros, as many create modular kernel
packages compared to debian/ubuntu. Maybe we'd need something like this:

```
distro-list-kernel-versions() { distro-resolve "$@"; }
distro-install-kernel-pkgs()  { distro-resolve "$@"; }
...
debian-list-kernel-versions() {
    debian-script-to-list-kernel-versions
}

debian-install-kernel-pkgs() {
    local version="$1"; shift
    local subpkgs="$*"
    if [ -n "$subpkgs" ]; then
        command-error "debian(-like) kernel packages are monolithic, no subpackages \"$subpkgs\""
    fi
    debian-script-to-install-kernel $version
}

fedora-list-kernel-versions() {
    fedora-script-to-list-kernel-versions
}

fedora-install-kernel-pkgs() {
    local version="$1"; shift
    local subpkgs="$*"
    fedora-script-to-install-kernel $version $subpkgs
}
```

Then you could use these something like this:
```
    oldone=$(distro-list-kernel-versions | sort | head -1)
    case $VM_DISTRO in
        debian*|ubuntu*)
            distro-install-kernel-pkgs $oldone
            ;;
        fedora*|centos*)
            distro-install-kernel-pkgs $oldone core modules modules-extra
            ;;
    esac
```

klihub (approved):
If we keep this FTTB as ubuntu-specific, just for a few selected special tests, then this is fine as it is. We can try to generalize it later if it looks necessary.

askervin:
> If we keep this FTTB as ubuntu-specific, just for a few selected special tests, then this is fine as it is. We can try to generalize it later if it looks necessary.

I'm fine with that as well. Now this function has only been a helper in manual testing. We can make it an integrated part of the framework later as well. (If we end up doing so, I hope we'll remember your comments in this PR.)

PR #440, opened 2020-10-12
Implement versioned documentation

marquiz:
This PR implements a complete pipeline of versioned documentation:
- adds a helper script for updating the `gh-pages` branch
- customizes the sphinx theme to include a version menu
- integrates versioned documentation into github actions

The basic idea here is that documentation is built whenever `master` or any `release-*` branch is update. It builds documentation for the updated "version" only, i.e. we don't always rebuild all documentation. The reasoning behind is:
- Only maintain documentation per major release. Latest point release (e.g. `v0.x.(y+1)` should be compatible with the previous point release (`v0.x.y`), also documentation-wise. This is what many other projects do, too.
- Build on updates of release branch, not on tags in order to make it easier to maintain older documentation. E.g. there is no need to make a new release if only documentation is updated. This logic is simple, stupid. Another way to achieve this with better "gating" would be to just build on tags by default, but, allow gh-pages update if the tip commit has a certain "meta-tag" e.g. `gh-pages: publish` or smth.
- Only build docs for the ref/rev that was updated. That's what we do for everything else (container images etc), too. The specifics how to build/update gh-pages for certain ref is coded in `.github/workflows/publish-docs.yml`. We don't want to guess/depend how documentation is built in other branches as tools/versions/the whole logic might change over time.

A demonstration, generated with github actions, with two versions (master and 0.4) can be seen in my personal repo:
https://marquiz.github.io/cri-resource-manager

An alternative approach to #433

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -4,7 +4,9 @@ on:
   push:
     branches:
         - master
-        - documentation/test
+        - release-*
+    tags:
+        - v*
 
 jobs:
   update-gh-pages:
@@ -13,24 +15,24 @@ jobs:
     - uses: actions/checkout@v1
     - name: Build docs builder image
       run: make site-build
-    - name: Branch off base for updated gh-pages
+
+    - name: Fetch gh-pages
+      run: git fetch --no-tags --prune --depth=1 origin refs/heads/gh-pages:refs/heads/gh-pages
+
+    - name: Install build dependencies
       run: |
-        git branch gh-pages
-        git checkout gh-pages
-    - name: Add latest generated docs to gh-pages
-      shell: bash
+        pip3 install --user -r docs/requirements.txt
+        echo "`python3 -m site --user-base`/bin" >> $GITHUB_PATH
+
+    - name: Add docs from this revision to gh-pages
       run: |
-        touch _build/.nojekyll
-        rm -rf _build/.doctrees _build/_sources
-        git --work-tree _build add --no-all -v .
-        author=$(git log --merges --pretty=format:%an -n 1)
-        email=$(git log --merges --pretty=format:%ae -n 1)
-        git config --global user.name "$author"
-        git config --global user.email "$email"
-        git commit -m "publish-docs.yml: generated latest html output."
-    - name: Publish/force-push to gh-pages
+        git config user.name "Github"
+        git config user.email "no-reply@github.com"
+        ./scripts/build/update-gh-pages.sh
+
+    - name: Publish/push to gh-pages
       shell: bash
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        git push -f https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git HEAD:gh-pages
+        git push https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git gh-pages
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -26,6 +26,7 @@ author = 'various'
 
 master_doc = 'docs/index'
 
+
 ##############################################################################
 #
 # This section determines the behavior of links to local items in .md files.
@@ -63,6 +64,9 @@ else:
 version = getenv("SITE_VERSION", default="unknown")
 release = getenv("BUILD_VERSION", default="unknown")
 
+# Versions to show in the version menu
+versions = getenv('VERSIONS', default="").split()
+html_context = {'versions': sorted(set(versions))}
 
 # -- General configuration ---------------------------------------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,7 +61,19 @@ else:
     githubFileURL = githubFileURL + baseBranch + "/"
     githubDirURL = githubDirURL + baseBranch + "/"
 
-version = getenv("SITE_VERSION", default="unknown")
+# Version displayed in the upper left corner of the site
+ref = getenv('GITHUB_REF', default="")
+if ref == "refs/heads/master":
+    version = "devel"
+elif ref.startswith("refs/heads/release-"):
+    # For release branches just show the latest tag name
+    buildVersion = getenv("BUILD_VERSION", default="unknown")
+    version = buildVersion.split('-')[0]
+elif ref.startswith("refs/tags/"):
+    version = ref[len("refs/tags/"):]
+else:
+    version = getenv("BUILD_VERSION", default="unknown")
+
 release = getenv("BUILD_VERSION", default="unknown")
 
 # Versions to show in the version menu
--- a/scripts/build/update-gh-pages.sh
+++ b/scripts/build/update-gh-pages.sh
@@ -1,0 +1,137 @@
+#!/bin/bash -e
+set -o pipefail
+
+script=`basename $0`
+
+usage () {
+cat << EOF
+Usage: $script [-h] [-a] [BUILD_SUBDIR]
+
+Options:
+  -h         show this help and exit
+  -a         amend (with --reset-author) instead of creating a new commit
+EOF
+}
+
+#
+# Argument parsing
+#
+while [ "${1#-}" != "$1" -a -n "$1" ]; do
+    case "$1" in
+        -a|--amend)
+            amend="--amend --reset-author"
+            ;;
+        -h|--help)
+            usage
+            exit 0
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+build_subdir="$1"
+
+# Check that no extra args were provided
+if [ $# -gt 1 ]; then
+    echo "ERROR: unknown arguments: $@"
+    usage
+    exit 1
+fi
+
+#
+# Build the documentation
+#
+build_dir="_build-ghpages"
+echo "Creating new Git worktree at $build_dir"
+git worktree add "$build_dir" gh-pages
+
+# Drop worktree on exit
+trap "echo 'Removing Git worktree $build_dir'; git worktree remove --force '$build_dir'" EXIT
+
+# Parse subdir name from GITHUB_REF
+if [ -z "$build_subdir" ]; then
+    case "$GITHUB_REF" in
+        refs/tags/*)
+            _base_ref=${GITHUB_REF#refs/tags/}
+            ;;
+        refs/heads/*)
+            _base_ref=${GITHUB_REF#refs/heads/}
+            ;;
+        *) _base_ref=
+    esac
+    echo "Parsed baseref: '$_base_ref'"
+
+    case "$GITHUB_REF" in
+        refs/tags/v*)
+            _version=${GITHUB_REF#refs/tags/v}
+            ;;
+        refs/heads/release-*)
+            _version=${GITHUB_REF#refs/heads/release-}
+            ;;
+        *) _version=
+    esac
+    echo "Detected version: '$_version'"
+
+    _version=`echo -n $_version | sed -nE s'!^([0-9]+\.[0-9]+).*$!\1!p'`
+
+    # Use version as the subdir
+    build_subdir=${_version:+v$_version}
+    # Fallback to base-ref i.e. name of the branch or tag
+    if [ -z "$build_subdir" ]; then
+        # For master branch we use the name 'devel'
+        [ "$_base_ref" = "master" ] && build_subdir=devel || build_subdir=$_base_ref
+    fi
+fi
+
+# Default to 'devel' if no subdir was given and we couldn't parse
+# it
+build_subdir=${build_subdir:-devel}
+echo "Updating site version subdir: '$build_subdir'"
+export SITE_BUILDDIR="$build_dir/$build_subdir"
+
+# Detect existing versions from the gh-pages branch
+# 'stable' is a symlink pointing to the latest version
+cd "$build_dir"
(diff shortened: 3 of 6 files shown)


docs/conf.py
```diff
@@ -26,6 +26,16 @@
 
 master_doc = 'docs/index'
 
+# Versions to show in the version menu
+# NOTE: we could be smarter and detect these from branch/tag names,
+# for example
+versions = [
+        'master',
+        'v0.4'
+        ]
```
klihub, line 35: Please, could we avoid hardcoding this ? Instead let's runtime-detect all the versions we want to show documentation for and then, for instance, pass that as a comma-separated string in an suitably named environment variable, splitting+stripping it here into versions.

.github/workflows/publish-docs.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
         - master
-        - documentation/test
+        - release-*
```
klihub, line 7: I continue to be of the opinion that we should generate versioned docs *at least* for tagged versions without an accompanying release-* branch. And preferably for all tagged versions which are new enough to have sphinx-based documentation.

docs/conf.py
```diff
@@ -60,7 +70,17 @@
     githubFileURL = githubFileURL + baseBranch + "/"
     githubDirURL = githubDirURL + baseBranch + "/"
 
-version = getenv("SITE_VERSION", default="unknown")
+# Version dieplayed in the upper left corner of the site
```
klihub, line 73: typo: s/dieplayed/displayed/

klihub (commented):
This looks pretty good to me already as it is. I really like the simplicity in the idea of generating incrementally just a single version, the right one for the git action corresponding to the current workflow. It is so much simpler than the approach I was prototyping with.

That said I have a few very simple things I'd like added/changed.

1) In addition to the rolling master- and release-branches, I'd like to generate documentation for tagged/point releases (v[0-9]+.[0-9]+.[0-9]+ tags). Documentation is just like the other artifacts (tarballs, packages and images, release notes) we generate for tagged releases. Tagging a release should produce documentation even if we don't branch off an accompanying rolling release-branch. Secondly, just like we don't nuke previous release artifacts for point-releases we should not nuke the documentation for them either.

2) I would like to get rid of the current hardcoded versions = [ 'master', '0.4' ] and have that automatically updated/generated.

klihub:
> * Only maintain documentation per major release. Latest point release (e.g. `v0.x.(y+1)` should be compatible with the previous point release (`v0.x.y`), also documentation-wise. This is what many other projects do, too.

Then there are [others](https://www.python.org/doc/versions/) which do provide docs for all point-releases.

> * Build on updates of release branch, not on tags in order to make it easier to maintain older documentation. E.g. there is no need to make a new release if only documentation is updated. This logic is simple, stupid. Another way to achieve this with better "gating" would be to just build on tags by default, but, allow gh-pages update if the tip commit has a certain "meta-tag" e.g. `gh-pages: publish` or smth.

I think instead of some meta-tag the easiest would be to simply generate tag-/version-specific docs for tagged releases (always produces a new versioned set of the docs), and then rolling docs for master and release-* branches (always updates the same master or rolling-release-specific versioned set of the docs).

scripts/build/update-gh-pages.sh
```diff
@@ -0,0 +1,125 @@
...
+            exit 0
+            ;;
+        a)  amend="--amend --reset-author"
+            ;;
+        *)  usage
+            exit 1
+            ;;
+    esac
+done
+shift "$((OPTIND - 1))"
```
klihub, line 31: minor nitpick: we consistently avoid using getopt in all of our other shell scripts (mostly because it's so limited). In the other scripts we use something like this with small variations:
```
while [ "${1#-}" != "$1" -a -n "$1" ]; do
    case "$1" in 
        -a|--amend)
            amend="--amend --reset-author"
            ;;
        -h|--help)
            usage
            exit 0
            ;;
        *)
            usage
            exit 1
            ;;
    esac
    shift
done
```

marquiz:
Thanks for prompt review!

> Then there are [others](https://www.python.org/doc/versions/) which do provide docs for all point-releases.

If you look closer, that's exactly what they DON'T do. They only have links per major relese, 3.5, 3.6, 3.7 etc. Of course there are projects that do that but I don't necessarily think it's a good idea 😄 


> I think instead of some meta-tag the easiest would be to simply generate tag-/version-specific docs for tagged releases (always produces a new versioned set of the docs), and then rolling docs for master and release-* branches (always updates the same master or rolling-release-specific versioned set of the docs).

I have hard time understanding the user-value for separate rolling release branch documentation. 


> 1. In addition to the rolling master- and release-branches, I'd like to generate documentation for tagged/point releases (v[0-9]+.[0-9]+.[0-9]+ tags). Documentation is just like the other artifacts (tarballs, packages and images, release notes) we generate for tagged releases. Tagging a release should produce documentation even if we don't branch off an accompanying rolling release-branch. Secondly, just like we don't nuke previous release artifacts for point-releases we should not nuke the documentation for them either.

I beg to disagree here. IMO this just creates unnecessary clutter from the reader's point of view. I still believe that point releases should be backwards compatible, also from the documentation point of view. It's also a lot easier to add deprecation notices and such this way (e.g. declare an old version branch unmaintained) when there's just one documentation "branch" per major release.

I agree on documentation on tags (rather than only on release branches), though.


> 2\. I would like to get rid of the current hardcoded versions = [ 'master', '0.4' ] and have that automatically updated/generated.

Yeah. That's something I had in mind but didn't have the time to create a thought-out implementation.

scripts/build/update-gh-pages.sh
marquiz: I don't think the limited nature of getopt really matters in script like this not intended to be really executed by users. But, I'm quite indifferent here and don't have a particular affection towards getopt, either. Consistency is always good. I'll change this

docs/conf.py
marquiz: Thanks, will fix

marquiz:
> Documentation is just like the other artifacts

Still returning to this. I disagree with this and I think this is the main difference in our views 😉  I don't believe that documentation is a "sacred" release artifact like the binaries (or source code). If there are mistakes or big gaps in the documentation  for a certain release there should be an easy way to fix this. History is preserved and there is a way to look back if needed but the readers (i.e. end users) really should not need to care.

klihub:
> Thanks for prompt review!
> 
> > Then there are [others](https://www.python.org/doc/versions/) which do provide docs for all point-releases.
> 
> If you look closer, that's exactly what they DON'T do.

Simply clicking open the above link proves my point. But I'll leave it at that.

> 
> > I think instead of some meta-tag the easiest would be to simply generate tag-/version-specific docs for tagged releases (always produces a new versioned set of the docs), and then rolling docs for master and release-* branches (always updates the same master or rolling-release-specific versioned set of the docs).
> 
> I have hard time understanding the user-value for separate rolling release branch documentation.

But that is what you have essentially implemented, it's not me asking for it. AFAICT, if we have a branch called release-X.Y, we'll produce documentation for it even if we never tag a single release from it. And no matter how many X.Y.Z releases we tag, if we don't create a release-X.Y branch there will be no documentation generated. There's a bit too much hand-work involved for my taste. That was my whole point: IMO it would be much simpler to have docs for tagged releases with no hand-work involved.

> 
> > 1. In addition to the rolling master- and release-branches, I'd like to generate documentation for tagged/point releases (v[0-9]+.[0-9]+.[0-9]+ tags). Documentation is just like the other artifacts (tarballs, packages and images, release notes) we generate for tagged releases. Tagging a release should produce documentation even if we don't branch off an accompanying rolling release-branch. Secondly, just like we don't nuke previous release artifacts for point-releases we should not nuke the documentation for them either.
> 
> I beg to disagree here. IMO this just creates unnecessary clutter from the reader's point of view. I still believe that point releases should be backwards compatible, also from the documentation point of view. It's also a lot easier to add deprecation notices and such this way (e.g. declare an old version branch unmaintained) when there's just one documentation "branch" per major release.
> 
> I agree on documentation on tags (rather than only on release branches), though.
> 
> > 2. I would like to get rid of the current hardcoded versions = [ 'master', '0.4' ] and have that automatically updated/generated.
> 
> Yeah. That's something I had in mind but didn't have the time to create a thought-out implementation.



> Thanks for prompt review!
> 
> > Then there are [others](https://www.python.org/doc/versions/) which do provide docs for all point-releases.
> 
> If you look closer, that's exactly what they DON'T do. They only have links per major relese, 3.5, 3.6, 3.7 etc. Of course there are projects that do that but I don't necessarily think it's a good idea 
> 
> > I think instead of some meta-tag the easiest would be to simply generate tag-/version-specific docs for tagged releases (always produces a new versioned set of the docs), and then rolling docs for master and release-* branches (always updates the same master or rolling-release-specific versioned set of the docs).
> 
> I have hard time understanding the user-value for separate rolling release branch documentation.
> 
> > 1. In addition to the rolling master- and release-branches, I'd like to generate documentation for tagged/point releases (v[0-9]+.[0-9]+.[0-9]+ tags). Documentation is just like the other artifacts (tarballs, packages and images, release notes) we generate for tagged releases. Tagging a release should produce documentation even if we don't branch off an accompanying rolling release-branch. Secondly, just like we don't nuke previous release artifacts for point-releases we should not nuke the documentation for them either.
> 
> I beg to disagree here. IMO this just creates unnecessary clutter from the reader's point of view. I still believe that point releases should be backwards compatible, also from the documentation point of view. It's also a lot easier to add deprecation notices and such this way (e.g. declare an old version branch unmaintained) when there's just one documentation "branch" per major release.
> 
> I agree on documentation on tags (rather than only on release branches), though.
> 
> > 2. I would like to get rid of the current hardcoded versions = [ 'master', '0.4' ] and have that automatically updated/generated.
> 
> Yeah. That's something I had in mind but didn't have the time to create a thought-out implementation.

docs/conf.py
```diff
@@ -26,6 +26,16 @@
 
 master_doc = 'docs/index'
 
+# Versions to show in the version menu
+# NOTE: we could be smarter and detect these from branch/tag names,
+# for example
+versions = [
+        'master',
+        'v0.4'
```
kad, line 34: @marquiz We need to have "human readable labels" for the branches. E.g. "master" -> "development", "release-v0.4" -> "stable" and "latest".

docs/conf.py
marquiz: Fixed

scripts/build/update-gh-pages.sh
marquiz: Changed

docs/conf.py
marquiz: Changed in the latest version:
- add symlink `stable` to the latest release
- use name `devel` for master branch

marquiz:
> Simply clicking open the above link proves my point. But I'll leave it at that.

Point taken 😇

> But that is what you have essentially implemented, it's not me asking for it. AFAICT, if we have a branch called release-X.Y, we'll produce documentation for it even if we never tag a single release from it. And no matter how many X.Y.Z releases we tag, if we don't create a release-X.Y branch there will be no documentation generated. There's a bit too much hand-work involved for my taste. That was my whole point: IMO it would be much simpler to have docs for tagged releases with no hand-work involved.

Yes that is true. My argument wasn't clearly stated. Probably my main concern is/was that I don't want to confuse the user and clustter the "version menu" with point releases. Thus, I take back my words and I agree that we could store the documentation for point releases, too. It might be useful e.g. to have a link to the exact version of documentation in the release notes/release page. But, I think those should be stored in a separate area, e.g. `releases/` subfolder (in `gh-pages`) or smth and not shown in the main version menu.

Anyway, I now pushed a new version of the PR that should address all the comments and more. It's now a set of patches on top of the previous ones so it should be easier for you to review what was changed. I'll squash this PR back into a meaningful set when we're happy with this.

Handling/storing point release documentation is not implemented yet, and, I believe that can be left as a future exercise.

klihub:
> > But that is what you have essentially implemented, it's not me asking for it. AFAICT, if we have a branch called release-X.Y, we'll produce documentation for it even if we never tag a single release from it. And no matter how many X.Y.Z releases we tag, if we don't create a release-X.Y branch there will be no documentation generated. There's a bit too much hand-work involved for my taste. That was my whole point: IMO it would be much simpler to have docs for tagged releases with no hand-work involved.
> 
> Yes that is true. My argument wasn't clearly stated. Probably my main concern is/was that I don't want to confuse the user and clustter the "version menu" with point releases. Thus, I take back my words and I agree that we could store the documentation for point releases, too. It might be useful e.g. to have a link to the exact version of documentation in the release notes/release page. But, I think those should be stored in a separate area, e.g. `releases/` subfolder (in `gh-pages`) or smth and not shown in the main version menu.

Yes, I agree. Better to have the full list in a separate page somewhere to avoid ever-growing clutter in the main documentation site.

> Anyway, I now pushed a new version of the PR that should address all the comments and more. It's now a set of patches on top of the previous ones so it should be easier for you to review what was changed. I'll squash this PR back into a meaningful set when we're happy with this.
> 
> Handling/storing point release documentation is not implemented yet, and, I believe that can be left as a future exercise.

Sure, no sweat. This is very much ready to be merged from my PoV. It improves hugely on the current situation.

klihub (approved):
LGTM. All in all, really good stuff ! Thanks !

marquiz:
I'm still doing some checks. Also, we need to reset gh-pages branch before merging this

klihub (changes requested):
Need to squash/edit history and reset gh-pages before merging this.

klihub (approved):
LGTM

PR #451, opened 2020-10-15
docs: create file references to github repo

marquiz:
When referencing files in this repo, create links to our github repo
instead of making a local copy.

Also, fixes handling of file references. When running through missing
references, only ignore file references with an absolute path. These
will reference the repo root directory and will be turned later on to
references to GitHub, directly. Missing references with a relative path
raise an error (as they should because the file really is missing).

Fix the broken references in the documentation, too, so that the
documentation builds without errors.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,15 +51,21 @@ master_doc = 'docs/index'
 baseBranch = "master"
 useGitHubURL = True
 commitSHA = getenv('GITHUB_SHA')
-githubBaseURL = "https://github.com/intel/cri-resource-manager/"
-githubFileURL = githubBaseURL + "blob/"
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -49,7 +49,7 @@
 ##############################################################################
 
 baseBranch = "master"
-useGitHubURL = False
+useGitHubURL = True
 commitSHA = getenv('GITHUB_SHA')
 githubBaseURL = "https://github.com/intel/cri-resource-manager/"
```
klihub, line 54: I think the default base URL should be possible to override, for instance from the workflow to get it
automatically working correctly in forks as well.

klihub (approved):
Looks fine to me.

PR #454, opened 2020-10-15
memtier, controllers: split out dynamic page demotion into a controller.

klihub:
This patch series is the first step in generalizing and improving dynamic page demotion by
introducing a new `page-migration` controller and splitting out the core of the functionality
from the `memtier` policy to this new controller. With this patch series applied any policy
can activate page demotion for container by simply attaching demotion policy decisions to
the desired containers.

In particular, the patch series
- adds functions for getting/setting page-migration policy/preferences to cache.Container
- adds a new `page-migration` controller
- moves the core of the demotion functionality to the new controller
- only leaves demotion/migration policy decision making in `memtier`
- fixes `resource-manager` to execute controller post-stop hooks for deleted containers

The current page-migration policy decision is implicitly 'idle page migration', with the policy
only providing the set of source and target nodes. Detecting idle pages and migrating them
from sources to targets is entirely taken care of in the page-migration controller. This should
be fairly easy to generalize in the future.

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -33,6 +33,7 @@ import (
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/config"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/kubernetes"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
+	system "github.com/intel/cri-resource-manager/pkg/sysfs"
 	"github.com/intel/cri-resource-manager/pkg/topology"
 )
 
@@ -45,6 +46,8 @@ const (
 	BlockIO = "blockio"
 	// Memory marks changes that can be applied by the Memory controller.
 	Memory = "memory"
+	// PageMigration marks changes that can be applied by the PageMigration controller.
+	PageMigration = "page-migration"
 
 	// TagAVX512 tags containers that use AVX512 instructions.
 	TagAVX512 = "AVX512"
@@ -347,6 +350,11 @@ type Container interface {
 	// GetToptierLimit returns the top tier memory limit for the container.
 	GetToptierLimit() int64
 
+	// SetPageMigration sets the page migration policy/options for the container.
+	SetPageMigration(*PageMigrate)
+	// GetPageMigration returns the current page migration policy/options for the container.
+	GetPageMigration() *PageMigrate
+
 	// SetCRIRequest sets the current pending CRI request of the container.
 	SetCRIRequest(req interface{}) error
 	// GetCRIRequest returns the current pending CRI request of the container.
@@ -401,9 +409,10 @@ type container struct {
 	LinuxReq  *cri.LinuxContainerResources // used to estimate Resources if we lack annotations
 	req       *interface{}                 // pending CRI request
 
-	RDTClass     string // RDT class this container is assigned to.
-	BlockIOClass string // Block I/O class this container is assigned to.
-	ToptierLimit int64  // Top tier memory limit.
+	RDTClass     string       // RDT class this container is assigned to.
+	BlockIOClass string       // Block I/O class this container is assigned to.
+	ToptierLimit int64        // Top tier memory limit.
+	PageMigrate  *PageMigrate // Page migration policy/options for this container.
 
 	pending map[string]struct{} // controllers with pending changes for this container
 
@@ -444,6 +453,27 @@ type Device struct {
 	Host string
 	// Permissions specify the device permissions for the container.
 	Permissions string
+}
+
+// PageMigrate contains the policy/preferences for container page migration.
+type PageMigrate struct {
+	SourceNodes system.IDSet // idle memory pages on these NUMA nodes
+	TargetNodes system.IDSet // should be migrated to these NUMA nodes
+}
+
+// Clone creates a copy of the page migration policy/preferences.
+func (pm *PageMigrate) Clone() *PageMigrate {
+	if pm == nil {
+		return nil
+	}
+	c := &PageMigrate{}
+	if pm.SourceNodes != nil {
+		c.SourceNodes = pm.SourceNodes.Clone()
+	}
+	if pm.TargetNodes != nil {
+		c.TargetNodes = pm.TargetNodes.Clone()
+	}
+	return c
 }
 
 //
--- a/pkg/cri/resource-manager/cache/container.go
+++ b/pkg/cri/resource-manager/cache/container.go
@@ -811,6 +811,15 @@ func (c *container) GetToptierLimit() int64 {
 	return c.ToptierLimit
 }
 
+func (c *container) SetPageMigration(p *PageMigrate) {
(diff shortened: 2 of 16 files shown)


pkg/cri/resource-manager/cache/cache.go
```diff
@@ -446,6 +455,26 @@ type Device struct {
 	Permissions string
 }
 
+// PageMigrate contains the policy/options for container page migration.
+type PageMigrate struct {
+	SourceNodes system.IDSet // idle memory pages on these NUMA nodes
+	TargetNodes system.IDSet // should be migrated to these NUMA nodes
+}
+
+func (pm *PageMigrate) Clone() *PageMigrate {
```
marquiz, line 464: nit: add a comment/docstring to the exported methods

pkg/cri/resource-manager/cache/container.go
```diff
@@ -811,6 +811,15 @@ func (c *container) GetToptierLimit() int64 {
 	return c.ToptierLimit
 }
 
+func (c *container) SetPageMigration(p *PageMigrate) {
```
marquiz, line 814: nit: comment/docstring

pkg/cri/resource-manager/control/page-migrate/flags.go
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pagemigrate
+
+import (
+	"github.com/intel/cri-resource-manager/pkg/config"
+)
+
+// Options captures our conigurable controller parameters.
```
marquiz, line 21: s/Options/options/

pkg/cri/resource-manager/control/page-migrate/page-migrate.go
```diff
@@ -0,0 +1,99 @@
...
+	return nil
+}
+
+// PostStopHook is the controller's post-stop hook.
+func (m *migration) PostStopHook(cc cache.Container) error {
+	return nil
+}
+
+// migrationError creates a memory I/O-controller-specific formatted error message.
+func migrationError(format string, args ...interface{}) error {
```
marquiz, line 183: memory I/O-controller??

pkg/cri/resource-manager/control/page-migrate/demoter.go
```diff
@@ -217,7 +246,7 @@ func resetDirtyBit(pid string) error {
 }
 
 // ResetDirtyBit unsets soft-dirty bits for all processes in a container.
-func (d *demoter) ResetDirtyBit(c *container) error {
+func (d *demoter) resetDirtyBit(c *container) error {
```
marquiz, line 249: nit: Change docstring correspondingly

pkg/cri/resource-manager/control/page-migrate/demoter.go
```diff
@@ -97,32 +98,69 @@ func copyPagePool(p pagePool) pagePool {
...
+	if d.pageScanInterval > 0 && d.pageMoveInterval > 0 && d.maxPageMoveCount > 0 {
+		log.Info("scanning pages every %s, moving max. %d pages every %s",
+			d.pageScanInterval.String(), d.maxPageMoveCount, d.pageMoveInterval.String())
+		d.startDirtyBitResetTimer()
+	} else {
+		log.Info("scanning pages is disabled")
+	}
+}
+
+func (d *demoter) Stop() {
```
marquiz, line 119: nit: I'd add comment to the exported methods. Or, alternatively, if demoter implements an interface state that in the demoter definition

marquiz (commented):
Looks good, just a few very small nits.

One thing that puzzled me is why do we track containers locally in the controller? 🤔

klihub:
> Looks good, just a few very small nits.
> 
> One thing that puzzled me is why do we track containers locally in the controller?

To avoid having to serialize access to the cache against request and event processing,
IOW against the resource-manager and policies. Or to put it another way to avoid having
to perform the potentially lengthy process of page scanning with the cache locked (which
then blocks request processing). 

An alternative would have been to duplicate what we had in the policy:
  - introduce controller events (basically callbacks invoked by the resource-manager with the cache locked)
  - periodically inject a controller event when we want to scan pages
  - scan pages from the event callback, with resource-manager holding the lock for us

Instead we now track a bare minimal copy of cached containers, just enough for doing
page scanning and migration completely on our own.

pkg/cri/resource-manager/cache/container.go
klihub: It is present in the interface definition. Ditto for GetPageMigration().

pkg/cri/resource-manager/control/page-migrate/demoter.go
klihub: Added docstring to exported methods Stop() and Reconfigure().

pkg/cri/resource-manager/control/page-migrate/page-migrate.go
klihub: Fixed.

pkg/cri/resource-manager/control/page-migrate/flags.go
klihub: Updated. And fixed the typo, too... ;-)

pkg/cri/resource-manager/cache/cache.go
klihub: Added docstring.

pkg/cri/resource-manager/cache/container.go
marquiz: Yes, of course, sorry about this

marquiz:
> To avoid having to serialize access to the cache against request and event processing,
> IOW against the resource-manager and policies. Or to put it another way to avoid having
> to perform the potentially lengthy process of page scanning with the cache locked (which
> then blocks request processing).
> 
> An alternative would have been to duplicate what we had in the policy:
> 
> * introduce controller events (basically callbacks invoked by the resource-manager with the cache locked)
> * periodically inject a controller event when we want to scan pages
> * scan pages from the event callback, with resource-manager holding the lock for us
> 
> Instead we now track a bare minimal copy of cached containers, just enough for doing
> page scanning and migration completely on our own.

Thanks for the explanation, makes sense 👍 Do we have this documented somewhere? If not, I think it would be geed to add e.g. a short comment in the code explaining the reason we're doing this. Might make it easier to read/maintain code in the future. WDYT?

klihub:
> Thanks for the explanation, makes sense  Do we have this documented somewhere? If not, I think it would be geed to add e.g. a short comment in the code explaining the reason we're doing this. Might make it easier to read/maintain code in the future. WDYT?

Yes, that's very much true. I'll add a comment explaining this to the code.

klihub:
> Yes, that's very much true. I'll add a comment explaining this to the code.

Added a comment about why we copy a subset of the containers to the controller. While at it, also updated the code to use a set of functions mimicking the corresponding cache functions when dealing with locally cached containers.

marquiz (approved):
Wow, that comment is pretty exhaustive 😊 👍 

LGTM

PR #455, opened 2020-10-16
e2e: fix no_proxy to include .svc

askervin:
https requests from kube-apiserver to webhook services will escape
from the node to https_proxy, resulting in x509 certificate error on
unknown authority, unless .svc is in no_proxy.


marquiz (dismissed):
lgtm

klihub (approved):

PR #458, opened 2020-10-16
memtier: properly reconfigure/start/stop dynamic demoter.

klihub:
The commit to disable the dynamic demoter by default accidentally put it on a permanent vacation. It is not configured correctly for the initial configuration (unless that has it disabled) and since it does not get properly reconfigured upon configuration changes, this prevents it from ever being activated. This patch should fix that.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/memtier/dynamic_demotion.go
+++ b/pkg/cri/resource-manager/policy/builtin/memtier/dynamic_demotion.go
@@ -79,10 +79,7 @@ type demoter struct {
 
 // Demoter dynamically demotes pages from DRAM to PMEM.
 type Demoter interface {
-	// StartDirtyBitResetTimer starts the timer for getting memory moving events.
-	StartDirtyBitResetTimer(policy *policy, timeout time.Duration)
-	// StopDirtyBitResetTimer stops the memory moving.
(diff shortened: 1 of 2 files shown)


pkg/cri/resource-manager/policy/builtin/memtier/dynamic_demotion.go
```diff
@@ -117,6 +114,28 @@ func copyPagePool(p pagePool) pagePool {
 	return c
 }
 
+func CreateDemoter(p *policy, pageMover PageMover) *demoter {
```
kad, line 117: Minor nit: most of go libraries are using convention for Object, there is NewObject function. Maybe here also NewDemoter?

kad (commented):
lgtm, with very small nit.

pkg/cri/resource-manager/policy/builtin/memtier/dynamic_demotion.go
klihub: Ok, I'll update.

marquiz (commented):
Looks, good. The only comment/nit I have is that the exported functions/method (`NewDemoter`, Reconfigure`, are there others(?)) should have a "docstring" comment

PR #460, opened 2020-10-19
build: add rule for building binary dist tarballs.

klihub:


askervin (commented):
binary-dist tarball includes binaries under `/opt/intel/bin`, but `/etc/systemd/system/cri-resource-manager.service` in the package wants to run `/usr/bin/cri-resmgr`. Perhaps we'd like to use the same `PREFIX` there as well.

klihub:
> binary-dist tarball includes binaries under `/opt/intel/bin`, but `/etc/systemd/system/cri-resource-manager.service` in the package wants to run `/usr/bin/cri-resmgr`. Perhaps we'd like to use the same `PREFIX` there as well.

Ah, thanks for spotting that. I missed it altogether.

kad (approved):
lgtm

askervin (approved):
LGTM

Tested manually by unpacking to Ubuntu 20.04 rootfs and launching with `systemctl`. Started up nicely after `mv /etc/cri-resmgr/fallback.cfg.sample /etc/cri-resmgr/fallback.cfg`. And I believe this is how it is supposed to be.

PR #465, opened 2020-10-19
e2e: add vm-install-cri-resmgr-webhook to the vm library

askervin:
(Fix a couple of shellcheck warnings in the vm library, too.)

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -484,7 +484,7 @@ vm-install-cri-resmgr() {
         local bin_change
         local src_change
         bin_change=$(stat --format "%Z" "$BIN_DIR/cri-resmgr")
-        src_change=$(find "$HOST_PROJECT_DIR" -name '*.go' -type f | xargs stat --format "%Z" | sort -n | tail -n 1)
+        src_change=$(find "$HOST_PROJECT_DIR" -name '*.go' -type f -print0 | xargs -0 stat --format "%Z" | sort -n | tail -n 1)
         if [[ "$src_change" > "$bin_change" ]]; then
             echo "WARNING:"
             echo "WARNING: Source files changed - installing possibly outdated binaries from"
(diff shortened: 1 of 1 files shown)


klihub:
I'll leave this to @marquiz ...

demo/lib/vm.bash
```diff
@@ -455,25 +455,144 @@ vm-install-cri-resmgr() {
...
         error "vm-install-cri-resmgr: unknown binsrc=\"$binsrc\""
     fi
 }
 
+vm-cri-import-image() {
+    local image_name="$1"
+    local image_tar="$2"
+    case "$VM_CRI" in
+        containerd)
+            vm-command "ctr -n k8s.io images import \"$image_tar\" >/dev/null 2>&1; crictl --runtime-endpoint unix:///run/containerd/containerd.sock images | grep \"$image_name\"" ||
```
marquiz, line 478: Stupid question: aren't we able to examine the return value of `ctr` directly?

demo/lib/vm.bash
```diff
@@ -455,25 +455,144 @@ vm-install-cri-resmgr() {
...
+    # remove the image immediately after impoting it as part of garbage collection.
+    docker image save "$webhook_image_repotag" > "$webhook_image_tar"
+    vm-put-file "$webhook_image_tar" "webhook/$(basename "$webhook_image_tar")" || {
+        command-error "copying webhook image to VM failed"
+    }
+    vm-cri-import-image cri-resmgr-webhook "webhook/$(basename "$webhook_image_tar")"
+    # Install tools for creating webhook certificate: cfssl and jq
+    vm-command-q "command -v cfssl" >/dev/null || {
+        cat <<EOF | vm-pipe-to-file webhook/install-cfssl.sh
+VERSION=\$(curl --silent "https://api.github.com/repos/cloudflare/cfssl/releases/latest" | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
```
marquiz, line 516: Isn't openssl available in the vm?

demo/lib/vm.bash
```diff
@@ -455,25 +455,144 @@ vm-install-cri-resmgr() {
...
+    sed -e "s|IMAGE_PLACEHOLDER|$webhook_image_repotag|" \
+        -e 's|^\(\s*\)tolerations:$|\1tolerations:\n\1  - {\"key\": \"cmk\", \"operator\": \"Equal\", \"value\": \"true\", \"effect\": \"NoSchedule\"}|g' \
+        -e 's/imagePullPolicy: Always/imagePullPolicy: Never/' \
+        < "${HOST_PROJECT_DIR}/cmd/cri-resmgr-webhook/webhook-deployment.yaml" \
+        | vm-pipe-to-file webhook/webhook-deployment.yaml
+    # Create secret that contains svc.crt and svc.key for webhook deployment
+    local server_crt_b64 server_key_b64
+    server_crt_b64="$(vm-command-q "cat webhook/server-crt.pem" | base64 -w 0)"
+    server_key_b64="$(vm-command-q "cat webhook/server-key.pem" | base64 -w 0)"
+    cat <<EOF | vm-pipe-to-file --append webhook/webhook-deployment.yaml
```
marquiz, line 577: Do we really need to create certificates through kubernetes? Could/would it be simpler just to create a self-signed ca and crt/key pair signed by that with ssl?

marquiz (commented):
Some stupid questions. The general approach looks ok

demo/lib/vm.bash
askervin: Good question. This hack forced the test to run on Ubuntu 20.04 despite error from `ctr`. Anyway, that was a bad idea: the import error exit status _is_ a good reason to bail out even though the image would appear in the image list after failed importing. (This is what happens with containerd versions < 4.0.)

I verified that second importing causes an error even in the case that the same image appears in the list (after the first import having failed but image being added to images list). Therefore we actually should rely on ctr exit status rather than image list contents.

Thanks, fixed!

demo/lib/vm.bash
askervin: Yes it is. Switched to it.

demo/lib/vm.bash
askervin: No we don't. How happy can one be when replacing a code like that with pretty much a single line... :) :) :) openssl 1.1.1 rocks.

Maybe we could have the openssl magic in our user guide, WDYT?

marquiz: > Maybe we could have the openssl magic in our user guide, WDYT?

Mm, might be useful, agree

marquiz (commented):
Looks good, thanks!

One thing that caught my eye is the use of escaping quotes. It might be a personal preference and partly about style uniformity, but, I'd only use escaping where needed. Some examples below:

Combined use of single and double quotes, e.g.:
`vm-command "ctr -n k8s.io images import \"$image_tar\""`
---> 
`vm-command "ctr -n k8s.io images import '$image_tar'"`

Drop unnecessary escaping:
`'s|^\(\s*\)tolerations:$|\1tolerations:\n\1  - {\"key\": \"cmk\", \"operator\": \"Equal\", \"value\": \"true\", \"effect\": \"NoSchedule\"}|g'`
--->
`'s|^\(\s*\)tolerations:$|\1tolerations:\n\1  - {"key": "cmk", "operator": "Equal", "value": "true", "effect": "NoSchedule"}|g'`

WDYT?

askervin:
@marquiz , thanks, now it's cleaner!

marquiz (approved):
LGTM

PR #467, opened 2020-10-19
e2e: add and record environment variables in backend

askervin:
- py_consts contains user-defined Python code to be run before every
  pyexec (that is, pp, verify, report etc.). User can define,
  for instance, configurations like exclusive_cores={...}
  and use those in verifications: 'cores["pod0c0"] in
  exclusive_cores'.
- cri_resmgr_agent_extra_args variable is a place holder for passing
  extra arguments to cri-resmgr-agent. This will be taken into use
  in later patches that implement and refactor launcher code.
- Values of all test environment variables are stored under test
  output directory for the record.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -71,6 +71,8 @@ usage() {
     echo "    cri_resmgr_cfg: configuration file forced to cri-resmgr."
     echo "    cri_resmgr_extra_args: arguments to be added on cri-resmgr"
     echo "             command line when launched"
+    echo "    cri_resmgr_agent_extra_args: arguments to be added on"
+    echo "              cri-resmgr-agent command line when launched"
     echo "    vm_files: \"serialized\" associative array of files to be created on vm"
     echo "             associative array syntax:"
     echo "             vm_files['/path/file']=file:/path/on/host"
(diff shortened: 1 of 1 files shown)


test/e2e/run.sh
```diff
@@ -748,8 +759,10 @@ vm=${vm:=$(basename ${TOPOLOGY_DIR})-${distro}-${cri}}
 vm_files=${vm_files-""}
 cri_resmgr_cfg=${cri_resmgr_cfg-"${SCRIPT_DIR}/cri-resmgr-memtier.cfg"}
 cri_resmgr_extra_args=${cri_resmgr_extra_args-""}
+cri_resmgr_agent_extra_args=${cri_resmgr_agent_extra_args-""}
 cleanup=${cleanup-0}
 reinstall_cri_resmgr=${reinstall_cri_resmgr-0}
+py_consts="${py_consts-''}"
```
klihub, line 765: nitpick: same nitpick here as for earlier patches regarding `:-` vs. `-`

askervin: Switched to `:-`.

klihub (approved):

PR #469, opened 2020-10-19
e2e: fix test suite detection in the test runner

askervin:
- Replace implicit detection of test suite root directory (based
  on *.cfg file) by finding test suite documentation file.
- Add test suite documentation files for the two existing suites:
  packages and policies.

PR #468, opened 2020-10-19
e2e: add extended-resources to script API

askervin:
- extended-resources enables test code to add and remove extended
  resources from the cluster.

PR #466, opened 2020-10-19
e2e: fix static-pools cache in pyexec

askervin:
- cache.PolicyJSON.allocations is not available in static-pools
  policy. Handle missing data gracefully when initializing variables
  for pyexec/pp/verify/report.

PR #461, opened 2020-10-19
e2e: add vm-wait-process and vm-wait-until to script API

askervin:

PR #472, opened 2020-10-19
e2e: add tests for static-pools

askervin:
Opening the first static-pools tests patch for comments & discussion:
- What would be the minimum features that you'd like to see tested as part of this commit?
- Do you see major issues with the test structure? For instance, unlike in memtier tests, here cri-resmgr, agent and webhook are restarted before every test execution.

### DIFF
--- a/test/e2e/policies.test-suite/static-pools/static-pools-lib.source.sh
+++ b/test/e2e/policies.test-suite/static-pools/static-pools-lib.source.sh
@@ -1,0 +1,25 @@
+# shellcheck disable=SC2148
+static-pools-relaunch-cri-resmgr() {
+    local webhook_running=0
+    out "# Relaunching cri-resmgr and agent, launch webhook if not already running"
+
+    vm-command-q "kubectl get mutatingwebhookconfiguration/cri-resmgr" >& /dev/null && {
+        webhook_running=1
+    }
+    # cleanup
+    terminate cri-resmgr
+    terminate cri-resmgr-agent
+    vm-command "rm -rf /var/lib/cri-resmgr"
+    extended-resources remove cmk.intel.com/exclusive-cpus >/dev/null
+
+    # launch again
+    launch cri-resmgr-agent
+    launch cri-resmgr
+    vm-run-until "! kubectl get node | grep NotReady" ||
+        error "kubectl node is NotReady after launching cri-resmgr-agent and cri-resmgr"
+    if [ "$webhook_running" == 0 ]; then
+        vm-command-q "[ -f webhook/webhook-deployment.yaml ]" ||
+            install cri-resmgr-webhook
+        launch cri-resmgr-webhook
+    fi
+}
--- a/test/e2e/policies.test-suite/static-pools/README.txt
+++ b/test/e2e/policies.test-suite/static-pools/README.txt
@@ -1,0 +1,11 @@
+# E2E static-pools policy test
+
+## Requirements
+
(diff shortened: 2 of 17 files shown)


askervin:
This PR depends on PR's #461, #463, #465, #466, #467, #468, #469, #470 and #471.

klihub:
I'll leave this to @marquiz ...

askervin:
@marquiz, @klihub, all tests now pass when PR #465 is picked.

I think this is patch is now ready, "RFC" cleared.

Ran all tests also rebased on top of PR #494 and VM deleted, and all passed there as well.

test/e2e/policies.test-suite/static-pools/static-pools-lib.source.sh
```diff
@@ -0,0 +1,25 @@
+# shellcheck disable=SC2148
+static-pools-relaunch-cri-resmgr() {
+    local webhook_running=0
+    out "# Relaunching cri-resmgr and agent, launch webhook if not already running"
+
+    vm-command-q "kubectl get mutatingwebhookconfiguration/cri-resmgr" >/dev/null 2>&1 && {
```
klihub, line 6: extreme nitpick: since we assume bash, isn't `>& /dev/null` an easier/more readable way to write this ?

klihub:
I skimmed through this and the test scripts et. al look good to me. I don't know enough about `CMK/static-pools` to tell whether the test set coverage is good enough. I let @marquiz pass judgement on that.

test/e2e/policies.test-suite/static-pools/static-pools-lib.source.sh
askervin: Thanks, fixed!

askervin:
Latest force-push change in addition to `s|>/dev/null 2>&1|>&/dev/null`: switched from `vm-command` to `vm-run-until` in the negative tests where checking error from pod description. It's more robust. Fixed a false test failure.

klihub:
Looks good to me.

klihub:
WDYT @marquiz ?

marquiz (approved):
Wow, really nice work @askervin ! And, apologies for the delay in my review.

The tests look really extensive. There are just two small things that popped into my mind:
- we could add check that the `CMK_CPUS_INFRA` and `CMK_CPUS_SHARED` env variables are correctly set
- we could have a negative test case for exclusive pool where no exclusive cores is requested

But these can be left as a future excercise (add an issue for those). I'm fine with merging this in its current state, too.

askervin:
@marquiz , @klihub , thanks for the review!

Because this inconveniently large patch looks approved, I added tests that @marquiz mentioned to PR #496  - in order to let this patch flow in now.

However, that PR heavily depends on this patch, so now the only commit in this patch is included in that PR, too. I don't know if this is very convenient way to handle this kind of dependency, but could not really come up with anything better either...

askervin:
Rebased and squashed #496 into this.

```
$ ./run_tests.sh policies.test-suite/static-pools/
...
PASS static-pools/n4c16/test00-node-status
PASS static-pools/n4c16/test01-exclusive-pods
PASS static-pools/n4c16/test02-pods-without-cmk
PASS static-pools/n4c16/test03-cmk-isolate
PASS static-pools/n4c16/test04-cmk-isolate-noaffinity
PASS static-pools/n4c16/test05-negative-tests
```

marquiz (approved):
LGTM!

PR #473, opened 2020-10-21
e2e,demo: check SSH key in/load SSH key to agent.

klihub:
During the environment check, in addition to checking the mere existence of a stock SSH public key
  - check if the corresponding private key is loaded to the agent,
  - try loading it if it is not loaded yet,
  - give up with an error message instructing the user what to do, if loading the key fails

Without this patch, if an encrypted/password-protected key is not loaded to the agent yet, the test
framework just sits in a loop forever, waiting for ssh'ing to the VM to eventually succeed, without
giving the user an indication about the real problem.

### DIFF
--- a/demo/lib/host.bash
+++ b/demo/lib/host.bash
@@ -197,3 +197,12 @@ host-delete-vm() {
         command-error "deleting govm \"$VM_NAME\" failed"
     }
 }
+
+host-is-encrypted-ssh-key() {
+    ssh-keygen -y -f "$1" < /dev/null >& /dev/null
+    if [ $? != 0 ]; then
+        return 0
+    else
+        return 1
+    fi
+}
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -69,16 +69,45 @@ vm-check-env() {
         echo "ERROR:"
         return 1
     }
-    if [ ! -e "${HOME}/.ssh/id_rsa.pub" ]; then
+    if [ ! -e "$SSH_KEY".pub ]; then
         echo "ERROR:"
         echo "ERROR: environment check failed:"
-        echo "ERROR:   id_rsa.pub SSH public key not found (but govm needs it)."
+        echo "ERROR:   $SSH_KEY.pub SSH public key not found (but govm needs it)."
         echo "ERROR:"
         echo "ERROR: You can generate it using the following command:"
         echo "ERROR:"
         echo "ERROR:     ssh-keygen"
         echo "ERROR:"
         return 1
+    fi
+    if [ -n "$SSH_AUTH_SOCK" ] && [ -e "$SSH_AUTH_SOCK" ]; then
+        if ! ssh-add -l | grep -q "$(ssh-keygen -l -f "$SSH_KEY" < /dev/null 2>/dev/null)"; then
+            if ! ssh-add "$SSH_KEY" < /dev/null; then
+                echo "ERROR:"
+                echo "ERROR: environment setup failed:"
+                echo "ERROR:   Failed to load $SSH_KEY SSH key to agent."
+                echo "ERROR:"
+                echo "ERROR: Please make sure an SSH agent is running, then"
+                echo "ERROR: try loading the key using the following command:"
+                echo "ERROR:"
+                echo "ERROR:     ssh-add $SSH_KEY"
+                echo "ERROR:"
+                return 1
+            fi
+        fi
+    else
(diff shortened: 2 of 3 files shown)


demo/lib/vm.bash
```diff
@@ -69,17 +69,31 @@ vm-check-env() {
...
-        echo "ERROR:   id_rsa.pub SSH public key not found (but govm needs it)."
+        echo "ERROR:   $SSH_KEY.pub SSH public key not found (but govm needs it)."
         echo "ERROR:"
         echo "ERROR: You can generate it using the following command:"
         echo "ERROR:"
         echo "ERROR:     ssh-keygen"
         echo "ERROR:"
         return 1
     fi
+    if ! ssh-add -l | grep -q "$(SSH_ASKPASS=/bin/false ssh-keygen -l -f "$SSH_KEY" < /dev/null 2>/dev/null)"; then
```
kad, line 83: before adding, it might be good to ensure that SSH agent is running/set up.
Not all environment have it.

askervin: I agree, this should probably depend on `[ -n "$SSH_AUTH_SOCK"]`.

Another thing to consider/comment, almost related to this PR: @klihub, @kad, what would you think if we'd reuse existing ssh socket for all ssh connections to the VM? We could store the socket file to govm's working directory for the VM. This speeds up test execution. Once we know the VM name, we could do something like:
```
     echo "# VM SSH server  : ssh $VM_SSH_USER@$VM_IP"

+    if [ -d "$HOME/vms/data/$VM_NAME" ]; then
+        SSH_OPTS="$SSH_OPTS -o ControlMaster=auto -o ControlPath=$HOME/vms/data/$VM_NAME/ssh -o ControlPersist=30"
+        SSH="${SSH%% *} $SSH_OPTS"
+        export SSH SSH_OPTS
+    fi
```

(Or perhaps even better, we could use '%h` in ControlPath so that we would not need to change `SSH` on the fly...)

klihub: > before adding, it might be good to ensure that SSH agent is running/set up.
> Not all environment have it.

I put in -n + -e $SSH_AGENT check, and added an else branch to complain about encrypted keys
if we have no agent.

klihub: > Another thing to consider/comment, almost related to this PR: @klihub, @kad, what would you think if we'd reuse existing ssh socket for all ssh connections to the VM? We could store the socket file to govm's working directory for the VM. This speeds up test execution. Once we know the VM name, we could do something like:

I have no experience in using ssh with session multiplexing. All things considered I am a bit doubtful about
any meaningful speed-ups, but if you think it's worth doing then I'm fine with it.

askervin (commented):
@klihub, I'd still suggest minor changes to make this run quiet and never prompt for password.

demo/lib/host.bash
```diff
@@ -197,3 +197,12 @@ host-delete-vm() {
         command-error "deleting govm \"$VM_NAME\" failed"
     }
 }
+
+host-is-encrypted-ssh-key() {
+    SSH_ASKPASS=/bin/false ssh-keygen -y -f "$1" < /dev/null 2>/dev/null
```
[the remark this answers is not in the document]
klihub: > * replacing `2>/dev/null` by `&>/dev/null`/ would clean up ssh key being printed from test startup.

Yes, that makes sense in `host-is-encrypted-key()`. I'll update.

> * regardless of `SSH_ASKPASS` and `< /dev/null`, `ssh-keygen` keeps asking password when `DISPLAY` is not set.
> * Possible solutions to make sure `ssh-keygen` and `ssh-add` will never ask passwords:
>   
>   * `export SSH_ASKPASS=/bin/false SSH_ASKPASS_REQUIRE=force` -- maybe we could have these in the beginning of `command.bash`?

This only works with pretty recent versions of `ssh` which do recognize the `SSH_ASKPASS_REQUIRE` environment variable. And then with such versions it would make more sense to use `SSH_ASKPASS_REQUIRE=never`.

But all things considered, I think the best is to just force DISPLAY to a bogus value instead. So something like
```
DISPLAY=none SSH_ASKPASS=/bin/false ssh-keygen -y -f $SSH_KEY < /dev/null
```
should do the trick for all current versions.

marquiz (approved):
Sharp-eyed review from @askervin . With the latest changes, looks good to me
LGTM

askervin (approved):
LGTM

PR #475, opened 2020-10-22
static-pools: drop all cmdline flags

marquiz:
Drop all command line flags specific to the static-pools policy. Add new
configuration options for legacy pool config filesystem paths.

After this patch all settings are taken from the configuration system,
making configuration simpler to maintain and understand.

Add examples of all config options to the sample configmap.


klihub (approved):
LGTM.

PR #476, opened 2020-10-22
static-pools: take only pools from legacy config

marquiz:
When reading legacy config (dir or file) from the filesystem only change
the pools in the active configuration. Do not override other config
options which are specified via the dynamic configuration system.

Based on top of #475

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/static-pools/config.go
+++ b/pkg/cri/resource-manager/policy/builtin/static-pools/config.go
@@ -27,6 +27,8 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+type pools map[string]poolConfig
+
 type cpuList struct {
 	Socket     uint64
 	Cpuset     string // TODO: might want to use cpuset from kubelet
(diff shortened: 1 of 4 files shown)


pkg/cri/resource-manager/policy/builtin/static-pools/config.go
```diff
@@ -81,17 +83,17 @@ var (
...
-func parseConfData(raw []byte) (*conf, error) {
+func parseConfData(raw []byte) (pools, error) {
 	conf := &conf{}
 
 	err := yaml.Unmarshal(raw, &conf)
 	if err != nil {
 		return nil, stpError("Failed to parse config file: %v", err)
 	}
-	return conf, nil
+	return conf.Pools, nil
```
klihub, line 93: With this implementation in place, isn't the behavior now such that `parseConfFile()` silently ignores extra configuration data which is present in the file but should only be given via global configuration (ConfigMap, fallback or forced configuration file) ?

If it is, wouldn't this be easily fixed by changing `parseConfData()` to internally use `pools{}` for unmarshalling instead of a full `&conf{}` ?

marquiz: Json/yaml parsing always silently ignores extra keys. But yes, I was thinking about this, too, but left it like this for backwards compatibility. I don't know if anyone is really using this so maybe we should just change it as you described. WDYT?

klihub: What I meant is that now you explicitly parse the data as `&conf{}` but then only take the `.Pool` subset from it, explicitly ignoring yourself silently anything else that might have been given by the user. This is a bit different than the user feeding non-existent fields which are then ignored by the YAML unmarshaller. My suggestion was to instead explicitly unmarshal the data as `pools{}` instead. I think that would be structurally different enough from a `conf{}`, that data for the latter would hopefully not pass the YAML unmarshaller as an instance of the former.

marquiz: Yeah, that's true, I was cutting some corners there. I changed it now.

klihub: Hmm, then I don't understand something. It still looks to me as though `parseConfData()` would unmarshal into a full `&conf{}`, which AFAICT allows the user to put a full configuration in the CMK-specific config file, with all other parts than conf.Pools getting silently ignored.

klihub (approved):

PR #478, opened 2020-10-23
static-pools: re-load legacy config on config update

marquiz:
Change the handling of pools configuration.

First, respect the legacy config path (ConfDirPath, ConfFilePath)
specified in the dynamic configuration. If these settings change we
try to (re-)load the legacy config from the given location.

Second, the pools config priority should now be more consistent and
evident. The priority in descending order is:
1. dynamic configuration
2. leagcy config file
3. legacy config directory structure

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
@@ -90,33 +90,7 @@ func CreateStpPolicy(opts *policy.BackendOptions) policy.Backend {
 
 	stp.Info("creating policy...")
 
-	// Read legacy pools configuration
-	if len(cfg.ConfDirPath) > 0 {
-		p, err := readConfDir(cfg.ConfDirPath)
-		if err != nil {
-			stp.Warn("failed to read configuration directory: %v", err)
-		} else {
-			cfg.Pools = p
-		}
-	}
-	if len(cfg.ConfFilePath) > 0 {
-		p, err := readConfFile(cfg.ConfFilePath)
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
```diff
@@ -136,7 +110,10 @@ func (stp *stp) Start(add []cache.Container, del []cache.Container) error {
 	var err error
 
 	if stp.conf == nil {
-		return stpError("cannot start without any configuration")
+		err := stp.setConfig(cfg)
+		if err != nil {
+			return err
```
klihub, line 115: Out of curiosity, is there a particular reason you 1) don't use the pre-declared `err` in this branch of code (I can't see any apparent reason based on the control flow), and 2) you don't use a combined assign-and-test here while for an almost identical piece of code you do use that later in `configNotify()`. IOW, why not also have this here:
```
    if err = stp.setConfig(cfg); err != nil {
        return err
    }
```

klihub (approved):

klihub (changes requested):
Apart from one minor nit, LGTM.

pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
marquiz: Well spotted, thanks. Bad style, caused by blindly copying code around. Now fixed. Also changed the whole function to use combined assign-and-test in unified way

klihub (approved):
LGTM.

PR #479, opened 2020-10-23
config: use defaults for unspecified values on config update

marquiz:
Unspecified values should be treated as defaults.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -259,22 +259,29 @@ func (m *Module) apply(cfg Data) error {
 
 	log.Debug("module %s: applying module configuration...", m.path)
 
-	if len(cfg) == 0 {
-		defcfg, err := DataFromObject(m.getdefault())
-		if err != nil {
-			return configError("module %s: failed to use module defaults: %v", m.path, err)
-		}
-		cfg = defcfg
-	}
-
-	raw, err := yaml.Marshal(cfg)
+	// First, reset module config to defaults
+	defcfg, err := DataFromObject(m.getdefault())
 	if err != nil {
-		return configError("module %s: failed to marshal configuration: %v", m.path, err)
+		return configError("module %s: failed to retrieve default configuration: %v", m.path, err)
+	}
+	raw, err := yaml.Marshal(defcfg)
+	if err != nil {
+		return configError("module %s: failed to marshal default configuration: %v", m.path, err)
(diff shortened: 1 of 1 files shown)


klihub (approved):

klihub (changes requested):
Let's check what's causing those failures.

marquiz:
> Let's check what's causing those failures.

Yup, need to investigate that

pkg/config/config.go
```diff
@@ -259,13 +259,11 @@ func (m *Module) apply(cfg Data) error {
...
-		defcfg, err := DataFromObject(m.getdefault())
-		if err != nil {
-			return configError("module %s: failed to use module defaults: %v", m.path, err)
-		}
-		cfg = defcfg
+	defcfg, err := DataFromObject(m.getdefault())
+	if err != nil {
+		return configError("module %s: failed to use module defaults: %v", m.path, err)
 	}
+	cfg = defcfg
```
klihub, line 266: Hmm... AFAICT, this now basically always discards the real configuration and instead applies the default one. That explains the test failures: with the default `null` policy we don't track containers, so we always end up with 0 containers in the cache, exactly what the test failure complains about.

Something like this is probably closer to what you meant to do:
```
...
	defcfg, err := DataFromObject(m.getdefault())
	if err != nil {
		return configError("module %s: failed to use module defaults: %v", m.path, err)
	}
	raw, err := yaml.Marshal(defcfg)
	if err != nil {
		return configError("module %s: failed to marshal default configuration: %v", m.path, err)
	}
	if err = yaml.Unmarshal(raw, m.ptr); err != nil {
		return configError("module %s: failed to pre-reset to default configuration: %v", m.path, err)
	}

        if len(cfg) > 0 {
	    raw, err = yaml.Marshal(cfg)
	    if err != nil {
		    return configError("module %s: failed to marshal configuration: %v", m.path, err)
	    }
	    if err = yaml.Unmarshal(raw, m.ptr); err != nil {
		    return configError("module %s: failed to apply configuration: %v", m.path, err)
	    }
        }
...
```

marquiz: Exactly! Re-reading `pkg/config` code, that looks correct. My code was utterly broken 😬  Now fixed. Hopefully the tests pass, now

klihub (approved):

PR #480, opened 2020-10-23
e2e: fix vm-pipe-to-file

askervin:
Fixes regression caused by double-quoting `$append` in commit a60565a69b4a2e548198a699cdbe63bd069b6ab2.

This causes failing proxy env setup when setting up new VM.

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -370,7 +370,7 @@ vm-pipe-to-file() { # script API
         shift
     fi
     cat > "$tmp"
-    vm-put-file --cleanup "$append" "$tmp" "$1"
+    vm-put-file --cleanup $append "$tmp" "$1"
 }
 
 vm-sed-file() { # script API


demo/lib/vm.bash
```diff
@@ -365,12 +365,12 @@ vm-pipe-to-file() { # script API
     #   echo 'Ahoy, Matey...' | vm-pipe-to-file /etc/motd
     local tmp append
     tmp="$(mktemp vm-pipe-to-file.XXXXXX)"
-    if [ "$1" = "--append" ]; then
+    if [ "$1" == "--append" ]; then
```
marquiz, line 368: Why this change to double equal signs?

askervin: That seems to be a bad habit - even man page says that `=` is to be used with `[`. Thanks!

askervin: Fixed

marquiz (approved):
lgtm

demo/lib/vm.bash
marquiz: I think both work in bash. But single equal sign is the POSIX standard or smth

PR #483, opened 2020-10-23
static-pools: lazier node updates

marquiz:
Handle node update in a separate goroutine. With this patch node updates
no longer blocks policy startup and the node object gets updated when
agent connection becomes available.

In addition, node updates are now re-triggered on every configuration
update, making it possible to dynamically change node labeling/tainting
and the extended resources.

Based on top of #478

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
@@ -90,33 +90,7 @@ func CreateStpPolicy(opts *policy.BackendOptions) policy.Backend {
 
 	stp.Info("creating policy...")
 
-	// Read legacy pools configuration
-	if len(cfg.ConfDirPath) > 0 {
-		p, err := readConfDir(cfg.ConfDirPath)
-		if err != nil {
-			stp.Warn("failed to read configuration directory: %v", err)
-		} else {
-			cfg.Pools = p
-		}
-	}
-	if len(cfg.ConfFilePath) > 0 {
-		p, err := readConfFile(cfg.ConfFilePath)
-		if err != nil {
-			stp.Warn("failed to read configuration file: %v", err)
-		} else {
-			if cfg.Pools != nil || len(cfg.Pools) > 0 {
-				stp.Info("Overriding pool configuration from %q with configuration from %q",
-					cfg.ConfDirPath, cfg.ConfFilePath)
-			}
-			cfg.Pools = p
-		}
-	}
-
-	stp.conf = cfg
-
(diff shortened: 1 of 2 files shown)


pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
```diff
@@ -135,13 +108,15 @@ func (stp *stp) Description() string {
 func (stp *stp) Start(add []cache.Container, del []cache.Container) error {
 	var err error
 
-	if stp.conf == nil {
-		return stpError("cannot start without any configuration")
+	if err := stp.nodeUpdater.start(); err != nil {
```
klihub, line 111: nitpick: you don't need a new error instance here.

pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
```diff
@@ -135,13 +108,15 @@ func (stp *stp) Description() string {
...
 	}
 
-	err = stp.updateNode(*stp.conf)
-	if err != nil {
-		stp.Fatal("%v", err)
+	if stp.conf == nil {
+		err := stp.setConfig(cfg)
+		if err != nil {
+			return err
+		}
```
klihub, line 119: Wondering about same as in [478](https://github.com/intel/cri-resource-manager/pull/478#discussion_r510990744).

askervin (commented):
Extra +1 for improved usability :)

I updated PR #472 to start using `static-policy` without Kubernetes (earlier the cluster was built using the `none` policy). Proxying works fine now: `kubeadm init` sets up Kubernetes without problems.

Test results look good, too:

```
$ ./run_tests.sh policies.test-suite/static-pools
...
Tests summary:
PASS static-pools/n4c16/test00-node-status
PASS static-pools/n4c16/test01-exclusive-cores
PASS static-pools/n4c16/test02-mixed-pods
```

klihub (changes requested):
Apart from minor nits, LGTM.

pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
marquiz: Changed now (in #478) so that the pre-defined err is removed and we only use combined assign-condition if statements

pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
marquiz: Fixed (in #478)

klihub (approved):
LGTM.

PR #486, opened 2020-10-26
e2e: store VM images and yamls under VM work dirs

askervin:
Enables running E2E tests from different working directories and still
using the same, once downloaded VM image files.


klihub (approved):
Looks fine.

PR #488, opened 2020-10-27
static-pools: look for _cmk_ isolate when parsing container commandline

marquiz:


pkg/cri/resource-manager/policy/builtin/static-pools/stp-policy.go
```diff
@@ -413,7 +413,7 @@ func parseCmkCmdline(args []string) *cmkLegacyArgs {
 	// Args that we're not really interested in
 	_ = cmkCmd.String("conf-dir", "", "CMK configuration directory")
 
-	if len(args) > 1 && args[1] == "isolate" {
+	if len(args) > 1 && args[0] == "cmk" && args[1] == "isolate" {
```
kad, line 416: `> 2` maybe?

marquiz: Hmm, I think `>1` is more future proof. Run argument parser even if args would be an empty list, like it's usually done. No modifications would be needed if `--pool`argument would get a default value, for example.

klihub (approved):
LGTM.

askervin:
This patch works great, passes all tests 00-05 in PR #473.

PR #490, opened 2020-10-27
rdt: use goresctrl

marquiz:
Drop the internal RDT low level functionality in favor of external goresctrl library.


klihub (commented):
This in itself looks fine, but I think we need to review this together with goresctrl itself.

pkg/cri/resource-manager/control/rdt/rdt.go
```diff
@@ -223,6 +240,11 @@ func rdtError(format string, args ...interface{}) error {
 // Register us as a controller.
 func init() {
 	control.Register(RDTController, "RDT controller", getRDTController())
-	config.GetModule(rdt.ConfigModuleName).AddNotify(getRDTController().configNotify)
+	config.Register(ConfigModuleName, "RDT control", getRDTController().opt, getRDTController().defaultOptions)
+	config.GetModule(ConfigModuleName).AddNotify(getRDTController().configNotify)
```
klihub, line 244: nitpick: you can specify the notification callback already during registration like this:
```
config.Register(ConfigModuleName, "RDT control", getRDTController().opt, getRDTController().defaultOptions,
        config.WithNotify(getRDTController().configNotify))
``` 
I think this should be the preferred way of registering a callback within the same package, unless you have to use a method of an object which has not been instantiated yet at the time of registration (which, for instance, is usually the case for policies).

marquiz: Ach, I totally messed up with the different versions of the PR 😕 Now, I think I managed to push the right version which actually works.

In this, we really need to separate `Register()` and `AddNotify()`. We cannot add the notifier before the RDT adaptation (goresctrl) has been initialized and is ready to accept config changes.

marquiz:
Removed the RFC status. This now works, but, I'm not sure if we want to make the change, yet

klihub:
I tested this with #497 + a commit that sets an rdt-specific logger. Everything looks to be ok.

go.mod
```diff
@@ -7,27 +7,28 @@ require (
 	contrib.go.opencensus.io/exporter/prometheus v0.1.1-0.20191218042359-6151c48ac7fa
 	github.com/cilium/ebpf v0.0.0-20200317182658-6f632cc9ee38
 	github.com/evanphx/json-patch v4.2.0+incompatible
-	github.com/golang/protobuf v1.3.2
-	github.com/google/go-cmp v0.3.1
+	github.com/golang/protobuf v1.4.3
+	github.com/google/go-cmp v0.4.0
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/intel/cri-resource-manager/pkg/topology v0.0.0
+	github.com/marquiz/goresctrl v0.0.0-20201110110642-51d677608af3
```
klihub, line 14: Let's update this to `github.com/marquiz/goresctrl v0.0.0-20201113093610-2b83b62c0854`. After that we could merge this in principle IMO.

klihub (approved):
LGTM

PR #492, opened 2020-10-29
e2e: prevent trying to use a VM with wrong distro, user and cri

askervin:
- Save static-over-boot information on VM to the data directory.
- Use saved information rather than defaults with an already
  existing VM.

### DIFF
--- a/demo/lib/host.bash
+++ b/demo/lib/host.bash
@@ -18,6 +18,22 @@ host-require-govm() {
     command -v "$GOVM" >/dev/null || error "cannot run govm \"$GOVM\". Check PATH or set GOVM=/path/to/govm."
 }
 
+host-get-vm-config() {
+    if [ -z "$1" ]; then
+        error "can't get VM configuration, name not set"
+    fi
+    VM_NAME="$1"
+    HOST_VM_DATA_DIR="$(eval "echo $HOST_VM_DATA_DIR_TEMPLATE")"
+    VM_DATA_CONFIG="$HOST_VM_DATA_DIR/vm-config"
+    if ! [ -f "$VM_DATA_CONFIG" ]; then
+        return 1
+    fi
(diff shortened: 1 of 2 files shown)


askervin:
This solves the problem of like:
```
distro=centos-7 vm=delme test/e2e/run.sh interactive
...
vm=delme test/e2e/run.sh interactive
```
=> cannot login to VM with ssh because the default distro is ubuntu-20.04 and thus VM_SSH_USER=ubuntu

demo/lib/host.bash
```diff
@@ -18,6 +18,22 @@ host-require-govm() {
...
+        error "can't get VM configuration, name not set"
+    fi
+    VM_NAME="$1"
+    HOST_VM_DATA_DIR="$(eval "echo $HOST_VM_DATA_DIR_TEMPLATE")"
+    VM_DATA_CONFIG="$HOST_VM_DATA_DIR/vm-config"
+    if ! [ -f "$VM_DATA_CONFIG" ]; then
+        return 1
+    fi
+    source "$VM_DATA_CONFIG"
+    if [ -z "$VM_DISTRO" ] || [ -z "$VM_CRI" ] || [ -z "$VM_SSH_USER" ]; then
```
marquiz, line 32: For consistently/completeness, should we check for VM_NAME, too?

askervin: Thanks, I agree it's better when complete, as `VM_NAME` is saved to the config file.

Little things that were and still are difficult to decide: should `VM_NAME` be in the config or not? This is why I struggled checking it, too. Anyway, it makes sense to check everything we have there.

marquiz: > should `VM_NAME` be in the config or not?

I was pondering the same thing, but, have no idea 😄  I think this good, now

marquiz (approved):
Looks Good To Me

PR #495, opened 2020-10-30
fixes: allow test to omit installing/launching agent.

klihub:
Add `omit_agent` for allowing test cases to run without installing, setting up and launching `cri-resmgr-agent`. Defaults to `0`. Setting `omit_agent=1` omits the agent. Packaging tests are a prime example of when this makes a lot of sense.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -50,6 +50,7 @@ usage() {
     echo "    reinstall_cri_resmgr: If 1, stop running cri-resmgr, reinstall,"
     echo "             and restart it on the VM before starting test run."
     echo "             The default is 0."
+    echo "    omit_agent: if 1, omit installing/starting/cleaning up cri-resmgr-agent."
     echo "    outdir:  Save output under given directory."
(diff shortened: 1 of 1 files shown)


test/e2e/run.sh
```diff
@@ -50,6 +50,7 @@ usage() {
     echo "    reinstall_cri_resmgr: If 1, stop running cri-resmgr, reinstall,"
     echo "             and restart it on the VM before starting test run."
     echo "             The default is 0."
+    echo "    omit_agent: if 1, omit installing/starting/cleaning up cri-resmgr-agent."
```
marquiz, line 53: Hmm, I would just use non-empty value instead of  `1`. I.e. you don't need to remember what the required values is, only that `omit_agent` is the name of the variable.

klihub: Me too. But that would not be consistent with the rest of the similar conventions @askervin put in place. For better consistency, I went with the established conventions.

marquiz: Ah, true. Let's go with this 👍 

LGTM

marquiz (approved):

PR #497, opened 2020-11-09
pkg/log: switch logger to use klog.

klihub:


pkg/log/klogcontrol/klogcontrol.go
```diff
@@ -0,0 +1,186 @@
...
+// klogflag wraps a klog flag for configuration.
+type klogflag struct {
+	flag *flag.Flag
+}
+
+// Set implements flag.Value.Set() for wrapped klog flags.
+func (klogf *klogflag) Set(value string) error {
+	if klogf.flag.Name == "stderrthreshold" {
+		value = strings.ToUpper(value)
+	}
```
marquiz, line 106: What is this for?

marquiz (commented):
Only managed to review partly because of familiy-life disturbances. Will continue tomorrow 😄

pkg/log/klogcontrol/klogcontrol.go
klihub: klog requires you to shout the logging threshold/level in ALL CAPS. I prefer to accept less aggressively formatted configuration, too. With this both (and admittedly all other combinations) are accepted.

pkg/log/grpc-logger.go
```diff
@@ -0,0 +1,94 @@
...
+func (g grpclogger) Warning(args ...interface{}) {
+	g.Logger.Warn("%s", fmt.Sprint(args...))
+}
+
+func (g grpclogger) Warningln(args ...interface{}) {
+	g.Logger.Warn("%s", fmt.Sprint(args...))
+}
+
+func (g grpclogger) Warningf(format string, args ...interface{}) {
+	g.Logger.Warn("%s", fmt.Sprintf(format, args...))
```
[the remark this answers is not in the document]
klihub: Good point. That's more straightforward and saves an extra expensive `Sprintf()`. I'll update.

pkg/log/grpc-logger.go
```diff
@@ -0,0 +1,94 @@
...
+	g.Logger.Debug("%s", fmt.Sprint(args...))
+}
+
+func (g grpclogger) Infoln(args ...interface{}) {
+	g.Logger.Debug("%s", fmt.Sprint(args...))
+}
+
+func (g grpclogger) Infof(format string, args ...interface{}) {
+	g.Logger.Debug("%s", fmt.Sprintf(format, args...))
+}
```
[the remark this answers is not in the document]
klihub: It's more the nature/semantics than the volume of messages. Anything below warnings from grpc really looks like debug messages to me. Since debugging messages we can still control individually per logger source, if you need to see the ones from gRPC, you can do so without necessarily/potentially getting flooded by a ton of others.

pkg/log/klogcontrol/klogcontrol.go
marquiz: K, make sense. Maybe a short comment in the code would be in order because this just looks like a strange "anomaly". Or maybe it's just me. WDYT?

pkg/log/log.go
```diff
@@ -402,7 +393,15 @@ func (l logger) Panic(format string, args ...interface{}) {
 }
 
 func (l logger) DebugBlock(prefix string, format string, args ...interface{}) {
-	l.block(LevelDebug, prefix, format, args...)
+	log.RLock()
+	suppress := !log.forced
+	if !suppress {
+		_, suppress = log.debug[l]
+	}
```
marquiz: What is the logic here 🤔  If global forced debug logging is turned on we take `suppress` from the individual debug setting??

pkg/log/grpc-logger.go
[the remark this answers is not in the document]
marquiz: Sounds reasonable, thanks

pkg/log/log.go
klihub: I just plain broke the logic in a last minute change that was supposed to be too small to go wrong... Thanks for spotting.

pkg/log/grpc-logger.go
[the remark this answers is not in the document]
klihub: Updated.

pkg/log/klogcontrol/klogcontrol.go
klihub: Added a comment explaining the change to uppercase.

marquiz (approved):
Thanks @klihub ! Looks good to me, now

kad (dismissed):
lgtm

askervin:
Verified with e2e tests, this is good to go.

askervin (approved):
LGTM

go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	k8s.io/apimachinery v0.17.2
 	k8s.io/client-go v0.17.2
 	k8s.io/cri-api v0.0.0
+	k8s.io/klog/v2 v2.0.0
```
mythi, line 29: not a big deal but the latest tag is `v2.4.0`.

klihub: Ah, ok. Good point. I'll update and check if everything works.

klihub:
@marquiz One remaining question I have is whether the current handling of the `--logger-debug` command line option (as changed by this PR) is acceptable. Earlier, the code tried to propagate debug options given on the command line both to the default and the active configuration. After this PR, debug options given on the command line act as the default which is only used if nothing else is given in the configuration.

TBH, I'm not sure which one is better. Every time I look at it, I seem to come to a different conclusion, so I suspect neither is self-evident behavior. The current setup is counter-intuitive in the sense that the debug command line does not have the highest precedence: it is only in effect until the first (re-)configuration event is handled... but stays in effect if the new configuration data does not specify debugging flags. But my practical experience when testing was that doing it the other way is also often counter-intuitive, when you specify debugging flags in the configuration and then the behavior does not fully match it because of some debug options given on the command line.

marquiz:
I didn't even realize that this PR changes the behavior of `--logger-debug`.

You're probably right in that there is no self-evident "correct behavior". The interplay between overlapping command-line flags and config file options is a bit involved. However, I still feel that the command-line flags should work as overrides (aimed primarily for develoment purposes) and should take precedence over anything specified in the config yaml. That is, I'm an advocate of the old behavior. We already have `--fallback-config` for providing defaults. Thus, in order to simplify things and avoid confusion, I think that it would be actually better to just drop logger-related command-line flags if we can't/don't want to have them behave as big hammer overrides. WDYT?

In any case, we should probably document those command line flags overlapping with config file options and be consistent with the handling of those. Are there any other than the logging-related `--logger-debug`, `--logger-level` and `--logger-sources`?

klihub:
> I didn't even realize that this PR changes the behavior of `--logger-debug`.
> 
> You're probably right in that there is no self-evident "correct behavior". The interplay between overlapping command-line flags and config file options is a bit involved. However, I still feel that the command-line flags should work as overrides (aimed primarily for develoment purposes) and should take precedence over anything specified in the config yaml. That is, I'm an advocate of the old behavior. We already have `--fallback-config` for providing defaults. Thus, in order to simplify things and avoid confusion, I think that it would be actually better to just drop logger-related command-line flags if we can't/don't want to have them behave as big hammer overrides. WDYT?

Yes, I tend to agree with that conclusion. The biggest reason why I'm a bit reluctant to eliminate `--logger-debug` altogether is because I'd like to be able to enable debugging early, much before configuration from an external source could ever be applied.  That said, I have a patch (so far excluded from this PR) which allows that even before the command line has been parsed, using environment variables similar to how the klog flags can now be set at a very early phase. I think that patch should probably added to this PR. And with `--logger-debug`, either the semantics should be changed back or close to the original, or it should then be removed.

> In any case, we should probably document those command line flags overlapping with config file options and be consistent with the handling of those. Are there any other than the logging-related `--logger-debug`, `--logger-level` and `--logger-sources`?

`--logger-level` and `--logger-sources` have been removed altogether (by this PR). The former is roughly what the klog's `stderrthreshold` controls, so it's been removed to avoid contradiction. But good point, I forgot to update the docs, which I will now do. The latter we have never been really using, favoring instead Debug()-messages for everything that we want to be off by default.

marquiz:
> The biggest reason why I'm a bit reluctant to eliminate `--logger-debug` altogether is because I'd like to be able to enable debugging early, much before configuration from an external source could ever be applied.

Mm, that's a fair point


> That said, I have a patch (so far excluded from this PR) which allows that even before the command line has been parsed, using environment variables similar to how the klog flags can now be set at a very early phase. I think that patch should probably be added to this PR. And with `--logger-debug`, either the semantics should be changed back or close to the original, or it should then be removed.

The env solution sounds nice and good enough for me. I would be in favor of dropping the `--logger-debug` flag for good. Would possibly decrease future maintenance burden and churn, too, no need to get back to this subject anymore 😉 


> `--logger-level` and `--logger-sources` have been removed altogether (by this PR)

Ah, yes of course

marquiz:
How is the env var behaving wrt. the configuration (yaml)? If I read this correctly config changes will take over the env var. Shouldn't it be the other way around?

klihub:
> How is the env var behaving wrt. the configuration (yaml)? If I read this correctly config changes will take over the env var. Shouldn't it be the other way around?

I ended up with the current approach after practical experience with playing around with both priority
setups. If it was the other way around then there was simply no way of changing a configuration that
has been set initially in the environment without shutting down CRI-RM, modifying the environment
and then restarting CRI-RM. That pretty much destroys the idea of dynamic configuration in the first
place and when you run into that, it feels both infuriating and counter-intuitive. So, with the current
setup you can set an initial configuration in the environment, and then still dynamically switch back
and forth between that and any other configuration you see fit without having to stop CRI-RM.

marquiz:
> So, with the current setup you can set an initial configuration in the environment, and then still
> dynamically switch back and forth between that and any other configuration you see fit without
> having to stop CRI-RM. You do have to be careful about what you set in the configuration, but
> at least you don't paint yourself in a corner.

Fair enough, I trust your judgement and experimentation 😄  Should we document this behavior (or is it documented already) in the docs(?)

docs/setup.md
```diff
@@ -317,9 +317,10 @@ to your real kubelet instance's configuration:
 
 ## Logging and Debugging
 
-You can control logging and debugging with the `--logger-*` command line options. By
-default debug logs are globally disabled. You can turn on full debug logs with the
-`--logger-debug '*'` command line option.
+You can control logging with the klog command line options or by setting the
+corresponding `LOGGER_*` environment variables. Additionally, the `LOGGER_DEBUG`
+environment variable controls debug logs. These are globally disabled by default.
+You can turn on full debugging by setting `LOGGER_DEBUG='*'`.
```
marquiz, line 323: A bit out-of-scope of this PR, but, should we mention the dynamic configuration here. I.e. `logger:` options can be changed dynamically at runtime

klihub: Yes, definitely. I'll update the docs.

klihub:
> > So, with the current setup you can set an initial configuration in the environment, and then still
> > dynamically switch back and forth between that and any other configuration you see fit without
> > having to stop CRI-RM. You do have to be careful about what you set in the configuration, but
> > at least you don't paint yourself in a corner.
> 
> Fair enough, I trust your judgement and experimentation
>  Should we document this behavior (or is it documented already) in the docs(?)

Yes, we should document it. I'll update the related bits to include a description/warning about
the behavior.

marquiz (approved):
I'm ready to merge this. Thanks @klihub !

LGTM

docs/setup.md
klihub: Updated the documentation to describe the precedence of the environment vs. configuration files and ConfigMaps.

PR #501, opened 2020-11-09
cross-build: don't ignore useradd errors, pass options as such.

klihub:
When building cross-build container images or cross-building packages
  - don't ignore errors from useradd
  - pass uid option to useradd without extra sed-trickery


marquiz (approved):
Thanks, looks cleaner  👍 
lgtm

PR #503, opened 2020-11-10
e2e: memtier memory-type and coldstart tests

askervin:
The first e2e test for the memtier coldstart feature.

Depends on PR #506

### DIFF
--- a/test/e2e/policies.test-suite/memtier/c4pmem4/test01-pmem-node-assigning/code.var.sh
+++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test01-pmem-node-assigning/code.var.sh
@@ -1,0 +1,18 @@
+# Test that CPU-less PMEM nodes are assigned to closest nodes with CPU.
+
+# Restart cri-resmgr in order to clear logs and make sure assignment
+# is successful with installed cri-resmgr.
+terminate cri-resmgr
+launch cri-resmgr
+
+CRI_RESMGR_OUTPUT_COMMAND="cat cri-resmgr.output.txt"
+
+echo "Verify PMEM node assignment to CPU-ful nodes"
+for expected_output in \
+    "PMEM node #4 assigned to #2" \
+    "PMEM node #5 assigned to #3" \
+    "PMEM node #6 assigned to #0" \
+    "PMEM node #7 assigned to #1"; do
+    vm-command "$CRI_RESMGR_OUTPUT_COMMAND | grep -E '$expected_output'" ||
+        command-error "expected PMEM assignment not found"
+done
--- a/test/e2e/policies.test-suite/memtier/c4pmem4/test02-annotation-memory-type/code.var.sh
+++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test02-annotation-memory-type/code.var.sh
@@ -1,0 +1,16 @@
+# Test that container memory is pinned according to memory-type annotation
+
+# pod0c0 runs on the first node, uses only dram
+# pod0c1 runs on the second node, uses only pmem
+# pod0c2 runs on the third node, uses dram+pmem
+# pod0c9 runs on the fourth node, no memory-type restrictions (use both)
+MEM=250M MEMTYPEC0=dram MEMTYPEC1=pmem MEMTYPEC2=pmem,dram create memtype-guaranteed
+
+verify 'cpus["pod0c0"] == {"cpu0"}' \
+       'mems["pod0c0"] == {"node0"}' \
+       'cpus["pod0c1"] == {"cpu1"}' \
+       'mems["pod0c1"] == {"node7"}' \
+       'cpus["pod0c2"] == {"cpu2"}' \
+       'mems["pod0c2"] == {"node2", "node4"}' \
+       'cpus["pod0c9"] == {"cpu3"}' \
+       'mems["pod0c9"] == {"node3", "node5"}'
--- a/test/e2e/policies.test-suite/memtier/c4pmem4/test02-annotation-memory-type/memtype-guaranteed.yaml.in
+++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test02-annotation-memory-type/memtype-guaranteed.yaml.in
@@ -1,0 +1,33 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ${NAME}
+  annotations:
+    cri-resource-manager.intel.com/memory-type: |
+      ${NAME}c0: ${MEMTYPEC0}
+      ${NAME}c1: ${MEMTYPEC1}
+      ${NAME}c2: ${MEMTYPEC2}
+spec:
+  containers:
+    $(for CONT in 0 1 2; do echo "
+    - name: ${NAME}c${CONT}
+      image: busybox
+      command: ['sh', '-c', 'echo ${NAME}c${CONT} \$(sleep inf)']
+      resources:
+        requests:
+          cpu: 500m
+          memory: ${MEM}
+        limits:
+          cpu: 500m
+          memory: ${MEM}
+    "; done)
+    - name: ${NAME}c9
+      image: busybox
+      command: ['sh', '-c', 'echo ${NAME}c9 \$(sleep inf)']
+      resources:
+        requests:
+          cpu: 500m
+          memory: ${MEM}
+        limits:
+          cpu: 500m
+          memory: ${MEM}
--- a/test/e2e/policies.test-suite/memtier/c4pmem4/test03-coldstart/bb-coldstart.yaml.in
+++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test03-coldstart/bb-coldstart.yaml.in
@@ -1,0 +1,29 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ${NAME}
+  annotations:
+    cri-resource-manager.intel.com/memory-type: |
+      ${NAME}c0: dram,pmem
+    cri-resource-manager.intel.com/cold-start: |
+      ${NAME}c0:
+        duration: ${DURATION}
+spec:
+  containers:
+    - name: ${NAME}c0
+      image: busybox
+      command:
+        - sh
+        - -c
+        - 'cold_alloc=\$(dd if=/dev/zero bs=${COLD_ALLOC} count=1 | tr \"\\\0\" \"x\");
+           sh -c \"paused after cold_alloc \\\$(sleep inf)\";
+           warm_alloc=\$(dd if=/dev/zero bs=${WARM_ALLOC} count=1 | tr \"\\\0\" \"x\");
+           sh -c \"paused after warm_alloc \\\$(sleep inf)\";
+           echo ${NAME}c0 \$(sleep inf); # needed for pod resource discovery'
+      resources:
+        requests:
+          cpu: 500m
+          memory: ${MEM}
+        limits:
+          cpu: 500m
+          memory: ${MEM}
--- a/test/e2e/policies.test-suite/memtier/c4pmem4/test03-coldstart/code.var.sh
+++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test03-coldstart/code.var.sh
@@ -1,0 +1,77 @@
+# Test that a cold-started pod...
+# 1. is allowed to allocate memory only from PMEM nodes
+#    during cold period (of length $DURATION).
+# 2. is restricted from the very beginning of pod execution:
+#    immediately allocated memory blob consumes PMEM from expected node.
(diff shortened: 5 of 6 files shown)


askervin:
@klihub , I'm pushing this as RFC already now due to an issue.

When running the test `test/e2e/run_tests.sh test/e2e/policies.test-suite/memtier/pmem/`, I'm getting
```
kubect
Warning  Failed     4m29s (x8 over 5m57s)  kubelet            Error: resource-manager: failed to allocate container resources: memtier: failed to allocate resourc
es for pod0:pod0c0: memtier: no suitable pool found for container pod0:pod0c0
```

yet there should be enough memory:
```
# numactl -H
available: 4 nodes (0-3)
node 0 cpus: 0 1
node 0 size: 3942 MB
node 0 free: 1512 MB
node 1 cpus: 2 3
node 1 size: 4031 MB
node 1 free: 2245 MB
node 2 cpus:
node 2 size: 16127 MB
node 2 free: 16096 MB
node 3 cpus:
node 3 size: 16127 MB
node 3 free: 16096 MB
```

cri-resmgr output says:
```
I: [      sysfs     ] NUMA nodes with CPUs: 0-1
I: [      sysfs     ] NUMA nodes with (any) memory: 0-3
I: [      sysfs     ] NUMA nodes with normal memory: 0-3
I: [      sysfs     ] node 0 has DRAM memory
I: [      sysfs     ] node 1 has DRAM memory
I: [      sysfs     ] node 2 has PMEM memory
I: [      sysfs     ] node 3 has PMEM memory
```

My guessing: later output looks like that node 2 and 3 do not appear in the topology tree - possibly using PMEM is then difficult. How should CPU-less (socket-less) memory nodes look like there?

klihub:
> @klihub , I'm pushing this as RFC already now due to an issue.
> 
> When running the test `test/e2e/run_tests.sh test/e2e/policies.test-suite/memtier/pmem/`, I'm getting
> 
> ```
> kubect
> Warning  Failed     4m29s (x8 over 5m57s)  kubelet            Error: resource-manager: failed to allocate container resources: memtier: failed to allocate resourc
> es for pod0:pod0c0: memtier: no suitable pool found for container pod0:pod0c0
> ```
> 
> yet there should be enough memory:
> 
> ```
> # numactl -H
> available: 4 nodes (0-3)
> node 0 cpus: 0 1
> node 0 size: 3942 MB
> node 0 free: 1512 MB
> node 1 cpus: 2 3
> node 1 size: 4031 MB
> node 1 free: 2245 MB
> node 2 cpus:
> node 2 size: 16127 MB
> node 2 free: 16096 MB
> node 3 cpus:
> node 3 size: 16127 MB
> node 3 free: 16096 MB
> ```
> 
> cri-resmgr output says:
> 
> ```
> I: [      sysfs     ] NUMA nodes with CPUs: 0-1
> I: [      sysfs     ] NUMA nodes with (any) memory: 0-3
> I: [      sysfs     ] NUMA nodes with normal memory: 0-3
> I: [      sysfs     ] node 0 has DRAM memory
> I: [      sysfs     ] node 1 has DRAM memory
> I: [      sysfs     ] node 2 has PMEM memory
> I: [      sysfs     ] node 3 has PMEM memory

> ```

> 
> My guessing: later output looks like that node 2 and 3 do not appear in the topology tree - possibly using PMEM is then difficult. How should CPU-less (socket-less) memory nodes look like there?

I think that sysfs should be still fine... we have to look at what memtier says when it builds the pools. The association logic between CPU-less PMEM nodes and the closest DRAM nodes is (IIRC) in memtier, as it is policy-specific.

I'll pull in your branch an take a closer look at what happens...

klihub:
> I think that sysfs should be still fine... we have to look at what memtier says when it builds the pools. The association logic
> between CPU-less PMEM nodes and the closest DRAM nodes is (IIRC) in memtier, as it is policy-specific.

This is how it looks like on real HW:
```
I: [      sysfs     ] node 7 has PMEM memory
I: [      sysfs     ] node 0 has DRAM memory
I: [      sysfs     ] node 1 has DRAM memory
I: [      sysfs     ] node 2 has DRAM memory
I: [      sysfs     ] node 3 has DRAM memory
I: [      sysfs     ] node 4 has PMEM memory
I: [      sysfs     ] node 5 has PMEM memory
I: [      sysfs     ] node 6 has PMEM memory
I: [     policy     ] activating 'memtier' policy...
...
D: [     policy     ] discovering CPU available at node numa node #0...
I: [     policy     ] *** [17 28 28 28 10 28 28 28]: PMEM node 4 assigned to numa node #0
D: [     policy     ] discovering CPU available at node numa node #2...
I: [     policy     ] *** [28 28 17 28 28 10 28 28]: PMEM node 5 assigned to numa node #2
...
D: [     policy     ] discovering CPU available at node numa node #1...
I: [     policy     ] *** [28 17 28 28 28 28 10 28]: PMEM node 6 assigned to numa node #1
D: [     policy     ] discovering CPU available at node numa node #3...
I: [     policy     ] *** [28 28 28 17 28 28 28 10]: PMEM node 7 assigned to numa node #3
...
D: [     policy     ] <pre-start>        <NUMA node #1>
D: [     policy     ] <pre-start>          - <numa node #1 capacity: CPU: isolated:5,61, sharable:1,9,13-57:4,65-109:4 (26000m), MemLimit: dram:33816088576, pmem:133143986176>
D: [     policy     ] <pre-start>          - <numa node #1 allocatable: CPU: isolated:5,61, sharable:1,9,13-57:4,65-109:4 (allocatable:26000m), MemLimit: dram:33816088576, pmem:133143986176>
D: [     policy     ] <pre-start>          - normal memory: 1
D: [     policy     ] <pre-start>          - HBM memory: 
D: [     policy     ] <pre-start>          - PMEM memory: 6
D: [     policy     ] <pre-start>          - parent: <socket #1>
...
```

askervin:
> This is how memtier looks like on real HW...

The problem is in node distances. With the `topology.var.json` in the first commit:
```
[
    {"cores": 1, "mem": "4G", "nodes": 2},
    {"cores": 0, "nvmem": "16G", "nodes": 2}
]
```

topology2qemuopts creates weights where distance between PMEM nodes (2 and 3) is lower than their distance to CPU nodes (0 and 1).

I'll replace that with topology where the first PMEM node is closer to CPU node 0 and the second to CPU node 1:
```
[
    {"mem": "4G", "cores": 1, "dies": 2},
    {"nvmem": "16G", "node-dist": {"0": 17}},
    {"nvmem": "16G", "node-dist": {"1": 17}}
]
```

With this topology I can see `...assigned to numa node...` messages in the output. So the remaining question would be: should we have a warning in the log in case there is a PMEM node that is not assigned to any numa node with CPU?


(I can switch "dies" to "packages" if we'd like to keep tests running with Qemu 4.X... but this would be a nice opportunity to have a single socket multidie topology, too.)

askervin:
@klihub, after walking around the block, would this be a better idea than a warning message: always assign memory of a PMEM node to a pool with a numa node that has
- CPUs
- least distance to the PMEM node, and
- least amount memory assigned memory?

klihub:
> The problem is in node distances. With the `topology.var.json` in the first commit:
...
> topology2qemuopts creates weights where distance between PMEM nodes (2 and 3) is lower than their distance to CPU nodes (0 and 1).

Hmm... is that really a valid reason to fail to associate the PMEM nodes with cpu-ful DRAM nodes ? I think it's not a deliberate/conscious choice, rather a bug in memtier.

> 
> I'll replace that with topology where the first PMEM node is closer to CPU node 0 and the second to CPU node 1:
> 
> ```
> [
>     {"mem": "4G", "cores": 1, "dies": 2},
>     {"nvmem": "16G", "node-dist": {"0": 17}},
>     {"nvmem": "16G", "node-dist": {"1": 17}}
> ]
> ```
> 
> With this topology I can see `...assigned to numa node...` messages in the output. So the remaining question would be: should we have a warning in the log in case there is a PMEM node that is not assigned to any numa node with CPU?

I think we should fix the memtier setup code so that this does not happen...

> 
> (I can switch "dies" to "packages" if we'd like to keep tests running with Qemu 4.X... but this would be a nice opportunity to have a single socket multidie topology, too.)

How about just having a single-socket single-die setup with 2 DRAM + 2 PMEM nodes ?

klihub:
> @klihub, after walking around the block, would this be a better idea than a warning message: always assign memory of a PMEM node to a pool with a numa node that has
> 
> * CPUs
> * least distance to the PMEM node, and
> * least amount memory assigned memory?

I would assign PMEM nodes to a CPU-ful node with the smallest distance if that is unique. If not then I'd probably use the node with the least amount of PMEM assigned.

askervin:
> > * CPUs
> > * least distance to the PMEM node, and
> > * least amount memory assigned memory?
> 
> I would assign PMEM nodes to a CPU-ful node with the smallest distance if that is unique. If not then I'd probably use from those nodes the one with the least amount of PMEM assigned.

I see we have similar thoughts here. I thought going through above bullets in this order, too: if there are > 1 equally good nodes after seen bullets, consider next bullet.

> How about just having a single-socket single-die setup with 2 DRAM + 2 PMEM nodes ?

I switched to this topology - and stepped on a mine in test infra topology reporting. Having that flagged with comments, this actually works pretty well. The only bad that I found is inability to tell if a node has DRAM or PMEM.

klihub:
> I see we have similar thoughts here. I thought going through above bullets in this order, too: if there are > 1 equally good nodes after seen bullets, consider next bullet.

Ah, sorry. My bad. Now that I reread your comment with both of my remaining brain cells turned on, I see that it's exactly the same as what I wrote...

> 
> > How about just having a single-socket single-die setup with 2 DRAM + 2 PMEM nodes ?
> 
> I switched to this topology - and stepped on a mine in test infra topology reporting. Having that flagged with comments, this actually works pretty well. The only bad that I found is inability to tell if a node has DRAM or PMEM.

Maybe we should either dump this to the cache, or preferably make it available via the introspection
interface (which is now disabled by default).

We already have info about all pods, containers and system topology available there, as well as some
policy-specific data. And it was exactly meant for external consumption. Of course, it seems to lack
some of the recent topology-related data, including dies, and NUMA memory type. Those should be
added anyway regardless of this PR.

Then you could retrieve the World According to Garp^H^H^H^HCRI-RM as JSON from the introspection
(HTTP) interface and pick out the necessary bits about NUMA nodes. The test case could short-circuit
itself to success, with a BIG FAT WARNING, if the interface is not available so the case can't be run. Or
probably we should add a third test case status: SKIPPED for cases that can't be run with the current
HW/CRI-RM combo setup.

klihub:
> > With this topology I can see `...assigned to numa node...` messages in the output. So the remaining question would be: should we have a warning in the log in case there is a PMEM node that is not assigned to any numa node with CPU?
> 
> I think we should fix the memtier setup code so that this does not happen...

We could try something like [this](https://github.com/klihub/cri-resource-manager/commit/a8e6cf65494a1a0bb28f8732395afb757f464d7b)...
Completely untested ATM. Will test and fix tomorrow.

askervin:
@klihub, taking the introspection API into use sounds good.

askervin:
@klihub, I added a test to see all non-empty combinations of `memory-type` working on a multicontainer pod case. That works fine (run on debian-sid and ubuntu 20.04).

All dependencies are merged => this patch is now based on current master.

I'm still puzzled on this error that keeps haunting on my vm:
```
distro=debian-sid ./run_tests.sh policies.test-suite/memtier/c4pmem4
...
root@vm> cat cri-resmgr.output.txt | grep 'pinning to memory 0,6'
D: [     policy     ]   => pinning to memory 0,6
### Verifying assertion 'cores["pod0c0"] == {"node0/core0"}'

### The assertion holds.

### Verifying assertion 'mems["pod0c0"] == {"node0", "node6"}'

Traceback (most recent call last):
  File "/home/ask/go/src/github.com/intel/cri-resource-manager/test/e2e/policies.test-suite/memtier/c4pmem4/test02-coldstart/output/pyexec.py", line 5, in <module>
    assert(mems["pod0c0"] == {"node0", "node6"})
AssertionError
```

No error on Ubuntu 20.04.

I'll debug deeper into this to see on what could be the issue. Needs to be solved before merging.

klihub:
@askervin That looks like a `runc`-related problem on sid. I see `UpdateContainerResources()` requests failing with runc errors. Replacing the stock distro `runc` with a self-compiled one works with the latest master/HEAD. Also anything after rc91 or anything after merged PR 2254/9167393cb8b6da7852157bc71960ecc7f1dc84f9 seems to work fine.

askervin:
Great finding, @klihub 🥇

askervin:
I'll finalize this patch with this test content and remove the draft status. I'll write dynamic demotion tests to another patch.

klihub (approved):
LGTM.

PR #514, opened 2020-11-18
e2e: catch last cri-resmgr output line if process disappeared

askervin:
This helped catching cri-resmgr config errors that blocked launching it, as they did not appear as FATAL ERRORs in the output. The latter are already catched right above this addition in the code.


klihub (approved):

marquiz (approved):
looks good to me

PR #516, opened 2020-11-19
e2e: test that memtier honors AvailableResources

askervin:


askervin:
These tests fail right now. Let's not merge yet...

klihub:
ReservedResources is not handled properly by memtier at the moment. cpusets in AllowedResources should be honored with [this patch in place](https://github.com/klihub/cri-resource-manager/commit/ee28ec278d6cf97ddcca3ffa89e6bf9869e78075).

askervin:
@klihub, @kad, @marquiz, I changed this PR from adding e2e tests into adding ReservedResources support to memtier and fixing shared allocation limits - otherwise the tests wouldn't pass.

A note from test runs: It's fine to request `CPU=1250m` for the last pod in the test, because there is enough CPUs left in node 1 (2 CPUs, 750m of which reserved to cri-resmgr in the config). However, when the pod is running, it's allowed to run only on single CPU despite it requested 1.25. Would this be correct behavior?

pkg/cri/resource-manager/policy/builtin/memtier/node.go
```diff
@@ -451,7 +451,7 @@ func (n *numanode) GetPhysicalNodeIDs() []system.ID {
 func (n *numanode) DiscoverSupply() Supply {
 	log.Debug("discovering CPU available at node %s...", n.Name())
 
-	noderes := n.sysnode.CPUSet().Intersection(n.policy.allowed)
+	noderes := n.sysnode.CPUSet().Intersection(n.policy.allowed).Difference(n.policy.reserved)
```
klihub, line 454: I don't think we can do it like this. With this implementation, `ReservedResources[CPU]` would be just another way of expressing `AllowedResources[CPU]`, by using negation. However, the semantics of `ReservedResources` should be 'resources reserved for consumption by containers in the `kube-system` namespace.`. So we should allocate `kube-system` containers to those as opposed to the current approach of just pushing them to the root pool.

pkg/cri/resource-manager/policy/builtin/memtier/node.go
```diff
@@ -595,7 +595,7 @@ func (n *dienode) DiscoverSupply() Supply {
 				log.Error("node has an unknown memory type/combination")
 			}
 		}
-		diecpus := n.syspkg.DieCPUSet(n.id).Intersection(n.policy.allowed)
+		diecpus := n.syspkg.DieCPUSet(n.id).Intersection(n.policy.allowed).Difference(n.policy.reserved)
```
klihub, line 598: Same applies here...

pkg/cri/resource-manager/policy/builtin/memtier/node.go
```diff
@@ -714,7 +714,7 @@ func (n *socketnode) DiscoverSupply() Supply {
 				log.Error("node has an unknown memory type/combination")
 			}
 		}
-		sockcpus := n.syspkg.CPUSet().Intersection(n.policy.allowed)
+		sockcpus := n.syspkg.CPUSet().Intersection(n.policy.allowed).Difference(n.policy.reserved)
```
klihub, line 717: Same applies here...

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -774,10 +774,10 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
 
 	// 1) a node with insufficient isolated or shared capacity loses
 	switch {
-	case (isolated2 < 0 && isolated1 >= 0) || (shared2 <= 0 && shared1 > 0):
+	case (isolated2 < 0 && isolated1 >= 0) || (shared2 < 0 && shared1 >= 0):
```
klihub, line 777: Hmm... these score comparison changes together with the subsequent change in `Allocate()` basically revert 6d7ac67e3c211faf41c5a8a0a3b1a2e4cf3db3ed, which was put in place to explicitly disallow fully exhausting
the shared subset of a pool. I wonder if that really was/is an unnecessarily strict restriction (because TBH, I
can't recall the exact details right now when/how that caused breakage).

klihub:
> @klihub, @kad, @marquiz, I changed this PR from adding e2e tests into adding ReservedResources support to memtier and fixing shared allocation limits - otherwise the tests wouldn't pass.
> 
> A note from test runs: It's fine to request `CPU=1250m` for the last pod in the test, because there is enough CPUs left in node 1 (2 CPUs, 750m of which reserved to cri-resmgr in the config). However, when the pod is running, it's allowed to run only on single CPU despite it requested 1.25. Would this be correct behavior?

This is something that I've already meant to change the default behavior for. By default, for fractional CPU requests in the range of 1 < request < 2, we allocate 1 exclusive CPU + the remaining fraction of the shared pool by default, unless the container is annotated to opt out from this. The target was containers that have/run one dedicated latency-sensitive process plus some small other non-sensitive generic one. I think the default should be opt-in and not opt-out, because the processes anyway need to be aware of this setup and they need to pin themselves to the correct subset of their assigned cpuset (advertised in the bind-mounted resources.sh file).

That said, it is still wrong to have just one CPU in the cpuset and if that happens, I guess it is the exclusive one. Instead the container's cpuset should container the shared subset of the pool + the allocated exclusive CPU. And (a bit dangerously) the CFS shares should be set to the equivalent of the fractional part of the CPU request. That's how much the container should be allowed to consume from the shared set, while it should have exclusive access to the remaining one CPU (although we can only try to guarantee this among the kubelet-managed containers/processes, hence my comment of being a 'bit dangerous').

pkg/cri/resource-manager/policy/builtin/memtier/node.go
askervin: Ah, sure! (Should have read the docs!) I'll take this back and see what I could do the `kube-system` allocation...

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
askervin: Yeah, now I recall we've discussed this earlier. It seems back then (https://github.com/intel/cri-resource-manager/commit/bf1642bdfac8a0de09fa0480a0202b1f723bea2e) you mentioned a pending commit that would change this behavior. But I don't remember anymore what unwanted behavior it would unleash...

klihub: That pending commit was dropped as it would have caused problems for containers in the BestEffort QoS class. Those come with a CPU request/limit of 0 milli-CPUs (which translates to 2 CFS cpus shares). These containers still need to be able run on some set of  CPUs. This set is the shared subset of the pool. If we allowed a pool to ever run out of shared CPUs and the pool had BestEffort QoS class containers assigned, then we'd end up trying to set cpuset.cpus for these containers to the empty set, which is then rejected by the kernel.

One easy kludge-around for this would be to account for BestEffort containers as having a request of  1 milli-CPU instead of 0. That would still leave the (theoretical) problem of Guaranteed QoS class containers first exhausting all the CPUs and then a scheduler BestEffort QoS class one getting assigned to the empty cpuset and failing.

IMO the real fix would be to really implement reserved CPUs and then co-locate BestEffort QoS class containers together with the ones in the kube-system namespace there.

askervin:
@klihub, I ended up rescoping this PR back to the roots - and beyond. That is, it will test only AvailableResources, and it makes a check on shared pool in the case where allocations look sane.

With `sed -i s/CONTCOUNT=2 CPU=980m/CONTCOUNT=1 CPU=2500m/' code.var.sh`, however, I'm getting (in interactive mode):
```
run.sh> pp 'cpus["pod1c0"]'                                                        |
{'cpu10'}
```
yet allocations in cache says:
```
run.sh> pp 'allocations["pod1c0"]'
{'ColdStart': 0,
 'Container': '076eac8c-b65f-40da-8f74-c68c7c53e228:pod1c0',
 'Exclusive': '8-9',
 'MemType': 'DRAM,PMEM,HBM',
 'MemoryLimit': {'0': 0, '1': 100000000, '2': 0, '4': 0, '7': 0},
 'MemoryPool': 'numa node #2',
 'Memset': '2',
 'Part': 500,
 'Pool': 'numa node #2'}
```

That is, it looks like real cpus_allowed do not include exclusive cpus. This happens on the (default) distro `ubuntu-20.04`, so it's probably not broken `runc` this time.

pkg/cri/resource-manager/policy/builtin/memtier/node.go
klihub: > @klihub, I ended up rescoping this PR back to the roots - and beyond. That is, it will test only AvailableResources, and it makes a check on shared pool in the case where allocations look sane.
> 
> With `sed -i s/CONTCOUNT=2 CPU=980m/CONTCOUNT=1 CPU=2500m/' code.var.sh`, however, I'm getting (in interactive mode):
> 
> ```
> run.sh> pp 'cpus["pod1c0"]'                                                        |
> {'cpu10'}
> ```
> 
> yet allocations in cache says:
> 
> ```
> run.sh> pp 'allocations["pod1c0"]'
> {'ColdStart': 0,
>  'Container': '076eac8c-b65f-40da-8f74-c68c7c53e228:pod1c0',
>  'Exclusive': '8-9',
>  'MemType': 'DRAM,PMEM,HBM',
>  'MemoryLimit': {'0': 0, '1': 100000000, '2': 0, '4': 0, '7': 0},
>  'MemoryPool': 'numa node #2',
>  'Memset': '2',
>  'Part': 500,
>  'Pool': 'numa node #2'}
> ```
> 
> That is, it looks like real cpus_allowed do not include exclusive cpus. This happens on the (default) distro `ubuntu-20.04`, so it's probably not broken `runc` this time.

Hmm, I need to reread the code and see if my earlier comment about fractional Guaranteed QoS allocations only in the 1 < cpu req. < 2 range is true. Because it looks to me as if here we'd try to do the same (IOW, allocate 2 CPUs exclusively, then assign the container to run on those + the shared pool with the remaining fractional (.5) CFS share ('Part')...

So what do you have as that container's cpuset.cpus entry ? We don't record the exact shared shared cpuset of a pool per container, only a reference to it via the 'Pool' here together with how big slice of the shared part of that to assign to the container. When assigning CPUs, what should happen is that if the shared part is not 0, then the cpuset should be union(exclusive, pool shared cpuset)...

klihub: I did a quick test with something like this:
```
...
      resources:
        requests:
          cpu: 2500m
          memory: 200M
        limits:
          cpu: 2500m
          memory: 200M
...
   resources:
        requests:
          cpu: 1500m
          memory: 200M
        limits:
          cpu: 1500m
          memory: 200M
```

That ends up with allocations in the cache like this:
```
  "ce782bf5-6162-4e6d-bc83-fcdf8cb3d745:c0": {
    "Exclusive": "0,12",
    "Part": 500,
    "Container": "ce782bf5-6162-4e6d-bc83-fcdf8cb3d745:c0",
    "Pool": "numa node #0",
    "MemoryPool": "numa node #0",
    "MemType": "DRAM,PMEM,HBM",
    "Memset": "0,4",
    "MemoryLimit": {
      "0": 0,
      "1": 0,
      "2": 200000000,
      "4": 0,
      "7": 0
    },
    "ColdStart": 0
  },
...
  "ce782bf5-6162-4e6d-bc83-fcdf8cb3d745:c1": {
    "Exclusive": "2",
    "Part": 500,
    "Container": "ce782bf5-6162-4e6d-bc83-fcdf8cb3d745:c1",
    "Pool": "numa node #2",
    "MemoryPool": "numa node #2",
    "MemType": "DRAM,PMEM,HBM",
    "Memset": "2,5",
    "MemoryLimit": {
      "0": 0,
      "1": 0,
      "2": 200000000,
      "4": 0,
      "7": 0
    },
    "ColdStart": 0
  },
```

With the NUMA node pools having this state  
```
D: [     policy     ] <post-alloc>        <NUMA node #0>
D: [     policy     ] <post-alloc>          - <numa node #0 capacity: CPU: isolated:60, sharable:0-48:12,72-108:12 (9000m), MemLimit: dram:33177960448, pmem:133143986176>
D: [     policy     ] <post-alloc>          - <numa node #0 allocatable: CPU: isolated:60, sharable:24-48:12,72-108:12 (granted:500m/500m local/subtree, allocatable:6500m), MemLimit: dram:33177960448, pmem:132943986176>
D: [     policy     ] <post-alloc>          - normal memory: 0
D: [     policy     ] <post-alloc>          - HBM memory: 
D: [     policy     ] <post-alloc>          - PMEM memory: 4
D: [     policy     ] <post-alloc>            + <CPU grant for fractional-busybox:c0 from numa node #0: exclusive: 0,12, shared: 24,36,48,72,84,96,108 (500m), MemLimit: pmem:200000000>
...
D: [     policy     ] <post-alloc>        <NUMA node #2>
D: [     policy     ] <post-alloc>          - <numa node #2 capacity: CPU: isolated:6, sharable:18-102:12 (8000m), MemLimit: dram:33816096768, pmem:133143986176>
D: [     policy     ] <post-alloc>          - <numa node #2 allocatable: CPU: isolated:6, sharable:18-102:12 (granted:500m/500m local/subtree, allocatable:7500m), MemLimit: dram:33816096768, pmem:133143986176>
D: [     policy     ] <post-alloc>          - normal memory: 2
D: [     policy     ] <post-alloc>          - HBM memory: 
D: [     policy     ] <post-alloc>          - PMEM memory: 5
D: [     policy     ] <post-alloc>            + <CPU grant for fractional-busybox:c1 from numa node #2: exclusive: 2, shared: 18,30,42,54,66,78,90,102 (500m), MemLimit: pmem:200000000>
D: [     policy     ] <post-alloc>          - parent: <socket #0>
D: [     policy     ] <post-alloc>          - children:

```

The containers end up having these cpusets:
```
r640-1:~ # kubectl exec -ti fractional-busybox -c c0 -- /bin/cat /proc/self/status | grep _allowed_list:
Cpus_allowed_list:      24,36,48,72,84,96,108
Mems_allowed_list:      0,4
r640-1:~ # kubectl exec -ti fractional-busybox -c c1 -- /bin/cat /proc/self/status | grep _allowed_list:
Cpus_allowed_list:      18,30,42,54,66,78,90,102
Mems_allowed_list:      2,5
```

So I think it works as I suspected. I think I'll change the defaults to opt-in from opt-out and file a PR...

klihub:
@askervin You're right, updating the shared subset for mixed allocations is broken. #527 should fix it.

klihub:
@askervin Is this one ready for review ?

askervin:
> @askervin Is this one ready for review ?

Yes it is. Somewhat related tests for honoring ReservedResources are in PR #543 .

askervin:
Added e2e test for PR #524 (cgroup cpuset dir in AvailableResources).

Draft: not to be merged before PR #524 and PR #543. (Needs update after the latter.)

test/e2e/policies.test-suite/memtier/n4c16/test04-available-resources/code.var.sh
```diff
@@ -0,0 +1,46 @@
...
+       "mems['pod1c0'] == {'node2'}" \
+       "mems['pod1c1'] == {'node2'}"
+kubectl delete pods --all --now
+reset counters
+
+# Test cgroup cpuset directory in AvailableResources.CPU
+terminate cri-resmgr
+CRIRM_CGROUP=/sys/fs/cgroup/cpuset/cri-resmgr-test-05
+vm-command "rm -rf $CRIRM_CGROUP; mkdir $CRIRM_CGROUP; echo 1-4 > $CRIRM_CGROUP/cpuset.cpus"
+AVAILABLE_CPU="\"$CRIRM_CGROUP\""
```
klihub, line 27: Could this test be also repeated with `AVAILABLE_CPU="\"${CRIRM_CGROUP#/sys/fs/cgroup/cpuset}"\"` ? It should produce identical results and that's the kubelet-compatible way of providing the path.

klihub: Otherwise it LGTM.

askervin:
^^^
1st force-push: rebased on top of master (most importantly, PR #543)
2nd force-push: updated tests to run with reserved CPUs (PR #543)
3rd force-push: add test for `AVAILABLE_CPU="\"${CRIRM_CGROUP#/sys/fs/cgroup/cpuset}\""`

(I modified the cpuset a little when using this cpuset syntax just to make sure new configuration is used and the test cannot be passed by luckily using old config due to any error...)

test/e2e/policies.test-suite/memtier/n4c16/test04-available-resources/code.var.sh
askervin: Thanks! Test added.

askervin:
Thanks for the review, @klihub.
Test added and proper cleanup (in the last force-push).
This is good to go in my perspective.

klihub (approved):
LGTM.

PR #517, opened 2020-11-19
memtier: honor allowed CPUs in resource discovery.

klihub:
If a cpuset is given among `AvailableResources`, use it as a bounding set during CPU resource discovery.


marquiz (approved):
lgtm

askervin (approved):
LGTM, partly fixes #516.

PR #518, opened 2020-11-19
memtier: honor CPU isolation opt-out preference.

klihub:
If a pod/container is annotated to opt out from isolated exclusive CPU allocation, or that is the configured memtier default and the pod/container is not annotated otherwise, honor this setting when allocating resources.


marquiz (approved):
lgtm

PR #519, opened 2020-11-20
memtier: get rid of hand-rolled policy/pools in most of the test cases.

klihub:
Get rid of the most of the cases where tests hand-rolled policy/pools instead of using the official policy 'constructor' for this.


marquiz (approved):
LGTM

PR #524, opened 2020-11-24
policy,memtier: better handling of bounding resource set for CPUs.

klihub:
This PR attempts to implement proper handling of the bounding/allowed CPU set
for the `memtier` policy. The patch set

- adds support for reading the bounding CPU set from a cgroup directory
- falls back to reallocating containers if reinstating allocations fails
- triggers rebuilding the topology pool if the reserved or allowed CPU set changes

Detecting a mismatch between the configuration of kubelet (`--cgroup-root`) and
CRI Resource Manager is not part of this patch set.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/memtier/cache.go
+++ b/pkg/cri/resource-manager/policy/builtin/memtier/cache.go
@@ -34,35 +34,63 @@ func (p *policy) saveAllocations() {
 	p.cache.Save()
 }
 
-func (p *policy) restoreAllocations() error {
-	// Get the allocations map.
-	if !p.cache.GetPolicyEntry(keyAllocations, &p.allocations) {
-		return nil
-	}
+func (p *policy) restoreAllocations(allocations *allocations) error {
+	savedAllocations := allocations.clone()
+	p.allocations = p.newAllocations()
 
 	//
-	// Based on the allocations
-	//   1) update the free supply of the grant's pool to account for the grant
-	//   2) set the extra memory allocations to the nodes below the grant in the tree
+	// Try to reinstate all grants with the exact same resource assignments
+	// as saved. If that fails, release and try to reallocate all corresponding
+	// containers with pool hints pointing to the currently assigned pools. If
+	// this fails too, save the original allocations unchanged to the cache and
+	// return an error.
 	//
-	// We assume (for now) that the allocations are correct and the workloads don't
-	// need moving.
-	//
-	for id, grant := range p.allocations.grants {
-		pool := grant.GetCPUNode()
 
-		log.Info("updating pool %s for container %s CPU grant", pool.Name(), id)
-		supply := pool.FreeSupply()
-		if err := supply.Reserve(grant); err != nil {
+	if err := p.reinstateGrants(allocations.grants); err != nil {
+		log.Error("failed to reinstate grants verbatim: %v", err)
+		containers, poolHints := allocations.getContainerPoolHints()
+		if err := p.reallocateResources(containers, poolHints); err != nil {
+			p.allocations = savedAllocations
+			p.saveAllocations() // undo any potential changes in saved cache
 			return err
 		}
+	}
 
-		log.Info("updating pool %s for container %s extra memory", pool.Name(), id)
+	return nil
+}
+
+// reinstateGrants tries to restore the given grants exactly as such.
+func (p *policy) reinstateGrants(grants map[string]Grant) error {
+	for id, grant := range grants {
+		c := grant.GetContainer()
+
+		pool := grant.GetCPUNode()
+		supply := pool.FreeSupply()
+
+		if err := supply.Reserve(grant); err != nil {
+			return policyError("failed to update pool %q with CPU grant of %q: %v",
+				pool.Name(), c.PrettyName(), err)
+		}
+
+		log.Info("updated pool %q with reinstated CPU grant of %q",
+			pool.Name(), c.PrettyName())
+
 		pool = grant.GetMemoryNode()
 		if err := supply.ReserveMemory(grant); err != nil {
-			return err
+			grant.GetCPUNode().FreeSupply().ReleaseCPU(grant)
+			return policyError("failed to update pool %q with extra memory of %q: %v",
+				pool.Name(), c.PrettyName(), err)
 		}
+
+		log.Info("updated pool %q with reinstanted memory reservation of %q",
+			pool.Name(), c.PrettyName())
+
+		p.allocations.grants[id] = grant
+		p.applyGrant(grant)
 	}
+
+	p.updateSharedAllocations(nil)
+
 	return nil
 }
 
--- a/pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
@@ -150,18 +150,8 @@ func (p *policy) AllocateResources(container cache.Container) error {
 		return policyError("failed to allocate resources for %s: %v",
 			container.PrettyName(), err)
 	}
-
-	if err := p.applyGrant(grant); err != nil {
-		if _, _, err = p.releasePool(container); err != nil {
-			log.Warn("failed to undo/release unapplicable grant %s: %v", grant, err)
-			return policyError("failed to undo/release unapplicable grant %s: %v", grant, err)
-		}
-	}
-
-	if err := p.updateSharedAllocations(grant); err != nil {
-		log.Warn("failed to update shared allocations affected by %s: %v",
-			container.PrettyName(), err)
-	}
+	p.applyGrant(grant)
+	p.updateSharedAllocations(grant)
 
 	p.root.Dump("<post-alloc>")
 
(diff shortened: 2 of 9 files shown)


kad:
Just additional idea to this: if we detect kubelet hierarchy during some container create/list operations, we can re-check how is the top kubelet group is configured. So, even if admin didn't specified flags for us, we can detect settings from kubelet.

askervin:
(Un)luckily tested with:
```
mkdir /sys/fs/cgroups/cpuset/cri-resmgr
echo 2-8,12-13 > /sys/fs/cgroups/cpuset/cri-resmgr/cpuset.cpus
```

and then with `cri-resmgr.cfg`:
```
policy:
  Active: memtier
  AvailableResources:
    CPU: "/sys/fs/cgroup/cpuset/cri-resmgr"
  ReservedResources:
    CPU: cpuset:8
logger:
  Debug: cri-resmgr,resource-manager,cache,policy
```

`-` in the cgroup path seems to cause an issue:
```
FATAL ERROR: [   cri-resmgr   ] failed to create resource manager instance: resource-manager: failed to load forced configuration cri-resmgr-cpuset.cfg: config error:
 module policy: failed to apply configuration: error unmarshaling JSON: while decoding JSON: policy: failed to parse CPU cpuset constraint "/sys/fs/cgroup/cpuset/cri-resmgr": strconv.Atoi: parsing "/sys/fs/cgroup/cpuset/cri": invalid syntax
```

klihub:
> FATAL ERROR: [   cri-resmgr   ] failed to create resource manager instance: resource-manager: failed to load forced configuration cri-resmgr-cpuset.cfg: config error:
>  module policy: failed to apply configuration: error unmarshaling JSON: while decoding JSON: policy: failed to parse CPU cpuset constraint "/sys/fs/cgroup/cpuset/cri-resmgr": strconv.Atoi: parsing "/sys/fs/cgroup/cpuset/cri": invalid syntax
> ```
> 
> Apart from this, new feature seemed to work fine already. When flowing in, I could update tests in #516 to cover this.

Thanks for giving it a go, I'll check that bug.

That said, the handling/exact details of how that particular bit works is subject to change. This is due to how `kubelet/the systemd cgroup driver` handles the cgroup root (`--cgroup-root foo`) option. It is used to prefix, with a dash-separator, the `kubepods.slice` default and then the resulted path is mangled heavily according to the systemd algorithm. One of the end results is that it is impossible to configure kubelet according to the above configuration. In practice, if you'd try something liek that (IOW pass the option `--cgroup-root /cri-resmgr`), it would literally turn in kubelet to something like `/sys/fs/cgroup/cpuset/cri_resmgr-kubepods.slice`).

klihub:
> Thanks for giving it a go, I'll check that bug.

That should be already fixed in the current incarnation. It was caused by the previous ordering of switch cases always trying to parse the value as a cpuset if it contained a dash.

klihub:
> That said, the handling/exact details of how that particular bit works is subject to change. This is due to how `kubelet/the systemd cgroup driver` handles the cgroup root (`--cgroup-root foo`) option. It is used to prefix, with a dash-separator, the `kubepods.slice` default and then the resulted path is mangled heavily according to the systemd algorithm. One of the end results is that it is impossible to configure kubelet according to the above configuration. In practice, if you'd try something liek that (IOW pass the option `--cgroup-root /cri-resmgr`), it would literally turn in kubelet to something like `/sys/fs/cgroup/cpuset/cri_resmgr-kubepods.slice`).

Sorry, that statement of mine is untrue. With the above cgroup root argument the top level kubelet cpuset directory simply becomes `/sys/fs/cgroup/cpuset/cri_resmgr.slice`. Only the next level becomes as stated above.

pkg/cri/resource-manager/policy/builtin/memtier/cache.go
```diff
@@ -34,35 +34,63 @@ func (p *policy) saveAllocations() {
...
+
 		pool = grant.GetMemoryNode()
 		if err := supply.ReserveMemory(grant); err != nil {
-			return err
+			grant.GetCPUNode().FreeSupply().ReleaseCPU(grant)
+			return policyError("failed to update pool %q with extra memory of %q: %v",
+				pool.Name(), c.PrettyName(), err)
 		}
+
+		log.Info("updated pool %q with reinstanted memory reservation of %q",
```
askervin, line 85: s/reinstanted/reinstated

pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
```diff
@@ -345,23 +321,145 @@ func (p *policy) ExportResourceData(c cache.Container) map[string]string {
...
+
+	cache.SortContainers(containers)
+
+	for _, c := range containers {
+		p.releasePool(c)
+	}
+	for _, c := range containers {
+		log.Debug("reallocating resources for %s...", c.PrettyName())
+
+		grant, err := p.allocatePool(c, pools[c.GetCacheID()])
```
askervin, line 338: I have a doubt on giving pool hints after changing _allowed_/_reserved_ sets, as the topology tree can be different. Is here any danger that pool hints are cached legacy information that may result in suboptimal workload placement after reconstructing the topology tree?

(Of course it would make sense to minimize moving workloads from a pool to another, especially in short term. But on the other hand, it would sound right to target optimal long term placement in _reallocateResources_.)

askervin (commented):
Looks pretty good. Two questions and a couple of typos...

Haven't tested yet...

pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
```diff
@@ -345,23 +321,145 @@ func (p *policy) ExportResourceData(c cache.Container) map[string]string {
...
+			if reserveCnt != p.reserveCnt {
+				log.Warn("CPU reservation has changed (%v, was %v)",
+					reserveCnt, p.reserveCnt)
+				reinit = true
+			}
+		}
+	}
+
+	if !allowed.Equals(p.allowed) {
+		if !(allowed.Size() == 0 && p.allowed.Size() == 0) {
```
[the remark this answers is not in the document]
klihub: Hmm... that's not a size change check. I think it tries to filter out the corner case when the old and new sets are not equal (IOW not reflect.DeepEqual) but they are both of size 0. I have no recollection why though. I guess that could only happen if one of the sets is uninitialized (so the elems map is nil) and the other is initialized to the empty set (so the elems map is empty). And I don't see how that could happen. I need to check this more carefully.

pkg/cri/resource-manager/policy/policy.go
```diff
@@ -431,3 +433,11 @@ func ConstraintToString(value Constraint) string {
 		return fmt.Sprintf("<???(type:%T)>", value)
 	}
 }
+
+// configNotify is the configuration change notification callback for the genric policy layer.
```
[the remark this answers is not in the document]
klihub: Fixed.

pkg/cri/resource-manager/policy/builtin/memtier/cache.go
klihub: Fixed.

pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
klihub: We are in damage control mode here. We try to avoid/minimize the costly disruption of moving Guaranteed containers across memory controllers/NUMA nodes, yet we (decided that we) need to reallocate them all. Once we release all containers, we sort them primarily by declared importance (QoS class) and secondarily by size (first memory then CPU). So effectively we end up reallocating Guaranteed containers from largest (as in terms of memory request) to smallest first before the rest.

Without using the old pools as NUMA node hints, even if all the Guaranteed containers would nicely fit into their currently allocated pools, there would be zero guarantee that any of them would end up there. The probability for that to happen for any single container is quasi-random, because where the container ended up originally depends on the order of all container allocation and release requests from the very beginning, and the sizes of the containers involved, both in terms of memory and CPU requirements.

So can we end up in sub-optimal overall workload placement after reallocation ? Yes, definitely we can. But that is not our primary concern here. That can easily happen just by a normal sequence of allocations and releases without any reconfiguration in between. We primarily try to avoid disruption of Guaranteed containers, if/when that is (easily) avoidable. 

With rebalancing, we might want to support opting for overall optimal workload placement instead of avoiding disruption. However, IMO there needs to be an explicit indication for that (e.g. as a preference among the configuration).

pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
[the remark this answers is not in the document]
askervin: Uh-oh, yeah, right. (Feeling blind and seeing only what I'm imagining.) So any difference in the allowed sets leads to reinitialization. That's very good.

I'm not marking this conversation resolved if you'd like to have a thorough check for the case where both sizes are 0 but cpusets inequal.

pkg/cri/resource-manager/policy/builtin/memtier/memtier-policy.go
askervin: Ok, not disrupting sounds good default policy.

askervin: I think the logic here is fine.

askervin (approved):
LGTM

PR #530, opened 2020-11-27
memtier/c4pmem4/test03-coldstart: don't jump the gun.

klihub:
Don't let previous cold start triggers in the log
fool us into thinking that the pending cold start
period would be over.


askervin (approved):
LGTM

PR #537, opened 2020-11-30
docs: minor fixes in html template customization

marquiz:
- remove 'v:' prefix from the version menu in order to avoid displaying
  something like "v: v0.4.1"
- correct usage of `<dt>` tag


klihub (approved):
LGTM.

PR #545, opened 2020-12-02
resource-manager: run post-update hooks after reconfiguration.

klihub:
Run post-update hooks after successful change of configuration to enforce any policy
decisions made as a side-effect of the active adjusting itself to the new configuration.

### DIFF
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -16,10 +16,13 @@ package resmgr
 
 import (
 	"context"
+	"fmt"
 
 	criapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 
+	pkgcfg "github.com/intel/cri-resource-manager/pkg/config"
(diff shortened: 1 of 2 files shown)


marquiz (approved):
Makes sense and looks good to me.

I was just wondering what was the reasoning behdind moving the code over to `requests.go`?

klihub:
> Makes sense and looks good to me.
> 
> I was just wondering what was the reasoning behdind moving the code over to `requests.go`?

I moved it because we've tried, with regular irregularity, to keep together in `requests.go` most of the code that either needs to lock the `resource-manager` against the request processing pipeline or needs to invoke controller hooks. It's irregular ATM, because by that logic `setAdjustment()` should be there as well.

Hmm... and now that you asked, I noticed that since the locking have been shuffled from the exported to the unexported function, we could probably do away with most of the `setConfig[FromFile]()/setAdjustment()` boilerplates which don't do anything else than log.

PR #546, opened 2020-12-04
go.mod: update to Kubernetes v1.19.4

marquiz:


klihub:
Here is the fix for that unit test failure... https://github.com/marquiz/cri-resource-manager/pull/3.

klihub:
This LGTM, if we still pass the full end-to-end test set with this.

klihub (approved):
LGTM. All end-to-end test pass in my test run.

PR #547, opened 2020-12-04
scripts: helper for maintaining replace lines in go.mod

marquiz:
Help by printing out updated replace directives that can then be copied
to go.mod. An example for updating to k8s.io modules to v0.19.4:

  $ sed -n '/replace/,$p' go.mod  | grep k8s.io | awk '{print $1}' | \
    xargs ./scripts/hack/go-mod-replace-helper.sh ../k8s-cache/ v0.19.4

Note that the script does not do any dependency checking/tidying, but,
just blindly prints out revisions of whatever it is given.

### DIFF
--- a/scripts/hack/go-mod-replace-helper.sh
+++ b/scripts/hack/go-mod-replace-helper.sh
@@ -1,0 +1,101 @@
+#!/bin/bash -e
+set -o pipefail
+
+this=`basename $0`
+
+usage () {
+cat << EOF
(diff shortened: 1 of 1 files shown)


marquiz:
A bit tidied version of what I used behind closed doors in my dark cave. I know this is ugly but @klihub asked for this...

klihub (approved):
Maybe it's a bit ugly, but better than nothing or doing all of this by hand.

PR #550, opened 2020-12-04
test/functional: prevent read/write data race in klog.

klihub:
Prevent grpc from triggering read/write data race (concurrent klog flag.Set()
by our klogcontrol and klogging connection complaints by default grpc log)
by overriding the grpc logger.

Should fix the latest test failures in #546


marquiz (approved):
Thanks @klihub for looking into this!
LGTM

Issue #552, opened 2020-12-07
Rudimentary support for SST-CP


marquiz:
**Describe the solution you'd like**

Rudimentary support for Speed Select Technology Core Power (SST-CP). More specifically, detect high priority CPUs by examinging the sysfs cpufreq parameters.

**Why this is needed**

Initial support for SST-CP (Ice Lake Xeon platform). Align with CMK.

msivosuo:
Implemented by https://github.com/intel/cri-resource-manager/pull/557, closing.

PR #557, opened 2020-12-08
Improve CPU prioritization

marquiz:
Improve detection of high priority CPUs. Examine
cpufreq/enenrgy_performance_profile setting of each CPU at system
discovery. All other than the least performance-oriented CPUs are
considered as prioritized. Some examples:
- if all cpus have the same epp value none will be flagged as prioritied
- cpus 0,1 are 'balance_performance', 2,3 are 'balance_power' and 4,5
  are 'power' -> cpus 0-3 are flagger as priority cpus

Implements #552

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -152,6 +152,7 @@ type CPU interface {
 	ThreadCPUSet() cpuset.CPUSet
 	BaseFrequency() uint64
 	FrequencyRange() CPUFreq
+	EPP() EPP
 	Online() bool
 	Isolated() bool
 	SetFrequencyLimits(min, max uint64) error
@@ -167,6 +168,7 @@ type cpu struct {
 	threads  IDSet   // sibling/hyper-threads
 	baseFreq uint64  // CPU base frequency
 	freq     CPUFreq // CPU frequencies
+	epp      EPP     // Energy Performance Preference from cpufreq governor
 	online   bool    // whether this CPU is online
 	isolated bool    // whether this CPU is isolated
 }
@@ -177,6 +179,17 @@ type CPUFreq struct {
 	max uint64   // maximum frequency (kHz)
 	all []uint64 // discrete set of frequencies if applicable/known
 }
+
+// EPP represents the value of a CPU energy performance profile
+type EPP int
+
+const (
+	EPPPerformance EPP = iota
+	EPPBalancePerformance
+	EPPBalancePower
+	EPPPower
+	EPPUnknown
+)
 
 // MemInfo contains data read from a NUMA node meminfo file.
 type MemInfo struct {
@@ -301,6 +314,7 @@ func (sys *system) Discover(flags DiscoveryFlag) error {
 			sys.Debug("    threads: %s", cpu.threads)
 			sys.Debug("  base freq: %d", cpu.baseFreq)
 			sys.Debug("       freq: %d - %d", cpu.freq.min, cpu.freq.max)
+			sys.Debug("        epp: %d", cpu.epp)
 		}
 
 		sys.Debug("offline CPUs: %s", sys.offline)
@@ -544,6 +558,9 @@ func (sys *system) discoverCPU(path string) error {
 	if _, err := readSysfsEntry(path, "cpufreq/cpuinfo_max_freq", &cpu.freq.max); err != nil {
 		cpu.freq.max = 0
 	}
+	if _, err := readSysfsEntry(path, "cpufreq/energy_performance_preference", &cpu.epp); err != nil {
+		cpu.epp = EPPUnknown
+	}
 	if node, _ := filepath.Glob(filepath.Join(path, "node[0-9]*")); len(node) == 1 {
 		cpu.node = getEnumeratedID(node[0])
 	} else {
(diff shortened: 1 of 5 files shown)


pkg/sysfs/system.go
```diff
@@ -995,3 +1017,40 @@ func (sys *system) discoverCache(path string) error {
...
+
+// EPPFromString converts string to EEP value
+func EPPFromString(s string) EPP {
+	for e, v := range eppStrings {
+		if s == v {
+			return EPP(e)
+		}
+	}
+	return EPPUnknown
+}
```
klihub, line 1056: nitpick: considering all the trouble we go through just to be able to use a fixed-sized array instead of a slice for Epp.String(), wouldn't it then be justified and more symmetric to use a map instead of iteration for the other direction, with something like this ? 
```
var namedEPPs = func() map[string]EPP {
    m := map[string]EPP{}
    for epp, str := range eppStrings {
        m[str] = epp
    }
    return m
}()

func EPPFromString(name string) EPP {
    if epp, ok := namedEPPs[name]; ok {
        return epp
    }
    return EPPUnknown
}
```

marquiz: Mm, not a nitpick. The code was quite hastily written and I wasn't totally happy with it myself. Your suggestion would be an improvement. Maybe this is a good place to discuss what would be the sanest/safest way to specify enum-string types like this(?) I don't have a good answer, maybe you do 😄 

I tried a bit different approach utilizing the stringer tool in the new version of the patch. That should better ensure that the enums and strings are kept in sync with each other. Not that I think there would really be any changes here, but, at least as an excercise...

pkg/sysfs/system.go
```diff
@@ -152,6 +152,7 @@ type CPU interface {
 	ThreadCPUSet() cpuset.CPUSet
 	BaseFrequency() uint64
 	FrequencyRange() CPUFreq
+	EPP() EPP
```
klihub, line 155: Just a note, not something immediate: it might be necessary/useful later to add a `System.EPPCPUSet(EPP) cpuset.CPUSet` or something to that effect, for better configurability for the CPU allocator by the caller.

pkg/sysfs/system.go
```diff
@@ -995,3 +1018,20 @@ func (sys *system) discoverCache(path string) error {
...
+	return n
+}()
+
+// EPPFromString converts string to EEP value
+func EPPFromString(s string) EPP {
+	if e, ok := eppNames[s]; ok {
+		return e
+	}
+	return EPPUnknown
+}
```
klihub, line 1037: I have to say that I'm genuinely surprised that with that much boilerplate and a full blown generator, stringer is unable to generate this direction at all.

pkg/sysfs/system.go
klihub: > I don't have a good answer, maybe you do 

Nope, unfortunately I don't. But having seen this updated patch set, I have to say that stringer really seems half-arsed to me: you still need to manually write the string-to-type direction/function, and that ends up being half of your original code without stringer. Plus a bit of extra boilerplate. And you have to remember to manually run `make generate` when you change something. Almost as if the cure would become the poison.
 
> I tried a bit different approach utilizing the stringer tool in the new version of the patch. That should better ensure that the enums and strings are kept in sync with each other. Not that I think there would really be any changes here, but, at least as an excercise...

I'm not opposing this, but I sure don't like it either. If others are fine with this, I'm fine with it, too.

pkg/sysfs/system.go
marquiz: Mm, certainly, that is a good idea as EPP has a limited pre-defined set of values. I'll keep that in mind as this needs to be revisited when implementing more extensive SST support

pkg/sysfs/system.go
marquiz: Yeah, I think that is a very unfortunate major omission from the tool. Maybe this is an overkill for this 5-value enum. Otoh, I tried to explore something nice and reliable that could possibly be used across cri-rm.

I think I'll let this proof for a while, at least sleep over it...

klihub: > Yeah, I think that is a very unfortunate major omission from the tool. Maybe this is an overkill for this 5-value enum. Otoh, I tried to explore something nice and reliable that could possibly be used across cri-rm.

I do agree that's a worthy goal. It's just that ideally

  - the generator should be able to generate code in both the enum-to-string and string-to-enum directions
  - we'd have dependency rules in place to only invoke the generator when necessary

Working dependency rules might be just too much of a hassle/require too much manual Makefile boilerplate to do with the available go tooling. In that case a perfectly acceptable alternative would be to not commit the generated enum-related go source to the repo but instead always generate it. Then the Makefile-related misery would be reduced to simply having the necessary rules in place for every go source to generate the corresponding stringer-go source files unconditionally.

> I think I'll let this proof for a while, at least sleep over it...

Fair enough.

marquiz: Dropped stringer and returned to my original implementation, with the improvement you suggested. I don't consider this the most beautiful but don't know how to make it neater, either 😬

klihub (approved):
LGTM

pkg/sysfs/system.go
klihub: All things considered, I think this is probably the best compromise of all the things we tried.

PR #561, opened 2020-12-09
memtier: rework building of pool tree by HW topology.

klihub:
memtier: rework construction of the pool tree.
    
When building the tree of pools based on HW
topology, eliminate any branchless part of the
tree. In praticular, discard any pool node for
a die or NUMA node if it otherwise ended up
being the only child of its parent. Assign the
resources of omitted DRAM NUMA nodes to their
would-be parent leaf node. Omit all PMEM NUMA
nodes from the tree and assign their resources
to one of the closest DRAM NUMA nodes instead.
For single socket systems, use the lone socket
as root instead of creating a virtual root.

Also rework the pool/node resource discovery
code, simplifying it and fixing bugs to work
with the new pool tree setup code.


pkg/cri/resource-manager/policy/builtin/memtier/node.go
```diff
@@ -755,6 +836,45 @@ func (n *socketnode) DiscoverMemset() {
 	}
 }
 
+// AssignNUMANodes assigns the given NUMA nodes to this one.
+func (n *socketnode) AssignNUMANodes(ids []system.ID) {
```
kad, line 840: looking at 3 instances of almost exact functions gives a feeling that it can be generalized into separate function?

klihub: Updated to use a common function.

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -27,108 +27,237 @@ import (
...
+			found := false
+			count := 0
+			for _, nodeID := range socket.DieNodeIDs(dieID) {
+				node := p.sys.Node(nodeID)
+				if !node.CPUSet().IsEmpty() {
+					count++
+				}
+				if nodeID == numaNode.ID() {
+					found = true
+				}
```
marquiz, line 225: Stupid question to understand the code better: aren't we able to get the die of the node directly?

marquiz (commented):
From what I understand about this, looks basically good. Just a few minor comments.

Also, in the commit message:
`s/it that/if that/` and `s/node.  Like/node. Like/`

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
klihub: No, not with the current `pkg/sysfs` code. That only tracks 'first-class' topology objects, those with a system-wide unique ID, explicitly. Die IDs are only unique within their socket. But that shouldn't prevent us from storing the die ID in the node anyway. I'll change that. It makes more sense, since we store the package ID there, too..

pkg/cri/resource-manager/policy/builtin/memtier/pools.go
```diff
@@ -27,108 +27,237 @@ import (
...
+
+		numaSysNode := p.sys.Node(numaNodeID)
+		switch numaSysNode.GetMemoryType() {
+		case system.MemoryTypePMEM:
+			log.Debug("NUMA node #%d: PMEM node (omitting pool node)", numaNodeID)
+			pmemNodes[numaNodeID] = numaSysNode
+			continue // don't create pool, will assign to a closest DRAM node
+		case system.MemoryTypeDRAM:
+			log.Debug("NUMA node #%d: DRAM node", numaNodeID)
+			dramNodes[numaNodeID] = numaSysNode
```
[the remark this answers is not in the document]
klihub: Added default case to actively ignore and warn about ignoring nodes with unhandled memory types.

askervin:
Tested with topologies:
- `[{"mem": "2G", "cores": 2, "nodes": 1, "dies": 2, "packages": 2}]` (2 packages == sockets, 2 dies/package, 1 NUMA node/die, 2 cores/NUMA node, 2 threads/core)
- `[{"mem": "2G", "cores": 2, "nodes": 2, "packages": 2}]`
- `[{"mem": "2G", "threads": 1, "cores": 1, "packages": 4}, {"mem": "4G", "node-dist": {"2": 17}}, {"mem": "4G", "node-dist": {"3": 17}}, {"mem": "4G", "node-dist": {"0": 17}}, {"mem": "4G", "node-dist": {"1": 17}}]` (that is 4 packages + 4G PMEM per package)

Looks good!

askervin (approved):
LGTM.

Tested with multisocket-PMEM and single-socket-multidie-PMEM setups. Works great.

askervin:
@klihub, a question regarding the march order with updating e2e PMEM node assignment test.

Currently the test relies on cri-resmgr output - yet the introspection interface would be a good option, too.

What do you think, should we first change the test so that it works for both old and new versions - or change the test in the same commit with these changes? (The worst option in my opinion would be fixing the test afterwards, as it would be nice to keep tests passing for every commit in the master.)

The change below makes the test generic enough to pass on both old and new versions. I'll make a PR if you'd like to take it in separately.

```
--- a/test/e2e/policies.test-suite/memtier/c4pmem4/test01-pmem-node-assigning/code.var.sh
+++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test01-pmem-node-assigning/code.var.sh
@@ -9,10 +9,10 @@ CRI_RESMGR_OUTPUT_COMMAND="cat cri-resmgr.output.txt"

 echo "Verify PMEM node assignment to CPU-ful nodes"
 for expected_output in \
-    "PMEM node #4 assigned to #2" \
-    "PMEM node #5 assigned to #3" \
-    "PMEM node #6 assigned to #0" \
-    "PMEM node #7 assigned to #1"; do
+    "PMEM node #4 assigned to .*#2" \
+    "PMEM node #5 assigned to .*#3" \
+    "PMEM node #6 assigned to .*#0" \
+    "PMEM node #7 assigned to .*#1"; do
     vm-command "$CRI_RESMGR_OUTPUT_COMMAND | grep -E '$expected_output'" ||
         command-error "expected PMEM assignment not found"
```

(cri-resmgr prints now rows like `- PMEM node #5 assigned to #socket #3 with distance 17`. In this test setup it has one extra `#`, but it may look better in a case where PMEM is assigned to a NUMA node level surrogate node instead of a socket.)

klihub:
> @klihub, a question regarding the march order with updating e2e PMEM node assignment test.
> 
> Currently the test relies on cri-resmgr output - yet the introspection interface would be a good option, too.

I'm always vary of tests relying on/parsing debug/log messages. Usually this causes messages which were not meant to be in any way become almost part of the public interface. Related bad memories from Push-To-Talk-Over-Cellular times still haunt me to date...

If we can avoid that by extending the introspection interface to expose the necessary data, then I'm all for it. So I guess we would need to make the pool hierarchy and the per-pool resources (including the NUMA node set) available for introspection. I think it would be the easiest for now to include a policy-specific section there and expose the memtier pool data in that. If we ever generalize the pool representation across policies, then we can turn it at the same time into a generic piece among the data.

> 
> What do you think, should we first change the test so that it works for both old and new versions - or change the test in the same commit with these changes? (The worst option in my opinion would be fixing the test afterwards, as it would be nice to keep tests passing for every commit in the master.)

Either of those work. Probably for an outside observer, changing the tests to work in the same commit or at least in the same PR would make more sense.

> 
> The change below makes the test generic enough to pass on both old and new versions. I'll make a PR if you'd like to take it in separately.
> 
> ```
> --- a/test/e2e/policies.test-suite/memtier/c4pmem4/test01-pmem-node-assigning/code.var.sh
> +++ b/test/e2e/policies.test-suite/memtier/c4pmem4/test01-pmem-node-assigning/code.var.sh
> @@ -9,10 +9,10 @@ CRI_RESMGR_OUTPUT_COMMAND="cat cri-resmgr.output.txt"
> 
>  echo "Verify PMEM node assignment to CPU-ful nodes"
>  for expected_output in \
> -    "PMEM node #4 assigned to #2" \
> -    "PMEM node #5 assigned to #3" \
> -    "PMEM node #6 assigned to #0" \
> -    "PMEM node #7 assigned to #1"; do
> +    "PMEM node #4 assigned to .*#2" \
> +    "PMEM node #5 assigned to .*#3" \
> +    "PMEM node #6 assigned to .*#0" \
> +    "PMEM node #7 assigned to .*#1"; do
>      vm-command "$CRI_RESMGR_OUTPUT_COMMAND | grep -E '$expected_output'" ||
>          command-error "expected PMEM assignment not found"
> ```

I can include this patch with you as the author in the same PR, if that's fine with you.

> 
> (cri-resmgr prints now rows like `- PMEM node #5 assigned to #socket #3 with distance 17`. In this test setup it has one extra `#`, but it may look better in a case where PMEM is assigned to a NUMA node level surrogate node instead of a socket.)

Thanks for spotting that. I'll remove the extra '#'. I did not think it properly through.

askervin:
> If we can avoid that by extending the introspection interface to expose the necessary data, then I'm all for it. So I guess we would need to make the pool hierarchy and the per-pool resources (including the NUMA node set) available for introspection. I think it would be the easiest for now to include a policy-specific section there and expose the memtier pool data in that. If we ever generalize the pool representation across policies, then we can turn it at the same time into a generic piece among the data.

Yeah, I think that would be the way to go. Hmm.. And/or would it make sense to include the topology tree into cache? I'm now thinking of toggling SNC HW configuration, for instance. Haven't thought this before, but I suppose it would be good if we'd notice if current topology is different from what was used when grants were cached.

..
> I can include this patch with you as the author in the same PR, if that's fine with you.

Sure.

klihub:
> Yeah, I think that would be the way to go. Hmm.. And/or would it make sense to include the topology tree into cache?

It can now be included in the policy-specific part of the cache (because ATM the internal representation is not common across policies). For instance, `memtier` stores references to pools as part of the grants it saves among its private policy-specific data.

> I'm now thinking of toggling SNC HW configuration, for instance. Haven't thought this before, but I suppose it would be good if we'd notice if current topology is different from what was used when grants were cached.

We do notice it, but only if it is relevant for the active policy. The policy itself realizes this, not any code in the agnostic layer.  With #524 in place, in `memtier` we notice this when we can't reinstate an existing grant because the associated pool does not exist any more. The attempted resolution then is to internally free and reallocate all grants, with preferred pool hints set to the existing/old pools.

askervin (approved):
LGTM!

PR #563, opened 2020-12-09
memtier/pool test: fix fd leakage causing test panics with more data.

klihub:
Don't leak a file descriptor for every file entry `uncompress()`ed from a testdata tarball.
Don't leak a file descriptor for every testdata tarball `uncompress()`ed. Also fix symlink
creation in `uncompress()` to work properly, instead of creating an empty file for each
symlink entry in the tarball.

Without this PR, adding more sample machine data causes a panic for any test that
tries to `uncompress()` the testdata tarball.


askervin (approved):
LGTM

Issue #564, opened 2020-12-14
Support for Kata Containers in CRI-RM


Dodan:
**Describe the solution you'd like**
At the moment, runc containers are the only types of container runtimes that CRI-RM can be aware of.

Kata containers are another type of runtime, that can be easily introduced to the project particularly for 2 types of controllers, RDT and cgroup.

The solution would require Kata containers to work in the SandboxCgroupOnly mode, and would extend the search for names of the containers subcgroups  to include the prefix "kata_".

**Why this is needed**

Adding Kata support, at least for these 2 controllers would be beneficial for both Kata users and CRI-RM users.

PR #566, opened 2020-12-15
Kata container support

Dodan:
This branch adds support for Kata containers in CRI-RM, particularly detection and classification support for the cgroup and RDT controllers.

The solution proposed should solve issue #564

### DIFF
--- a/pkg/utils/cgroups.go
+++ b/pkg/utils/cgroups.go
@@ -80,7 +80,7 @@ func isProcess(processID string) bool {
 	return true
 }
 
-func getTasksInContainer(cgroupParentDir, containerID string, onlyProcesses bool) ([]string, error) {
+func getTasksInContainer(cgroupParentDir, podID, containerID string, onlyProcesses bool) ([]string, error) {
 	var entries []string
 
 	// Find Cpuset sub-cgroup directory of this container
@@ -92,6 +92,7 @@ func getTasksInContainer(cgroupParentDir, containerID string, onlyProcesses bool
 			filepath.Join(CpusetCgroupDir, cgroupParentDir, "crio-"+containerID+".scope"),
 			filepath.Join(CpusetCgroupDir, cgroupParentDir, "docker-"+containerID+".scope"),
 			filepath.Join(CpusetCgroupDir, cgroupParentDir, containerID),
+			filepath.Join(CpusetCgroupDir, cgroupParentDir, "kata_"+podID),
 		}
 		for _, d := range dirs {
 			info, err := os.Stat(d)
@@ -130,11 +131,12 @@ func getTasksInContainer(cgroupParentDir, containerID string, onlyProcesses bool
 }
 
 // GetProcessesInContainer gets the IDs of all processes in the container.
-func GetProcessesInContainer(cgroupParentDir, containerID string) ([]string, error) {
-	return getTasksInContainer(cgroupParentDir, containerID, true)
+func GetProcessesInContainer(cgroupParentDir, podID, containerID string) ([]string, error) {
+	return getTasksInContainer(cgroupParentDir, podID, containerID, true)
 }
 
 // GetTasksInContainer gets the IDs of all tasks in the container.
-func GetTasksInContainer(cgroupParentDir, containerID string) ([]string, error) {
-	return getTasksInContainer(cgroupParentDir, containerID, false)
+func GetTasksInContainer(cgroupParentDir, podID, containerID string) ([]string, error) {
+	return getTasksInContainer(cgroupParentDir, podID, containerID, false)
+
 }
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -84,3 +84,4 @@ policying container resource allocations. Next, you could see:
 - [Setup](setup.md) for details on setup and usage
 - [Node Agent](node-agent.md) for seting up cri-resmgr-agent for dynamic configuration and more
 - [Webhook](webhook.md) for setting up our resource-annotating webhook
+- [Kata support](setup.md#kata-containers) for setting up CRI-RM with Kata containers
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -286,6 +286,46 @@ provided [sample configuration](/sample-configs/cri-full-message-dump.cfg)
 for doing this.
 
 
+## Kata Containers
+
+[Kata Containers](https://katacontainers.io/) is an open source container runtime, 
+building lightweight virtual machines that seamlessly plug into the containers ecosystem.
+
+In order to enable Kata container in a Kubernetes-CRI-RM stack, both Kubernetes 
+and the Container Runtime need to be aware of the new runtime environment:
+
+  * The Container Runtime can only be CRI-O or containerd, and will need to
+   have the runtimes enabled in their configuration files.
+  * Kubernetes will also have to be made aware of the CRI-O/containerd runtimes via a "RuntimeClass" [resource](https://kubernetes.io/docs/concepts/containers/runtime-class/) 
+
+After these prerequisites are satisfied, the configuration file for the target  Kata Container, will have to have the flag "SandboxCgroupOnly" set to true. As of Kata 2.0 this is the only way Kata containers can work with the Kubernetes cgroup naming conventions.
+
+   ```toml
+   ...
+   # If enabled, the runtime will add all the kata processes inside one dedicated cgroup.
+   # The container cgroups in the host are not created, just one single cgroup per sandbox.
+   # The runtime caller is free to restrict or collect cgroup stats of the overall Kata sandbox.
+   # The sandbox cgroup path is the parent cgroup of a container with the PodSandbox annotation.
+   # The sandbox cgroup is constrained if there is no container type annotation.
+   # See: https://godoc.org/github.com/kata-containers/runtime/virtcontainers#ContainerType
+   sandbox_cgroup_only=true
+   ...
+   ``` 
+
+### Reference
+
+If you have a pre-existing Kubernetes cluster, for an easy deployement follow this [document](https://github.com/kata-containers/packaging/blob/master/kata-deploy/README.md#kubernetes-quick-start).
+
+
+Starting from scratch:
+
+   * [Kata installation guide](https://github.com/kata-containers/kata-containers/tree/2.0-dev/docs/install#manual-installation)
+   * [Kata Containers + CRI-O](https://github.com/kata-containers/documentation/blob/master/how-to/run-kata-with-k8s.md)
+   * [Kata Containers + containerd](https://github.com/kata-containers/documentation/blob/master/how-to/containerd-kata.md)
+   * [Kubernetes Runtime Class](https://kubernetes.io/docs/concepts/containers/runtime-class/)
+   * [Cgroup and Kata containers](https://github.com/kata-containers/kata-containers/blob/stable-2.0.0/docs/design/host-cgroups.md)
+
+
 ## Using Docker as the Runtime
 
 If you must use `docker` as the runtime then the proxying setup is slightly more
--- a/pkg/cri/resource-manager/control/page-migrate/demoter.go
+++ b/pkg/cri/resource-manager/control/page-migrate/demoter.go
@@ -249,7 +249,7 @@ func resetDirtyBit(pid string) error {
 
 // resetDirtyBit unsets soft-dirty bits for all processes in a container.
 func (d *demoter) resetDirtyBit(c *container) error {
-	pids, err := utils.GetProcessesInContainer(c.GetCgroupParentDir(), c.GetID())
+	pids, err := utils.GetProcessesInContainer(c.GetCgroupParentDir(), c.GetPodID(), c.GetID())
 	if err != nil {
 		return err
 	}
@@ -308,7 +308,7 @@ func (d *demoter) getPagesForContainer(c *container, sourceNodes system.IDSet) (
 		pages:        make(map[int][]page, 0),
 		longestRange: 0,
 	}
-	pids, err := utils.GetProcessesInContainer(c.GetCgroupParentDir(), c.GetID())
+	pids, err := utils.GetProcessesInContainer(c.GetCgroupParentDir(), c.GetPodID(), c.GetID())
 	if err != nil {
 		return pagePool{}, err
 	}
--- a/pkg/cri/resource-manager/control/page-migrate/page-migrate.go
+++ b/pkg/cri/resource-manager/control/page-migrate/page-migrate.go
@@ -73,6 +73,7 @@ type migration struct {
 type container struct {
 	cacheID    string
 	id         string
+	podID      string
 	prettyName string
 	parentDir  string
 	pm         *cache.PageMigrate
@@ -170,6 +171,7 @@ func (m *migration) insertContainer(cc cache.Container) error {
 	c := &container{
 		cacheID:    cc.GetCacheID(),
 		id:         cc.GetID(),
+		podID:      cc.GetPodID(),
 		prettyName: cc.PrettyName(),
 		parentDir:  pod.GetCgroupParentDir(),
 		pm:         pm.Clone(),
@@ -210,6 +212,11 @@ func (m *migration) deleteContainer(cc cache.Container) error {
 // GetCacheID replicates the respective cache.Container function.
 func (c *container) GetCacheID() string {
 	return c.cacheID
+}
+
+// GetPodID replicates the respective cache.Container function.
+func (c *container) GetPodID() string {
+	return c.podID
 }
 
 // GetID replicates the respective cache.Container function.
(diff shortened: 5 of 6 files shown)


klihub (changes requested):
A few general review comments first:

1. Ideally (with very few exceptions to this rule), every commit in your PR should be self-contained, at least from the compilation point of view. IOW, it should be possible to go through your PRs git history step-by-step, type make after every commit, and expect it to compile everything successfully.
2. One implication of that is that every commit in a PR which fixes up typos/thinkos, comment spelling/grammar, etc. of any commit which was introduced by the PR in the first place, should be squashed in the offending commit. => Applying that principle to this PR as an example, here the last two commits should be a single one.
3. Another implication of that is that every commit which changes the signature of a function (or does other similar things like adding/removing members to structures, etc), should also contain any corrective changes necessary to make the code base compile after this change. => Applying that principle to this PR as an example, the second commit which updates the calls to `GetProcessesInContainer()` and `GetTasksInContainer()` should be squashed into the first commit, which was the one that updated the signature in the first place.
4. We keep the bar pretty low for splitting documentation additions into commits of their own. If a commit alters or changes something which would render existing documentation incorrect, then it is practically mandatory to update the docs in the same commit. Otherwise, if you introduce new functionality it is not so clear cut and varies case by case. Here I could see it work in either way.

Then one real review comment related to the functionality:

- I'd be surprised if the current `GetProcessesInContainer()` wrapper passing containerID as the podID really can work correctly. Even if it did by luck, I think the page-migrate controller needs to be properly fixed by adding GetPodID() to the internal minimally reduced container abstraction and saving the pod ID among the cached minimal container data when it gets mirrored, together with the container ID. With that in place, `GetProcessesInContainer()` can be properly analogous to the more correct looking `GetTasksInContainer()`.

marquiz (changes requested):
The documentation does not build (you can try `make site-serve`, for example). Within the current documentation I'd be inclined to add Kata as a separate section under the existing `setup.md`. What do others think? If we add a new `.md` file that needs to be added into some toctree, `index.rst` would be most obvious in which case it would show up as a new item on the main menu level

klihub:
> The documentation does not build (you can try `make site-serve`, for example). Within the current documentation I'd be inclined to add Kata as a separate section under the existing `setup.md`. What do others think? 

I also tend to think a separate section would be warranted for Kata. It is so special compared to the rest of the runtimes.

pkg/utils/cgroups.go
```diff
@@ -80,7 +80,7 @@ func isProcess(processID string) bool {
 	return true
 }
 
-func getTasksInContainer(cgroupParentDir, containerID string, onlyProcesses bool) ([]string, error) {
+func getTasksInContainer(cgroupParentDir, containerID string, podID string, onlyProcesses bool) ([]string, error) {
```
klihub, line 83: nitpick: it could be a bit more consistent with the rest of the functions if we swapped the `containerID, podID` arguments the other way around to `podID, containerID`.

docs/kata-containers.md
```diff
@@ -0,0 +1,48 @@
...
+
+Starting from scratch:
+
+* Kata installation guide: https://github.com/kata-containers/kata-containers/tree/2.0-dev/docs/install#manual-installation
+* Kata Containers + CRI-o: https://github.com/kata-containers/documentation/blob/master/how-to/run-kata-with-k8s.md
+* Kata Containers + containerd: https://github.com/kata-containers/documentation/blob/master/how-to/containerd-kata.md
+* Kubernetes Runtime Class: https://kubernetes.io/docs/concepts/containers/runtime-class/
+* Cgroup and Kata containers: https://github.com/kata-containers/kata-containers/blob/stable-2.0.0/docs/design/host-cgroups.md
+
+
```
klihub, line 48: These seem to use Windows-/DOS-like CR/LF for newlines. Please remove the CR's (ASCII 0x0d).

klihub:
@Dodan #570 adds the missing `container.GetPodID()` support you'll need for proper Kata-enabled `utils.Get{Processes,Tasks}InContainer()` implementations.

klihub:
@Dodan commits [c4b7247462ea64f9e120e47eac81b1a172dd04a9](https://github.com/klihub/cri-resource-manager/commit/c4b7247462ea64f9e120e47eac81b1a172dd04a9) and [ca076ae67ec35acea31e67e79d6042a1900b63dd](https://github.com/klihub/cri-resource-manager/commit/ca076ae67ec35acea31e67e79d6042a1900b63dd) from [this updated branch of yours](https://github.com/klihub/cri-resource-manager/commits/devel/kata-container-support)

  - swap the `podID, containerID` argument to this preferred order, and
  - update the page-migrate controller to pass the pod ID to `utils.GetProcessesInContainer()`

Both of these should be squashed to the preceding Kata-implementation commit.

Of the remaining commits on that branch, the first is a cherry-picked #570. The second one is your original PR with your latter two commits squashed. That is the way how we would like to see the code change parts from that commit.

Dodan:
Thanks for the review guys! I'll try to make the changes as soon as possible.

One more thing that isn't clear yet is documentation:
  * Where should I put the documentation for Kata? (setup.md or leave it separate?)
  * Can I consider that documentation passes is ok if it passes `make site-serve`?

klihub:
> Thanks for the review guys! I'll try to make the changes as soon as possible.
> 
> One more thing that isn't clear yet is documentation:
> 
> * Where should I put the documentation for Kata? (setup.md or leave it separate?)

Leave it separate and try to find a good spot to link to it from elsewhere in the documentation.
This is necessary to pass the CI checks: everything included should be possible to reach by
starting at the TOC/index and following links.

> * Can I consider that documentation passes is ok if it passes `make site-serve`?

Yes. Even building the `html` target which is a dependency of `site-serve` would
strictly be enough as the minimum requirement verification for docs, roughly the
equivalent of code compiling.

`make site-serve` is better because it also allows you to previewing locally how your
documentation updates will look like. Just point your browser at [the locally served docs](http://127.0.0.1:8081/)
and verify the changes before filing a PR.

docs/kata-containers.md
```diff
@@ -0,0 +1,49 @@
+# Kata containers support in CRI-RM
+
+Kata containers represent an alternative runtime to Runc containers,
+and are conceptually defined by a containers which replace their sandbox 
+with a lightweight Virtual Machine.
```
klihub, line 5: Here I'd also link to the main Kata site. I'd also steal the short summary of Kata/Kata Containers from their site, to avoid having to think too hard if the description is right. So something like this:

[Kata Containers](https://katacontainers.io/) is an open source container runtime, building lightweight virtual machines that seamlessly plug into the containers ecosystem.

docs/kata-containers.md
```diff
@@ -0,0 +1,49 @@
...
+
+In order to enable Kata container in a Kubernetes-CRI-RM stack, both Kubernetes 
+and the Container Runtime Engine need to be aware of the new runtime environment:
+
+  * The Container Runtime Engine can only be CRI-o or containerd, and will need to
```
klihub, line 10: s/CRI-o/CRI-O/

docs/kata-containers.md
```diff
@@ -0,0 +1,49 @@
...
+   have the runtimes enabled in their configuration files.
+  * Kubernetes will also have to be made aware via a "RuntimeClass"  resource
+
+After these prerequesites are satisfied, the configuration file for the target  Kata Container, will have to have the flag "SandboxCgroupOnly" to true. As of Kata 2.0 this is the only way Kata containers can with the Kubernetes cgroup naming conventions.
```
klihub, line 14: typos: prerequisites... "SandboxCgroupOnly" set to true... can work with the Kubernetes cgroup naming conventions.

docs/kata-containers.md
```diff
@@ -0,0 +1,49 @@
...
+
+   ```toml
+   ...
+   # if enabled, the runtime will add all the kata processes inside one dedicated cgroup.
```
klihub, line 18: typo: If enabled, ...

docs/kata-containers.md
```diff
@@ -0,0 +1,49 @@
...
+   # The sandbox cgroup path is the parent cgroup of a container with the PodSandbox annotation.
+   # The sandbox cgroup is constrained if there is no container type annotation.
+   # See: https://godoc.org/github.com/kata-containers/runtime/virtcontainers#ContainerType
+   sandbox_cgroup_only=true
+   ...
+   ``` 
+
+Controllers known to be compatible with Kata and CRI-RM:
+   * cri & cgroup controller
+   * RDT controller
```
klihub, line 30: I'd remove this. Commit 85c8d9507894f621e8e635c9caaf3ae0cff490f1 should have fixed the `page-migrate` controller to work s well.

klihub (changes requested):
Looks pretty good now. I'd say almost there. There are a few typos I've spotted in the doc bits that need fixing.

Also, I woudn't mind @marquiz, @kad, and @askervin taking a look at the docs.

Dodan:
Should the typos be solved in a new commit or in the exisiting documentation commit?

klihub:
> Should the typos be solved in a new commit or in the exisiting documentation commit?

Existing commit. There is no value in seeing the intermediate versions with typos.

docs/kata-containers.md
```diff
@@ -0,0 +1,44 @@
+# Kata containers support in CRI-RM
```
marquiz, line 1: i'd suggest to use a shorter title like `Kata Containers`. Especially if this is going to be a separate title on the main level. The headings should be in Capital Case

docs/kata-containers.md
```diff
@@ -0,0 +1,44 @@
...
+
+[Kata Containers](https://katacontainers.io/) is an open source container runtime, 
+building lightweight virtual machines that seamlessly plug into the containers ecosystem.
+
+In order to enable Kata container in a Kubernetes-CRI-RM stack, both Kubernetes 
+and the Container Runtime Engine need to be aware of the new runtime environment:
+
+  * The Container Runtime Engine can only be CRI-O or containerd, and will need to
+   have the runtimes enabled in their configuration files.
+  * Kubernetes will also have to be made aware via a "RuntimeClass"  resource
```
marquiz, line 11: aware ..._of smth_ via

Also, link to kubernetes documentation (about enabling/configuring runtime classes) would be good

docs/kata-containers.md
```diff
@@ -0,0 +1,44 @@
+# Kata containers support in CRI-RM
+
+[Kata Containers](https://katacontainers.io/) is an open source container runtime, 
+building lightweight virtual machines that seamlessly plug into the containers ecosystem.
+
+In order to enable Kata container in a Kubernetes-CRI-RM stack, both Kubernetes 
+and the Container Runtime Engine need to be aware of the new runtime environment:
+
+  * The Container Runtime Engine can only be CRI-O or containerd, and will need to
+   have the runtimes enabled in their configuration files.
```
marquiz, line 10: I think the terminology is a bit misleading here. I consider *container runtime engine* to be the lower level OCI compliant runtime like runc or kata. The upper level compoment (like containerd or CRI-O) would be *container runtime* or *container engine*. Probably *container runtime* in our context. 

What do others think?

docs/kata-containers.md
```diff
@@ -0,0 +1,44 @@
...
+   # The container cgroups in the host are not created, just one single cgroup per sandbox.
+   # The runtime caller is free to restrict or collect cgroup stats of the overall Kata sandbox.
+   # The sandbox cgroup path is the parent cgroup of a container with the PodSandbox annotation.
+   # The sandbox cgroup is constrained if there is no container type annotation.
+   # See: https://godoc.org/github.com/kata-containers/runtime/virtcontainers#ContainerType
+   sandbox_cgroup_only=true
+   ...
+   ``` 
+
+## Kata useful documentation
```
marquiz, line 27: I'd just name this *Reference* or smth

docs/kata-containers.md
```diff
@@ -0,0 +1,44 @@
...
+   * https://github.com/kata-containers/packaging/blob/master/kata-deploy/README.md#kubernetes-quick-start
+
+
+Starting from scratch:
+
+   * Kata installation guide: https://github.com/kata-containers/kata-containers/tree/2.0-dev/docs/install#manual-installation
+   * Kata Containers + CRI-O: https://github.com/kata-containers/documentation/blob/master/how-to/run-kata-with-k8s.md
+   * Kata Containers + containerd: https://github.com/kata-containers/documentation/blob/master/how-to/containerd-kata.md
+   * Kubernetes Runtime Class: https://kubernetes.io/docs/concepts/containers/runtime-class/
+   * Cgroup and Kata containers: https://github.com/kata-containers/kata-containers/blob/stable-2.0.0/docs/design/host-cgroups.md
```
marquiz, line 40: Please make these "clickable" links

marquiz (changes requested):
I'm still pondering should the documentation be just a section under `setup.md` instead of a separate subject on the "main level"? I don't oppose if others think this is the correct location.

klihub:
> I'm still pondering should the documentation be just a section under `setup.md` instead of a separate subject on the "main level"? I don't oppose if others think this is the correct location.

That's a good point. I agree, that should be either added directly a  subsection in `setup.md`, or linked to from that instead of the main `index.rst`.

marquiz:
> > I'm still pondering should the documentation be just a section under `setup.md` instead of a separate subject on the "main level"? I don't oppose if others think this is the correct location.
> 
> That's a good point. I agree, that should be either added directly a subsection in `setup.md`, or linked to from that instead of the main `index.rst`.

Yup. We already have the "Docker as a Runtime" there

docs/kata-containers.md
Dodan: Okay. I thought it would make sense since we are also referencing Kata containers & Runc as being different runtimes for containers, which might create confusion if containerd and CRI-O become container runtimes themselves.

Dodan:
> > > I'm still pondering should the documentation be just a section under `setup.md` instead of a separate subject on the "main level"? I don't oppose if others think this is the correct location.
> > 
> > 
> > That's a good point. I agree, that should be either added directly a subsection in `setup.md`, or linked to from that instead of the main `index.rst`.
> 
> Yup. We already have the "Docker as a Runtime" there

Would it be ok to add the content of the Kata documentation to setup.md, as a new section, called "Kata Containers" before "Using Docker as the Runtime" and after "Using CRI Resource Manager as a Message Dumper"?

klihub:
> > > > I'm still pondering should the documentation be just a section under `setup.md` instead of a separate subject on the "main level"? I don't oppose if others think this is the correct location.
> > > 
> > > 
> > > That's a good point. I agree, that should be either added directly a subsection in `setup.md`, or linked to from that instead of the main `index.rst`.
> > 
> > 
> > Yup. We already have the "Docker as a Runtime" there
> 
> Would it be ok to add the content of the Kata documentation to setup.md, as a new section, called "Kata Containers" before "Using Docker as the Runtime" and after "Using CRI Resource Manager as a Message Dumper"?

Sounds reasonable to me. @marquiz ?

Dodan:
A minor thing I noticed: when the documentation is compiled and the website is served, the links in the Kata Reference section have their ".md" suffix stripped and no longer lead to the right pages. This works from the github documents, but not from the sphinx website. Do you have any ideas how to overcome this?

marquiz:
> > > > > I'm still pondering should the documentation be just a section under `setup.md` instead of a separate subject on the "main level"? I don't oppose if others think this is the correct location.
> > > > 
> > > > 
> > > > That's a good point. I agree, that should be either added directly a subsection in `setup.md`, or linked to from that instead of the main `index.rst`.
> > > 
> > > 
> > > Yup. We already have the "Docker as a Runtime" there
> > 
> > 
> > Would it be ok to add the content of the Kata documentation to setup.md, as a new section, called "Kata Containers" before "Using Docker as the Runtime" and after "Using CRI Resource Manager as a Message Dumper"?
> 
> Sounds reasonable to me. @marquiz ?

Yes, that sounds good to me 👍

marquiz:
> A minor thing I noticed: when the documentation is compiled and the website is served, the links in the Kata Reference section have their ".md" suffix stripped and no longer lead to the right pages. This works from the github documents, but not from the sphinx website. Do you have any ideas how to overcome this?

Argh, not a minor thing, thanks for pointing this out! The reference-mangling kludge in `docs/conf.py` obviously doesn't does the wrong this 🤯 We (i.e. me and @klihub) need to fix this in a separate PR

marquiz:
> The reference-mangling kludge in `docs/conf.py` obviously doesn't does the wrong this 🤯 We (i.e. me and @klihub) need to fix this in a separate PR

It looks like this wasn't a bug in our kludge, but, in the markdown parser that is being used. #580 should fix the issue with the links

klihub (commented):
This LGTM now. @marquiz, how about you ?

marquiz:
Thanks @Dodan for the PR!
lgtm

marquiz (approved):

klihub (approved):

PR #567, opened 2020-12-15
config: handle composite structs in Module.validate()

marquiz:


pkg/config/config.go
```diff
@@ -324,6 +324,28 @@ func (m *Module) check() {
...
+	fields := make(map[string]struct{})
+
+	for i := 0; i < val.NumField(); i++ {
+		f := val.Field(i)
+
+		if f.Type.Kind() == reflect.Struct && f.Anonymous {
+			for name := range getFields(f.Type) {
+				fields[name] = struct{}{}
+
+			}
```
klihub, line 339: nitpick: if you would pass an externally created `fields` map as an argument to `getFields()`, you could avoid collecting to an intermediate map and then copying here...

marquiz: Yeah, I considered that but I thought the function interface was neater and safer this way: you're guaranteed to get a clean map containing the fields. The performance does not matter here. I pushed a patch changing this but I'm not happy with that. WDYT?

klihub: Fair enough. I'm fine with either way. If you're feel safer with the original, we can go with that.

That said, if this happened to be perf-critical, you could have the best of both by splitting out the field collection and recursion to a `collectFields()` with the above suggested signature and have `getFields()` create an empty map and call `collectFields()` on the outermost type with that.

marquiz: > That said, if this happened to be perf-critical, you could have the best of both by splitting out the field collection and recursion

How stupid I was 🤦  Thanks, this ^ inspired me to a solution I like, WDYT about that?

klihub: Yes, I like that better, too. Same idea, without the extra function being visible to the outside.

klihub (approved):

PR #568, opened 2020-12-15
control/rdt: add option to disable rdt monitoring

marquiz:
TODO: Remove existing monitoring groups when changing from enabled to disabled. However, for symmetry, assign() should be called on all existing containers on re-configure.

Based on #567 

Also requires https://github.com/intel/goresctrl/pull/16

### DIFF
--- a/docs/policy/rdt.md
+++ b/docs/policy/rdt.md
@@ -76,6 +76,8 @@ data:
   rdt: |+
     # Common options
     options:
+      # Set to true to disable creation of monitoring groups
+      monitoringDisabled: false
       l3:
         # Make this false if CAT must be available
         optional: true
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/google/go-cmp v0.4.0
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/intel/cri-resource-manager/pkg/topology v0.0.0
-	github.com/intel/goresctrl v0.0.0-20201123171612-0d2cbe434ecd
+	github.com/intel/goresctrl v0.0.0-20201216151002-453bb7ae55ff
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.8.0
 	github.com/prometheus/client_model v0.2.0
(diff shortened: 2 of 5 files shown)


klihub (approved):
LGTM.

With the monitoring state transitions and any state changes related to configuration in general, perhaps the cleanest would be to add a `(ctl *rdtctl) reconfigure()` or similar function. This would go through all the running containers found in the cache and perform any actions which is considered necessary for the new configuration. This would at least create/delete monitoring groups as necessary. But it could also do other things, for instance trying to re-`assign()` a container to its class to handle cases where an assigned non-existent class has been created by the updated configuration. If that behavior is considered to be semantically more correct than the current one.

klihub (approved):

marquiz:
> With the monitoring state transitions and any state changes related to configuration in general, perhaps the cleanest would be to add a `(ctl *rdtctl) reconfigure()` or similar function. This would go through all the running containers found in the cache and perform any actions which is considered necessary for the new configuration. This would at least create/delete monitoring groups as necessary. But it could also do other things, for instance trying to re-`assign()` a container to its class to handle cases where an assigned non-existent class has been created by the updated configuration. If that behavior is considered to be semantically more correct than the current one.

Yup. I'm working on this.

We could merge this PR as is for now and the reconfigure part in a separate PR. Or, we can hold this until I add the reconfigure part. I'm ok with both options. Merging this now would enable rudimentary support for testing purposes.

klihub:
> > With the monitoring state transitions and any state changes related to configuration in general, perhaps the cleanest would be to add a `(ctl *rdtctl) reconfigure()` or similar function. This would go through all the running containers found in the cache and perform any actions which is considered necessary for the new configuration. This would at least create/delete monitoring groups as necessary. But it could also do other things, for instance trying to re-`assign()` a container to its class to handle cases where an assigned non-existent class has been created by the updated configuration. If that behavior is considered to be semantically more correct than the current one.
> 
> Yup. I'm working on this.
> 

ACK.

Wrt. reconfiguring, one thing I forgot to mention is that I think the very first configuration change notification callback will hit the controller before it has been `Start()`ed. At that phase the controller has not been passed the cache yet, so it will need to skip trying to do anything that would require it and delay that, preferably only until `Start()`.

> We could merge this PR as is for now and the reconfigure part in a separate PR. Or, we can hold this until I add the reconfigure part. I'm ok with both options. Merging this now would enable rudimentary support for testing purposes.

I think it would be better to merge this as such and have a separate PR for the reconfiguration part.

marquiz:
> I forgot to mention is that I think the very first configuration change notification callback will hit the controller before it has been `Start()`

That's not the case anymore as we register the config notifier in `Start()`

PR #569, opened 2020-12-15
config: add support for promoted struct fields.

klihub:
This patch adds support for handling unambiguous promoted fields in configuration data. It also reject configuration data that tries to set ambiguous conflicting promoted fields with a diagnostic error message pointing out the reason for rejection. Otherwise such errors would simply be silently ignored by the JSON/YAML marshallers and unmarshallers.

Currently this patch does not handle golang's explicit disambiguation of conflicting fields (which is done by providing a JSON tag only for the winning one).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -16,8 +16,10 @@ package config
 
 import (
 	"reflect"
-	"sigs.k8s.io/yaml"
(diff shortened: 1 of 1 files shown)


klihub:
Closing in favor of #567

PR #570, opened 2020-12-16
page-migrate: add cache-like container.GetPodID().

klihub:
Storing and retrieving the Pod ID is needed for properly discovering
the process IDs of tasks/processes belonging to a Kata container in
cgroupfs.


askervin (approved):
LGTM.

PR #572, opened 2020-12-17
control/rdt: fix and simplify handling of implicit disabling

marquiz:
- Take into account that the system root class always exists if rdt has
been initialized, even if the configuration contains no classes.
- Remove the special handling for class names matching Pod QoS class
names. Either the controller is disabled or not, the controller should
be agnostic to the class name.

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -59,6 +59,9 @@ const (
 	// ToptierLimitKey is the pod annotation key for specifying container top tier memory limits.
 	ToptierLimitKey = "toptierlimit" + "." + kubernetes.ResmgrKeyNamespace
 
+	// RDTClassPodQoS denotes that the RDTClass should be taken from PodQosClass
+	RDTClassPodQoS = "/PodQos"
+
 	// ToptierLimitUnset is the reserved value for indicating unset top tier limits.
 	ToptierLimitUnset int64 = -1
 
--- a/pkg/cri/resource-manager/cache/container.go
+++ b/pkg/cri/resource-manager/cache/container.go
@@ -189,7 +189,7 @@ func (c *container) fromListResponse(lrc *cri.Container) error {
 func (c *container) setDefaults() error {
 	class, ok := c.GetEffectiveAnnotation(RDTClassKey)
 	if !ok {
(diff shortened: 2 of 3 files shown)


klihub:
> * Remove the special handling for class names matching Pod QoS class
>   names. Either the controller is disabled or not, the controller should
>   be agnostic to the class name.

Hmm... I guess, this is fine for now. Eventually we need to update this to
achieve the intention (as opposed to the implementation effects) of the
original code. The intention there was to not give errors when the controller
is implicitly disable (IOW, the default case) and the assigned class is the
default fallback (IOW, the default case).

With the currently merged code for `cache.Container`, the latter condition
above was not possible to verify, so it was (admittedly poorly) approximated
by simply testing if the class name was one of the defaults. A much better
approximation would have been something like
`c.GetRDTClass() == string(c.GetQOSClass()`
If you are not too strongly against it, I'd like to get that approximation back
in this PR as well.

Of course, the proper way would be to provide testability for whether
the class was set as the fallback default, if it was set from an annotation,
or if it was set explicitly by the active policy. I have an initial patch doing
that. I'll clean it up and file a PR for it (and a similar one for Block I/O).
If/once that is merged, we can revisit this and make sure that at least the
user will be given an error if an explicitly set non-default class cannot be
assigned (IOW, the configuration is inconsistent), and at least an warning
if an explicitly annotated class cannot be assigned (IOW, the user has
false assumptions about the configuration).

WDYT ?

marquiz:
Makes sense, we don't want to be printing errors/warnings if the user is running cri-rm with a default (empty) config. I made another quite straightforward attempt in the latest version. (includes #573)

klihub:
> Makes sense, we don't want to be printing errors/warnings if the user is running cri-rm with a default (empty) config. I made another quite straightforward attempt in the latest version. (includes #573)

I think this is good enough. If we ever merge tag-source checking PR I mentioned earlier, then we can check if we want to revisit this and get rid of the magic value.

klihub (approved):

marquiz:
> If we ever merge tag-source checking PR I mentioned earlier, then we can check if we want to revisit this and get rid of the magic value.

Yeah, sure. If that becomes available we should use it here and prolly drop the magic value

PR #574, opened 2020-12-17
control/rdt: implement assignAll()

marquiz:
- `assignAll` is basically a controller-level (re-)configuration handler. Takes
care of re-assigning all containers as the set of available classes
might have changed and/or monitoring might have been enabled or
disabled.
- Only remove monitoring group of a container if it exist in a wrong class
(the container rdt class was changed) or if monitoring is set to
disabled. This way we don't unnecessarily reset the monitoring data and
history is preserved.


klihub (approved):
Looking good...

PR #577, opened 2020-12-18
control/rdt: implement operating modes

marquiz:
Add a new 'mode' rdt config option to control the operation mode of the
controller. There are three possible modes:

- Disabled:  RDT controller is effectively disabled and containers will
             not be assigned. When activating this mode, the controller
             applies an empty rdt configuration removing all cri-rm
             specific groups from the resctrl filesystem.
- Discovery: RDT controller detects existing non-cri-rm specific
             classes from the resctrl filesystem and uses these.
             CRI-RM will treat the configuration of these as read-only
             and does not alter it.  Similarly to 'Disabled' mode, all
             cri-rm specific groups from the resctrl filesystem are
             dropped when this mode is activated.
- Full:      Full operating mode. RDT controller behaves like
             previously. This is the default.

Depends on https://github.com/intel/goresctrl/pull/21

### DIFF
--- a/sample-configs/cri-resmgr-configmap.example.yaml
+++ b/sample-configs/cri-resmgr-configmap.example.yaml
@@ -297,6 +297,8 @@ data:
   rdt: |+
     # Common options
     options:
+      # Mode must be one of Full, Discovery or Disabled. Defaults to Full.
+      mode: Full
       monitoringDisabled: false
       l3:
         optional: true
--- a/docs/policy/rdt.md
+++ b/docs/policy/rdt.md
@@ -36,11 +36,31 @@ the builtin policies do that.
 
 ## Configuration
 
-The RDT configuration in CRI-RM is a two-level hierarchy consisting of
+### Operating Modes
+
+The RDT controller supports three operating modes, controlled by
+`rdt.options.mode` configuration option.
+
+- Disabled: RDT controller is effectively disabled and containers will not be
+  assigned and no monitoring groups will be created. Upon activation of this
+  mode all CRI-RM specific control and monitoring groups from the resctrl
+  filesystem are removed.
+- Discovery: RDT controller detects existing non-CRI-RM specific classes from
+  the resctrl filesystem and uses these. The configuration of the discovered
+  classes is considered read-only and it will not be altered. Upon activation
+  of this mode all CRI-RM specific control groups from the resctrl filesystem
+  are removed.
+- Full: Full operating mode. The controller manages the configuration of the
+  resctrl filesystem according to the rdt class definitions in the CRI-RM
+  configiration. This is the default operating mode.
+
+### RDT Classes
+
+The RDT class configuration in CRI-RM is a two-level hierarchy consisting of
 partitions and classes. It specifies a set of partitions each having a set of
 classes.
 
-### Partitions
+#### Partitions
 
 Partitions represent a logical grouping of the underlying classes, each
 partition specifying a portion of the available resources (L3/MB) which will be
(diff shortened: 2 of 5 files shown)


sample-configs/cri-resmgr-configmap.example.yaml
```diff
@@ -297,6 +297,8 @@ data:
   rdt: |+
     # Common options
     options:
+      # Mode must be one of Full, Discovery or Disabled
```
klihub, line 300: Should we mention here that the operating mode defaults to `Full` ? Otherwise I'd suspect most people would assume it to be 'Disabled'. AFAICT it's not mentioned in the docs either. And even if it was, for those 99%'s sake who look at sample configs instead of reading the docs, I'd still put it also here...

marquiz: Very good point, fixed

marquiz:
Worth noting: I couldn't decide to do with the `idle` flag in the discovery mode. I..e whether/when to print out warnings of missing classes when trying to assign the container to the "factory default" i.e. Pod QoS class. Now we always print warnings but that's possibly not the right thing to do 🤔  WDYT?

klihub:
> Worth noting: I couldn't decide to do with the `idle` flag in the discovery mode. I..e whether/when to print out warnings of missing classes when trying to assign the container to the "factory default" i.e. Pod QoS class. Now we always print warnings but that's possibly not the right thing to do  WDYT?

Good question. If we end up being idle in discovery mode, that so suspicious that I'd print out an error or warning for that during (re)configuration. But I don't know whether based on that we should silence later errors during assignment failures to default classes. I see a some good points with both possible choices, so I don't know.

Generally my suggestion would be that first we take some time to play a bit more actively around with the full chain, both active policy and meaningful RDT configuration in place, and then see whether we feel that overall configurability, the built in defaults, etc. need some more work, and if they do, in which components. Also, just as an example, I have an itching suspicion that it should be possible to have a setup where the policy placement decision (basically pool assignment and allocation type) have a direct effect on the assigned RDT class. But we need to collect more practical experience and preferably data to really know.

marquiz:
> Good question. If we end up being idle in discovery mode, that so suspicious that I'd print out an error or warning for that during (re)configuration. But I don't know whether based on that we should silence later errors during assignment failures to default classes. I see a some good points with both possible choices, so I don't know.

Sleeping over this and taking a fresh look I start to be inclined towards disabling warnings (i.e. "enablng idle") in case none of Pod QoS class equivalens exists. The use case I'm thinking of would be to when the user wants to disable the default assignment and thus

- just simplifying the configuration
- be sparing on  CLOSes, the number of available CLOSes is very limited anyway


> Generally my suggestion would be that first we take some time to play a bit more actively around with the full chain, both active policy and meaningful RDT configuration in place, and then see whether we feel that overall configurability, the built in defaults, etc. need some more work, and if they do, in which components. Also, just as an example, I have an itching suspicion that it should be possible to have a setup where the policy placement decision (basically pool assignment and allocation type) have a direct effect on the assigned RDT class. But we need to collect more practical experience and preferably data to really know.

Probably, but that should mostly concern the policy and it's configuration. Not necessarily the RDT controller. Of course, the "idle" logic might need revisiting which might be the case anyway as the current concept is somewhat unclear as can be figured out from this discussion 🙂

klihub:
> Sleeping over this and taking a fresh look I start to be inclined towards disabling warnings (i.e. "enablng idle") in case none of Pod QoS class equivalens exists.

I'm fine with that.

klihub (approved):
LGTM.

PR #579, opened 2020-12-21
memtier: don't imply error by signature for functions that never fail.

klihub:


marquiz (approved):
Good streamlining 👍 
lgtm

PR #581, opened 2020-12-22
rdt: accept missing default classes in Discovery mode

marquiz:
Suppress warnings about missing classes for RDT default assignments in
Discovery mode if all of the Pod QoS class equivalents (BestEffort,
Burstable, Guaranteed) are missing from the configuration. That is, none
of the Pod QoS class equivalents are found pre-configured in the resctrl
filesystem.

### DIFF
--- a/pkg/cri/resource-manager/control/rdt/rdt.go
+++ b/pkg/cri/resource-manager/control/rdt/rdt.go
@@ -17,15 +17,16 @@ package rdt
 import (
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
+
+	pkgcfg "github.com/intel/cri-resource-manager/pkg/config"
 	"github.com/intel/cri-resource-manager/pkg/cri/client"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/control"
-	"github.com/intel/cri-resource-manager/pkg/utils"
-	"github.com/intel/goresctrl/pkg/rdt"
-
-	pkgcfg "github.com/intel/cri-resource-manager/pkg/config"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
 	"github.com/intel/cri-resource-manager/pkg/metrics"
+	"github.com/intel/cri-resource-manager/pkg/utils"
+	"github.com/intel/goresctrl/pkg/rdt"
 )
 
 const (
@@ -40,10 +41,10 @@ const (
 
 // rdtctl encapsulates the runtime state of our RTD enforcement/controller.
 type rdtctl struct {
-	cache cache.Cache   // resource manager cache
-	idle  bool          // true if we run without any classes configured
-	mode  OperatingMode // track the mode here to capture mode changes
-	opt   *config
+	cache        cache.Cache   // resource manager cache
+	noQoSClasses bool          // true if we run without any classes configured
+	mode         OperatingMode // track the mode here to capture mode changes
+	opt          *config
 }
 
 type config struct {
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/control/rdt/rdt.go
```diff
@@ -303,7 +304,17 @@ func (ctl *rdtctl) configure() error {
 		if err := rdt.DiscoverClasses(""); err != nil {
 			return rdtError("failed to discover classes from fs: %v", err)
 		}
-		ctl.idle = false
+
+		// Set idle to true if none of the Pod QoS class equivalents exist
+		ctl.idle = true
```
klihub, line 309: Let's not call this `idle` if we now use it like this. It is pretty misleading, probably implying for most casual readers that there are no classes defined whatsoever. Maybe something like `noQoSClasses` would be more apt.

Probably just because I'm too ignorant about RDT/resctrlfs, I have another question related to the recent assignment changes in this controller: What is the practical difference between leaving something completely unassigned as opposed to assigning it to the root class ? Or to put it another way, why did we choose to assign containers with the empty ("") RDT class to the root, but leave classes with non-existing QoS-derived classes unassigned ?

marquiz: > Let's not call this `idle` if we now use it like this. It is pretty misleading, probably implying for most casual readers that there are no classes defined whatsoever. Maybe something like `noQoSClasses` would be more apt.

Agree, I'll come up with a better name

> Probably just because I'm too ignorant about RDT/resctrlfs, I have another question related to the recent assignment changes in this controller: What is the practical difference between leaving something completely unassigned as opposed to assigning it to the root class ? Or to put it another way, why did we choose to assign containers with the empty ("") RDT class to the root, but leave classes with non-existing QoS-derived classes unassigned ?

Because `""` will always be explicit assignment by the user (by using an annotation) – it could also be the policy but still that would be explicit. Pod QoS is the implicit default that is being set in the cache. Makes sense?

klihub: Yes, that part I understand. But what is the effective difference from the RDT/resctrlfs point of view between the unassigned and the root-assigned ones ? Which CLOS the unassigned ends up using ? Isn't it automatically added to the root then ?

klihub: Hmm... so I guess it inherits the CLOS from it's parent process over fork. If it is so, then wouldn't an unassigned container end up (very probably) running with the CLOS of the container runtime/daemon itself ? And if yes, then isn't that's something we want to actively avoid ?

marquiz: > Yes, that part I understand. But what is the effective difference from the RDT/resctrlfs point of view between the unassigned and the root-assigned ones ? Which CLOS the unassigned ends up using ? Isn't it automatically added to the root then ?

But couldn't the class be changed dynamically by editing the pod annotation?

> Hmm... so I guess it inherits the CLOS from its parent process over fork. If it is so, then wouldn't an eventually unassigned container end up (very probably) running with the CLOS of the container runtime/daemon itself ? And if yes, then isn't that's something we want to actively avoid, for instance by treating them the same way as containers assigned to the "" class ?

Hmm, yes, that is actually a very good point! Unassigned will currently inherit the resctrl group of the container runtime 😬  That needs to be changed like you described: if a class is missing assign it to the root (unless RDT is in `Disabled` mode)

marquiz: Renamed the `idle` flag to `noQoSClasses`.

I think addressing the default assignment to root class should be handled in a separate PR

klihub (approved):

PR #582, opened 2020-12-22
docs: refer to v0.4.1 in installation instructions

marquiz:


klihub (approved):

klihub (dismissed):

klihub (commented):
So for making a maintenance release, now you need to add to the docs a tag which does not exist yet ? And then tag the release. And if something went wrong, then update the docs, and tag it again ?

marquiz:
> So for making a maintenance release, now you need to add to the docs a tag which does not exist yet ? And then tag the release. And if something went wrong, then update the docs, and tag it again ?

The first part, yes: the references need to be updated before the tag exists. It's a chicken-and-egg problem.

Regarding the second question, I don't see why you'd need to re-tag?

Going forward we need at least to semi-automate this.

docs/installation.md
```diff
@@ -5,7 +5,7 @@
 You can install CRI Resource Manager from `deb` or `rpm` packages
 for supported distros.
 
-  - [download](https://github.com/intel/cri-resource-manager/releases) packages
+  - [download](https://github.com/intel/cri-resource-manager/releases/tag/v0.4.1) packages
```
klihub, line 8: So, before we make a point release, are we supposed to change the docs to point to a tag that does not exist yet ? Then tag the release ? And if something goes wrong, update the docs again and repeat ?

docs/quick-start.md
```diff
@@ -17,14 +17,14 @@ First, install and setup cri-resource-manager.
 #### CentOS, Fedora and SUSE
 
 ```
-CRIRM_VERSION=0.4.0
+CRIRM_VERSION=0.4.1
```
klihub, line 20: A similar comment applies here than for the [related PR to the main branch](https://github.com/intel/cri-resource-manager/pull/583#discussion_r547535884)...

klihub:
> > So for making a maintenance release, now you need to add to the docs a tag which does not exist yet ? And then tag the release. And if something went wrong, then update the docs, and tag it again ?

There are already number of manual verification steps in the release process we do before we can publish a tagged release. If something turns out to be wrong and you already have pushed the tag (which you had to,  otherwise you can't proceed with the preparation and checks), then after having tried to fix the problem you have two choices left: 1) delete the tag then re-tag, which is not the safest/most social behavior as someone might have already pulled it, or 2) roll a new tag and start redoing the manual verification steps.

So the basic point of my comment is that the more manual steps/things to manually verify we add to the process, the more probably something goes wrong. 

> 
> The first part, yes: the references need to be updated before the tag exists. It's a chicken-and-egg problem.
> 
> Regarding the second question, I don't see why you'd need to re-tag?
> 
> Going forward we need at least to semi-automate this.

I agree with that. It'd be nice to be able to run something like `./scripts/release/prepare-release --version $NEW_VERSION --sha1 $GIT_SHA1_OF_RESULTING_TAG`, or a makefile target to that effect, which would then locally add the new annotated tag and run automated tests for as many of the release preparation tests as possible.

That said we'll not be able to get rid of all the problems, especially the ones that might show up in a workflow triggered by pushing the new tag.

marquiz:
> There are already number of manual verification steps in the release process we do before we can publish a tagged release. If something turns out to be wrong and you already have pushed the tag (which you had to, otherwise you can't proceed with the preparation and checks), then after having tried to fix the problem you have two choices left: 1) delete the tag then re-tag, which is not the safest/most social behavior as someone might have already pulled it, or 2) roll a new tag and start redoing the manual verification steps.

I understand. But in principle I don't think this particular case is much different from other bugs or documentation mistakes. You only can fix issues after the fact by creating a new release.

> I agree with that. It'd be nice to be able to run something like `./scripts/release/prepare-release --version $NEW_VERSION --sha1 $GIT_SHA1_OF_RESULTING_TAG`, or a makefile target to that effect, which would then locally add the new annotated tag and run automated tests for as many of the release preparation tests as possible.

We should definitely move into that direction, and explore if e.g. templating could be used with Sphinx so that the release number would be specified in a single place.

Nevertheless, I think we should address the imminent problem by merging this PR and then do the further release process refinements in master and in a separate PR as that will be a bigger task

klihub (approved):
Fair enough. Let's merge this.

PR #583, opened 2020-12-22
docs: refer to the latest release in the installation instructions

marquiz:


docs/quick-start.md
```diff
@@ -17,14 +17,14 @@ First, install and setup cri-resource-manager.
 #### CentOS, Fedora and SUSE
 
 ```
-CRIRM_VERSION=0.4.0
+CRIRM_VERSION=0.4.1
```
klihub, line 20: Wouldn't it be better to have here something relatively easy to understand yet not any particular exact version ? Maybe something like `$YOUR_VERSION_OF_CHOICE` or similar. Otherwise making a point release of the latest maintenance branch will always force us to make an additional commit to the main branch, just to keep these in sync. Which IMO is just an artificially created thing waiting to go wrong.

klihub: Okay, so how about changing this to something like this instead ?
```
CRIRM_VERSION=$(curl -Ls -o /dev/null -w '%{url_effective}\n' https://github.com/intel/cri-resource-manager/releases/latest | sed 's:.*/v::')
```

marquiz: > Maybe something like `$YOUR_VERSION_OF_CHOICE` or similar.

I think we need to think about the user/reader here instead of our convenience. Especially the quick start instructions should be point-and-shoot style without any additional manual operations or research by the user.

> Okay, so how about changing this to something like this instead

Inspired by this I changed to a curl based approach.

marquiz: Hmm, the latest version relies on `jq` which might not be the best thing. WDYT?

klihub: It doesn't matter too much if the distro has `jq` in the stock repo. IIRC, for CentOS 7 that's not true, but we should anyway encourage folks to move away from that, to 8 if nothing else. Personally I'd prefer curl just because it has both smaller and  more common dependency footprint, and is also more likely to be already installed. Was/is there any particular reason not to go with the curl-based suggestion ?

klihub: If the URI was the concern, I think the stability of the latest release URI is considered part of the [stable github API](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/linking-to-releases):
```
Alternatively, right click Latest Release and copy the URL to share it. The suffix of this URL is always /releases/latest.
```

marquiz: I think the jq-based one is simpler and thus easier to read/decipher for the casual reader. Also, I think relying on the github API might be better. I think `jq` could be replaced by `grep` if needed

klihub: > I think the jq-based one is simpler and thus easier to read/decipher for the casual reader. Also, I think relying on the github API might be better. I think `jq` could be replaced by `grep` if needed

I'm fine with going with any of these, even `jq`. Doesn't matter too much. All of these beat having to manually maintain an extra/artificial dependency.

klihub (approved):

PR #587, opened 2020-12-29
e2e: set imagePullPolicy on every test pod

askervin:
- Missing imagePullPolicy can cause false negative results.
- Tests pull images only IfNotPresent.
- Reorder lines in pod yamls so that "image" is followed by
  "imagePullPolicy".


klihub (approved):
LGTM.

PR #588, opened 2020-12-31
memtier: unify syntax for annotated preferences

klihub:
Prefer using the 'effective' syntax for isolated exclusive CPU, shared CPU, memory type
and cold start period annotated preferences. Remove support for 'elevation' in shared CPU
annotations. It was never implemented.


kad:
Corresponding documentation/examples update ? Otherwise we will get out of sync with code/docs.

klihub:
> Corresponding documentation/examples update ? Otherwise we will get out of sync with code/docs.

Updated annotation-related parts of the documentation.

askervin (approved):
LGTM!

Passes both current and updated E2E tests in PR #592.

PR #590, opened 2021-01-07
memtier: more regular annotation interpretation for CPU allocation preferences.

klihub:
This PR is based on/requires PR #588.

Update the interpretation of shared/isolated CPU opt-in and opt-out annotations
(and global policy defaults) to produce less irregular and hopefully more intuitive
CPU allocation preferences.

TODO/Question:
Would it provide better usability to use a single explicit non-boolean annotation
(for instance `cpu-preferences`) to express all CPU allocation preferences ?

### DIFF
--- a/docs/policy/memtier.md
+++ b/docs/policy/memtier.md
@@ -1,70 +1,306 @@
 # Memtier Policy
 
+## Background
+
+On server-grade hardware the CPU cores, I/O devices and other peripherals
+form a rather complex network together with the memory controllers, the
+I/O bus hierarchy and the CPU interconnect. When a combination of these
+resources are allocated to a single workload, the performance of that
+workload can vary greatly, depending on how efficiently data is transferred
+between them or, in other words, on how well the resources are aligned.
+
+There are a number of inherent architectural hardware properties that,
+unless properly taken into account, can cause resource misalignment and
+workload performance degradation. There are a multitude of CPU cores
+available to run workloads. There are a multitude of memory controllers
+these workloads can use to store and retrieve data from main memory. There
+are a multitude of I/O devices attached to a number of I/O buses the same
+workloads can access. The CPU cores can be divided into a number of groups,
+with each group having different access latency and bandwidth to each
+memory controller and I/O device.
+
+If a workload is not assigned to run with a properly aligned set of CPU,
+memory and devices, it will not be able to achieve optimal performance.
+Given the idiosyncrasies of hardware, allocating a properly aligned set
+of resources for optimal workload performance requires identifying and
+understanding the multiple dimensions of access latency locality present
+in hardware or, in other words, hardware topology awareness.
+
 ## Overview
 
-The `memtier` policy extends the `topology-aware` policy. It supports
-the same features and configuration options, such as `topology hints`
-and `annotations`, which the `topology-aware` policy does. Please see
-the [documentation for topology-aware
-policy](topology-aware.md) for the description of how
-`topology-aware`policy works and how it is configured.
+The memtier policy automatically builds a tree of pools based on the detected
+hardware topology. Each pool has a set of CPUs and memory zones assigned as
+their resources. Resource allocation for workloads happens by first picking
+the pool which is considered to fit the best the resource requirements of
+the workload and then assigning CPU and memory from this pool.
 
-The main goal of `memtier` policy is to let workloads choose the kinds
-of memory it wants to use. The `topology-aware` policy scoring algorithm
-for selecting topology nodes is changed so that a workload can belong to
-both a CPU node and a memory node in the topology tree -- the CPU
-allocation is reserved from the CPU node and the memory controllers are
-selected from the memory node. Typically the aim is that the CPU and
-memory allocations are done from the same node so that the memory
-locality is as good as possible, but the memory allocation may happen
-also from a wider pool of memory controllers if the amount of free
-memory on a topology node is too low.
+The pool nodes at various depths from bottom to top represent the NUMA nodes,
+dies, sockets, and finally the whole of the system at the root node. Leaf NUMA
+nodes are assigned the memory behind their controllers / zones and CPU cores
+with the smallest distance / access penalty to this memory. If the machine
+has multiple types of memory separately visible to both the kernel and user
+space, for instance both DRAM and [PMEM](https://www.intel.com/content/www/us/en/products/memory-storage/optane-dc-persistent-memory.html), each zone of special type of memory
+is assigned to the closest NUMA node pool.
 
-## Activation of the Memtier Policy
+Each non-leaf pool node in the tree is assigned the union of the resources of
+its children. So in practice, dies nodes end up containing all the CPU cores
+and the memory zones in the corresponding die, sockets nodes end up containing
+the CPU cores and memory zones in the corresponding socket's dies, and the root
+ends up containing all CPU cores and memory zones in all sockets.
 
-You can activate the `memtier` policy by setting `--policy` parameter of
-`cri-resmgr` to `memtier`. For example:
+With this setup, each pool in the tree has a topologically aligned set of CPU
+and memory resources. The amount of available resources gradually increases in
+the tree from bottom to top, while the strictness of alignment is gradually
+relaxed. In other words, as one moves from bottom to top in the tree, it is
+getting gradually easier to fit in a workload, but the price paid for this is
+a gradually increasing maximum potential cost or penalty for memory access and
+data transfer between CPU cores.
 
+Another property of this setup is that the resource sets of sibling pools at
+the same depth in the tree are disjoint while the resource sets of descendant
+pools along the same path in the tree partially overlap, with the intersection
+decreasing as the the distance between pools increases. This makes it easy to 
+isolate workloads from each other. As long as workloads are assigned to pools
+which has no other common ancestor than the root, the resources of these
+workloads should be as well isolated from each other as possible on the given
+hardware.
+
+With such an arrangement, the memtier policy should handle topology-aware
+alignment of resources without any special or extra configuration. When
+allocating resources, the policy
+
+  - filters out all pools with insufficient free capacity
+  - runs a scoring algorithm for the remaining ones
+  - picks the one with the best score
+  - assigns resources to the workload from there
+
+Although the details of the scoring algorithm are subject to change as the
+implementation evolves, its basic principles are roughly
+
+  - prefer pools lower in the tree, IOW stricter alignment and lower latency
+  - prefer idle pools over busy ones, IOW more remaining free capacity and
+    fewer workloads
+  - prefer pools with better overall device alignment
+
+## Features
+
+The memtier policy has the following features:
+
+  - topologically aligned allocation of CPU and memory
+    * assign CPU and memory to workloads with tightest available alignment
+  - aligned allocation of devices
+    * pick pool for workload based on locality of devices already assigned
+  - shared allocation of CPU cores
+    * assign workload to shared subset of pool CPUs
+  - exclusive allocation of CPU cores
+    * dynamically slice off CPU cores from shared subset and assign to workload
+  - mixed allocation of CPU cores
+    * assign both exclusive and shared CPU cores to workload
+  - discovering and using kernel-isolated CPU cores (['isolcpus'](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html#cpu-lists))
+    * use kernel-isolated CPU cores for exclusively assigned CPU cores
+  - exposing assigned resources to workloads
+  - notifying workloads about changes in resource assignment
+  - dynamic relaxation of memory alignment to prevent OOM
+    * dynamically widen workload memory set to avoid pool/workload OOM
+  - multi-tier memory allocation
+    * assign workloads to memory zones of their preferred type
+    * the policy knows about three kinds of memory:
+      - DRAM is regular system main memory
+      - PMEM is large-capacity memory, such as
+        [Intel® Optane™ memory](https://www.intel.com/content/www/us/en/products/memory-storage/optane-dc-persistent-memory.html)
+      - [HBM](https://en.wikipedia.org/wiki/High_Bandwidth_Memory) is high speed memory,
+        typically found on some special-purpose computing systems
+  - cold start
+    * pin workload exclusively to PMEM for an initial warm-up period
+  - dynamic page demotion
+    * forcibly migrate read-only and idle container memory pages to PMEM
+
+## Activating the Policy
+
+You can activate the `memtier` policy by using the following configuration
+fragment in the configuration for `cri-resmgr`:
+
+```yaml
+policy:
+  Active: memtier
+  ReservedResources:
+    CPU: 750m
 ```
-cri-resmgr --policy memtier --reserved-resources cpu=750m
-```
 
-## Configuration
+## Configuring the Policy
 
-The `memtier` policy knows of three kinds of memory: `DRAM`, `PMEM`, and
-`HBM`. The various memory types are accessed via separate memory controllers.
+The policy has a number of configuration options which affect its default behavior.
+These options can be supplied as part of the
+[dynamic configuration](../setup.md#using-cri-resource-manager-agent-and-a-configmap)
+received via the [`node agent`](../node-agent.md), or in a fallback or forced
+[configuration file](../setup.md#using-a-local-configuration-from-a-file). These
+configuration options are
 
-  * DRAM (dynamic random-access memory) is regular system main memory.
-  * PMEM (persistent memory) is large-capacity memory, such as
-    Intel® Optane™ memory.
-  * HBM (high-bandwidth memory) is high speed memory, typically found
-    on some special-purpose computing systems.
+  - `PinCPU`
+    * whether to pin workloads to assigned pool CPU sets
+  - `PinMemory`
+    * whether to pin workloads to assigned pool memory zones
+  - `PreferIsolatedCPUs`
+    * whether isolated CPUs are preferred by default for workloads that are
+      eligible for exclusive CPU allocation
+  - `PreferSharedCPUs`
+    * whether shared allocation is preferred by default for workloads that
+      would be otherwise eligible for exclusive CPU allocation
 
-In order to configure a container to use a certain memory type, use the
-`memory-type.cri-resource-manager.intel.com` effective annotation in the pod
-spec. For example, to make `container1` request both `PMEM` and `DRAM` memory
-types, you could use pod metadata such as this:
+## Policy CPU Allocation Preferences
 
-```
+There are a number of workload properties the memtier policy actively checks to
+decide if the workload could potentially benefit from extra resource allocation
+optimizations. Unless configured differently, containers fulfilling certain
+corresponding criteria are considered eligible for these optimizations. This
+will be reflected in the assigned resources whenever that is possible at the
+time the container's creation / resource allocation request hits the policy.
+
+The set of these extra optimizations consist of
+
+  - assignment of `kube-reserved` CPUs
+  - assignment of exclusively allocated CPU cores
+  - usage of kernel-isolated CPU cores (for exclusive allocation)
+
+The policy uses a combination of the QoS class and the resource requirements of
+the container to decide if any of these extra allocation preferences should be
+applied. Containers are divided into five groups, with each group having a
+slightly different set of criteria for eligibility.
+
+  - `kube-system` group
+    * all containers in the `kube-system` namespace
+  - `low-priority` group
+    * containers in the `BestEffort` or `Burstable` QoS class
+  - `sub-core` group
+    * Guaranteed QoS class containers with `CPU request < 1 CPU`
+  - `mixed` group
+    * Guaranteed QoS class containers with `1 <= CPU request < 2`
+  - `multi-core` group
+    * Guaranteed QoS class containers with `CPU request >= 2`
+
+The eligibility rules for extra optimization are slightly different among these
+groups.
+
+  - `kube-system`
+    * not eligible for extra optimizations
+    * eligible to run on `kube-reserved` CPU cores
+    * always run on shared CPU cores
+  - `low-priority`
+    * not eligible for extra optimizations
+    * always run on shared CPU cores
+  - `sub-core`
+    * not eligible for extra optimizations
+    * always run on shared CPU cores
+  - `mixed`
+    * by default eligible for exclusive and isolated allocation
+    * not eligible for either if `PreferSharedCPUs` is set to true
+    * not eligible for either if annotated to opt out from exclusive allocation
+    * not eligible for isolated allocation if annotated to opt out
+  - `multi-core`
+    * CPU request fractional (`(CPU request % 1000 milli-CPU) != 0`):
+      - by default not eligible for extra optimizations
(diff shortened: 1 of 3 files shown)


pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
```diff
@@ -217,62 +325,123 @@ func coldStartPreference(pod cache.Pod, container cache.Container) (ColdStartPre
...
-
-	if !preferShared {
-		switch {
-		case full == 1:
-			if explicit {
-				isolate = preferIsol
-			} else {
-				isolate = true
+	// sub-core CPU request
+	case cores == 0 && fraction < 1000:
```
askervin, line 398: `fraction < 1000` is always true after line 392 (`fraction = fraction % 1000`)

pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
```diff
@@ -217,62 +325,123 @@ func coldStartPreference(pod cache.Pod, container cache.Container) (ColdStartPre
...
+		// 1 <= CPU request < 2
+	case cores < 2:
+		// fractional allocation, potentially mixed
+		if fraction > 0 {
+			if preferShared {
+				return 0, 1000*cores + fraction, false, cpuNormal
 			}
-		case full > 1:
-			isolate = preferIsol && explicit
+			return cores, fraction, preferIsolated, cpuNormal
```
askervin, line 407: This may be just because of lacking some background information, but I find it somewhat difficult to understand/explain/document why scheduling on isolated cores requires explicit annotation when requesting 2100m CPU but not when requesting 1900m CPU.

Do you think we could simplify the behavior by using `preferIsolated && explicitIsolated` in every case, or am I missing some important use case?

askervin (commented):
Looks good already.

Added a nitpick and a question on isolated cores to understand why the logic is like this...

pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
marquiz: @klihub can correct me if I'm wrong, but, I guess the thinking is that allocating exactly one isolated core is "safe" from the workload scheduling point of view. When/if allocating more than one isolated cores the workload must be "isolcpus aware" and handle scheduling/cpu affinity of its threads by itself. In the case of one core this is not a problem as all threads will be forced on this one core

askervin:
> Would it provide better usability to use a single explicit non-boolean annotation
(for instance cpu-preferences) to express all CPU allocation preferences ?

Elaborating this a bit... At minimum current annotations would be replaced by:

`cpu-preferences.cri-resource-manager.intel.com/pod: shared`
`cpu-preferences.cri-resource-manager.intel.com/pod: isolated`

...and if we'd like to enable running a `kube-system` pod on exclusive CPUs (instead of running it on always-shared reserved CPUs, this might be useful), or a non-kube-system pod on reserved CPUs, I could imagine these to make sense as well:

`cpu-preferences.cri-resource-manager.intel.com/pod: normal`
`cpu-preferences.cri-resource-manager.intel.com/pod: reserved`

Some combinations would be valid and useful:

`cpu-preferences.cri-resource-manager.intel.com/pod: normal,shared`
`cpu-preferences.cri-resource-manager.intel.com/pod: normal,isolated`

Some combinations would not be supported:

`cpu-preferences.cri-resource-manager.intel.com/pod: reserved,isolated`

Some combinations would be contradictory:

`cpu-preferences.cri-resource-manager.intel.com/pod: shared,isolated`
`cpu-preferences.cri-resource-manager.intel.com/pod: normal,reserved`

Did you have something like this in mind? (I chose these words and combinations directly based on the implementation, which doesn't guarantee perfect user experience...)

pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
klihub: Yes, it's exactly as @marquiz said. And that's why we don't assume all workloads to be safe for multi-core mixed allocations but require them to be annotated so.

marquiz: K. Agree with @askervin that this should be documented. A comment in the source code would probably be in order

klihub:
> > Would it provide better usability to use a single explicit non-boolean annotation
> > (for instance cpu-preferences) to express all CPU allocation preferences ?
> 
> Elaborating this a bit... At minimum current annotations would be replaced by:
> 
...
> Did you have something like this in mind? (I chose these words and combinations directly based on the implementation, which doesn't guarantee perfect user experience...)

Yes, something like that. A set of allocation preferences with further possible qualifiers. Eg. `{ shared, exclusive, mixed }`, and `{isolated, non-isolated}`, although I'd probably throw in EPP/SS-related ones as well to allow explicitly expressed opt-in/opt-out preferences for CPUs in different performance profile classes.

There are a few clear benefits of using a single annotation. For instance, it is obvious that you're tweaking inter-related things, and in many cases it could be more intuitive/descriptive what you want/mean and easier to understand than with a set of implicitly inter-related booleans. But as you said, it is not completely without problems.

pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
klihub: Thanks, true. Removed tautological comparison.

pkg/cri/resource-manager/policy/builtin/memtier/pod-preferences.go
klihub: I added a rationale section to the comments explaining the reasons behind the choices when to allocate
isolated vs. ordinary cores in the various cases.
Fixed a comparison thinko in the comments and a typo.
Removed a useless ` + 1000 * fraction` from a `return` on a branch where `fraction` was guaranteed to
always be `0`.

askervin (approved):
Nice comment improvement!

LGTM, works great on top of current master.

kad:
my line of thoughts: those so far are "preferred". what about "required" ?
Also, if we change some treatments of annotations - that should be reflected in the documentation as well.

klihub:
> my line of thoughts: those so far are "preferred". what about "required" ?

Wouldn't the easiest/most natural way to handle required vs. preferred CPU allocation preferences be the above outlined single, combined annotation, and qualifiers to denote the preferred vs. mandatory nature of the request ?

> Also, if we change some treatments of annotations - that should be reflected in the documentation as well.

The exact interpretation of these annotations was never properly documented in detail, and I agree that we
need to document it.

kad:
> > Also, if we change some treatments of annotations - that should be reflected in the documentation as well.
> 
> The exact interpretation of these annotations was never properly documented in detail, and I agree that we
> need to document it.

My comment was more about general practice: if we change some behaviour, and while documenting it in code comments is ok, we need to change userfacing doc in the same PR, so it wouldn't be lost.

klihub:
> My comment was more about general practice: if we change some behaviour, and while documenting it in code comments is ok, we need to change userfacing doc in the same PR, so it wouldn't be lost.

Ah, ok. AFAICT, this PR does not change documented behavior. Partly this is because the behavior was
never properly documented in the first place. But in reality, mostly because it was not the intention. This
PR is about spelling out saner behavior for some less used/obvious edge cases, and in a more explicit
manner.

Anyway, I wrote an initial version of a more proper memtier description. It's now at the tip of this PR. I'm
fine with including it as part of this PR. IMO, ideally it should be a separate PR. @kad, @marquiz,
@askervin, let me know guys what is the consensus preference and I'll adjust my PR accordingly.

marquiz (commented):
I've really only looked at the documentation which is a huge improvement over the previous. Really good job 👍  One comment there is that I'd add some references (links) to some terms like "PMEM", "HBM", "2-layer mode", "isolcpus" for example. WDYT?

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+  - prefer idle pools over busy ones, IOW more remaining free capacity and
+    fewer workloads
+  - prefer pools with better overall device alignment
+
+## Features
+
+The memtier policy has the following features:
+
+  - topologically aligned allocation of CPU and memory
+    * assign CPU and memory to a workloads with tightest available alignment
```
askervin, line 94: "a workloads" -> workloads

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+  - aligned allocation of devices
+    * pick pool for workload based on localoty of devices already assigned
```
askervin, line 96: "localoty" -> "locality"

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+  - shared allocation of CPU cores
+    * assign workload to shared subset of pool CPUs
+  - exclusive allocation of CPU cores
+    * slice off CPU cores from shared subset and assign to workload
```
askervin, line 100: "slice off"
->
"dynamically slice off"

This definitely is a matter of taste... but I'd emphasize that the size of the set of shared CPUs in a pool indeed changes during runtime. This is pretty cool feature.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+  - mixed allocation of CPU cores
+    * assign both exclusive and shared CPU cores to workload
+  - discover and use kernel-isolated CPU cores
```
askervin, line 103: "discover and use" -> "discovering and using"

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+
+The policy uses a combination of the QoS class and the resource requirements of
+the container to decide if any of these extra allocation preferences should be
+applied. Containers are divided into five groups, with each group having a
+slightly different set of criteria for eligibility.
+
+  - `kube-system` group
+    * all containers in the `kube-system` namespace
+  - `low-priority` group
+    * containers in the `BestEffort or Burstable` QoS class
```
askervin, line 177: `BestEffort or Burstable`
->
 `BestEffort` or `Burstable`

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+  - `sub-core`
+    * not eligible for extra optimizations
+    * always run on shared CPU cores
+  - `mixed`
+    * by default eligible for exclusive and isolated allocation
+    * not eligible for either if `PreferSharedCPUs` is set to true
+    * not eligible for either if annotated to opt out from exclusive allocation
+    * not eligible for isolated allocation if annotated to opt out
+  - `multi-core`
+    * CPU request fractional (`(CPU request % 1000 milli-CPU) != CPU request`):
```
askervin, line 204: `(CPU request % 1000 milli-CPU) != CPU request`
->
`(CPU request % 1000 milli-CPU) != 0`
or
`int(CPU request / 1000 milli-CPU) != CPU request`
or
(`CPU request` is not divisible by 1000 milli-CPU)

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+      - by default not eligible for isolated allocation
+      - not eligible for exclusive allocation if annotated to opt out
+      - eligible for isolated allocation if annotated to opt in
+
+Eligibility for kube-reserved CPU core allocation should always be possible to
+honor. If this is not the case, it is probably due to an incorrect configuration
+which underdeclares `ReservedResources`. In that case, ordinary shared CPU cores
+will be used instead of kube-reserved ones.
+
+Eligibility for exclusive CPU allocation should always be possible to honor. If
```
askervin, line 218: "Eligibility for exclusive CPU allocation should always be possible to honor. ..."

I wouldn't bother to mention the possibility of a bug in the algorithm. There shouldn't be one, and on the other hand, if there is, it could break any other claim as well, not just this.

However, a configuration error like too large ReservedResources cpuset could cause problems. That takes CPUs from real pools without kube scheduler knowing about it.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+    topologyhints.cri-resource-manager.intel.com/container.$C1: "false"
+    # disregard hints for all containers by default
+    topologyhints.cri-resource-manager.intel.com/pod: "false"
+    # but take hints into account for container $C2
+    topologyhints.cri-resource-manager.intel.com/container.$C2: "true"
 ```
 
-The `memtier` policy will then aim to allocate resources from a topology
-node which can satisfy the memory requirements.
+While this might change in the future, currently implicit topology hint
```
askervin, line 306: "While this might change in the future, currently implicit topology hint..."
->
"Topology hint..."

(I'd prefer documenting only the current behavior and avoid speculating on what could change... unless we already know for sure something is going to change soon.)

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+
+### Shared, Exclusive, and Isolated CPU Preference
+
+A container can opt in to or opt out from shared CPU allocation using the
+following annotation.
+
+```yaml
+metadata:
+  annotations:
+    # opt in container $C1 to shared CPU core allocation
```
askervin, line 250: In existing examples we don't use `$` as part of container names. Perhaps we could go here with `$`, too? (At least it made me wonder if one needs to prefix container names by `$` when using this notation.)

askervin:
Great stuff! Didn't notice any big gap, just tiny fixes.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+in 2-layer mode with all types separately visible, each zone of special type
+of memory is assigned to the closest NUMA node pool.
+
+Each non-leaf pool node in the tree is assigned the union of the resources of
+its children. So in practice, dies end up containing the closest CPU cores
+and the memory zones in the die, sockets end up containing the CPU cores and
+memory zones in the socket's dies, and the root ends up containing all CPU
+cores and memory zones in all sockets. The actual pool tree is built with all
+non-branching paths eliminated, by omitting any child node which would have
+ended up being the lone single child of its parent. This is merely an minor
```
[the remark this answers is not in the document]
klihub: Fixed.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+## Background
+
+On server-grade hardware the CPU cores, I/O devices and other peripherals
+form a rather complex network together with the memory controllers, the
+I/O bus hierarchy and the CPU interconnect. When a combination of these
+resources are allocated to a single workload, the performance of that
+workload can vary greatly, depending on how efficiently data is transferred
+between them or, in other words, on how well the resources are aligned.
+
+There are a number of inherent architectural properties of server grade
```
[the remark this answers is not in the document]
klihub: Fixed.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+
+With this setup, each pool in the tree has a topologically aligned set of CPU
+and memory resources. The amount of available resources gradually increases in
+the tree from bottom to top, while the strictness of alignment is gradually
+relaxed. In other words, as one moves from bottom to top in the tree, it is
+getting gradually easier to fit in a workload, but the price paid for this is
+a gradually increasing maximum potential cost or penalty for memory access and
+data transfer between CPU cores.
+
+Another property of this setup is that the resources sets of sibling pools at
```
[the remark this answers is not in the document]
klihub: Fixed.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+    * the policy knows about three kinds of memory:
+      - DRAM is regular system main memory
+      - PMEM (persistent memory) is large-capacity memory, such as Intel® Optane™
+        memory
+      - HBM (high-bandwidth memory) is high speed memory, typically found on
+        some special-purpose computing systems
+  - cold start
+    * pin workload exclusively to PMEM for an initial warm-up period
+  - dynamic page demotion
+    * forcibly migrate read-only and idle container memory pages to PMEM
```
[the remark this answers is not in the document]
klihub: It's there: "- discover and use kernel-isolated CPU cores"...

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+misalignment and workload performance degradation. There are a multitude of
+CPU cores available to run workloads. There are a multitude of memory
+controllers these workloads can use to store and retrieve data from main
+memory. There are a multitude of I/O devices attached to a number of I/O
+buses the same workloads can access. The CPU cores can be divided into a
+number of groups, with each group having different access latency and
+bandwidth to each memory controller and I/O device.
+
+If a workload is not assigned to run with a properly aligned set of CPU,
+memory, and devices, it will not be able to achieve optimal performance.
```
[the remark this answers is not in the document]
klihub: Fixed.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+
+Containers can be annotated to diverge from the default CPU allocation
+preferences the policy would otherwise apply to them. These annotations
+can be given both with per pod and per container resolution. If for any
+container both of these exist, the container-specific one takes precedence.
+
+### Shared, Exclusive, and Isolated CPU Preference
+
+A container can opt in to or opt out from shared CPU allocation using the
+following annotation.
```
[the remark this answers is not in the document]
klihub: I replaced all references to 'annotations' immediately preceding examples with 'Pod annotations'.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+groups, are all pinned to run on the shared subset of CPU cores in the
+container's assigned pool. This shared subset can and usually does change
+dynamically as exclusive CPU cores are allocated and released in the pool.
+
+## Container CPU Allocation Preference Annotations
+
+Containers can be annotated to diverge from the default CPU allocation
+preferences the policy would otherwise apply to them. These annotations
+can be given both with per pod and per container resolution. If for any
+container both of these exist, the container-specific one takes precedence.
```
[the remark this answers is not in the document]
klihub: I replaced most references to 'annotations' with 'Pod annotations'.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
-```
-cri-resmgr --policy memtier --reserved-resources cpu=750m
+The memtier policy automatically builds a tree of pools based on the detected
+hardware topology. Each pool has a set of CPUs and memory zones assigned as
+their resources. Resource allocation for workloads happens by first picking
+the pool which is considered to fit the best the resource requirements of
+the workload and then assigning CPU and memory from this pool.
+
+The pool nodes at various depths from bottom to top represent the NUMA nodes,
+dies, sockets, and finally the whole of the system a the root node. Leaf NUMA
```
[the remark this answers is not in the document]
klihub: Fixed.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+nodes are assigned the memory behind their controllers / zones and CPU cores
+with the smallest distance / access penalty to this memory. If the machine
+has multiple types of memory, for instance both DRAM and PMEM, and is running
+in 2-layer mode with all types separately visible, each zone of special type
+of memory is assigned to the closest NUMA node pool.
+
+Each non-leaf pool node in the tree is assigned the union of the resources of
+its children. So in practice, dies end up containing the closest CPU cores
```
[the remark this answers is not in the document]
klihub: Reworded and eliminated 'closest' from the text.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
+and the memory zones in the die, sockets end up containing the CPU cores and
+memory zones in the socket's dies, and the root ends up containing all CPU
+cores and memory zones in all sockets. The actual pool tree is built with all
+non-branching paths eliminated, by omitting any child node which would have
+ended up being the lone single child of its parent. This is merely an minor
+optimisation in the implementation and it has no practical significance for
+this discussion.
```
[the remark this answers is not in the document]
klihub: Removed the last few irrelevant sentences that state the slight divergence of the implementation from the preceding description due to a small practical optimization.

docs/policy/memtier.md
klihub: Fixed.

docs/policy/memtier.md
klihub: Fixed.

docs/policy/memtier.md
klihub: That is mentioned/explained elsewhere in the text: "This shared subset can and usually does change
dynamically as exclusive CPU cores are allocated and released in the pool." Nevertheless, I updated feature list entry as suggested.

docs/policy/memtier.md
klihub: Updated.

docs/policy/memtier.md
klihub: Updated.

docs/policy/memtier.md
klihub: Fixed.

docs/policy/memtier.md
klihub: Removed reference to possible bugs in resource accounting.

docs/policy/memtier.md
```diff
@@ -1,70 +1,314 @@
...
-types, you could use pod metadata such as this:
+A container can opt in to or opt out from isolated exclusive CPU core
+allocation using the following annotation.
 
-```
+```yaml
 metadata:
   annotations:
-    memory-type.cri-resource-manager.intel.com/container.container1: dram,pmem
+    # opt in container $C1 to isolted exclusive CPU core allocation
```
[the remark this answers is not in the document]
klihub: Fixed.

docs/policy/memtier.md
klihub: Removed.

docs/policy/memtier.md
klihub: Removed $ from all example container names.

klihub:
> I've really only looked at the documentation which is a huge improvement over the previous. Really good job  One comment there is that I'd add some references (links) to some terms like "PMEM", "HBM", "2-layer mode", "isolcpus" for example. WDYT?

Good idea. I'll try to find appropriate online content to link to (it's easy for isolcpus) and add links.

docs/policy/memtier.md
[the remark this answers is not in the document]
klihub: Added explicit reference to 'isolcpus'.

klihub:
> > I've really only looked at the documentation which is a huge improvement over the previous. Really good job  One comment there is that I'd add some references (links) to some terms like "PMEM", "HBM", "2-layer mode", "isolcpus" for example. WDYT?
> 
> Good idea. I'll try to find appropriate online content to link to (it's easy for isolcpus) and add links.

Okay, added the following external links as explanation:
  - isolcpus: https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html#cpu-lists
  - PMEM: https://www.intel.com/content/www/us/en/products/memory-storage/optane-dc-persistent-memory.html
  - HBM: https://en.wikipedia.org/wiki/High_Bandwidth_Memory

Any suggestions to better docs are welcome.

I also removed any references to '2-layer mode'. It was anyway incorrect as it
should have been 'two level memory' in the first place. However, since
1LM, 2LM, Memory Mode and App Direct Mode are just prone to create
more confusion than clarity, I opted for a short verbal clarification of what
'tiering' is in the scope of 'memtier'.

marquiz (approved):
lgtm

PR #592, opened 2021-01-08
e2e: unified memory-type and cold-start annotation syntax

askervin:
Update memtier E2E tests to use unified pod/container annotation syntax implemented in PR #588.


klihub (commented):
LGTM.

But shouldn't we keep around tests for the old syntax, until we remove support for it altogether ?

askervin:
In order to minimize complexity, ended up duplicating test directories.

When we drop the old syntax support, we can remove `*-deprecated-syntax` test directories and their contents.

In terms of maintenance, it should be enough to _extend_ new syntax tests only. And I can only hope we don't need to _update_ these tests before dropping deprecated syntax...

klihub:
> In order to minimize complexity, ended up duplicating test directories.
> 
> When we drop the old syntax support, we can remove `*-deprecated-syntax` test directories and their contents.

That's a good idea, helps maintenance.

klihub (approved):
LGTM.

PR #595, opened 2021-01-14
docs: recognize/handle .md-links to element IDs.

klihub:
Properly recognize a `[]()`-link to an `.md`-file with a `#`-prefixed element ID.
This will cause the generation of a proper link to the corresponding final `HTML`
element in the generated output, instead of a link to the base `.md`-file itself.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -143,7 +143,7 @@ def isHTTPLink(uri):
     return uri.startswith('http://') or uri.startswith('https://')
 
 def isMDFileLink(uri):
-    return uri.endswith('.md')
+    return uri.endswith('.md') or '.md#' in uri
 
 def isRSTFileLink(uri):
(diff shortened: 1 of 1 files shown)


kad (approved):
lgtm

docs/conf.py
```diff
@@ -143,7 +143,7 @@ def isHTTPLink(uri):
     return uri.startswith('http://') or uri.startswith('https://')
 
 def isMDFileLink(uri):
-    return uri.endswith('.md')
+    return uri.endswith('.md') or uri.find('.md#') != -1
```
kad, line 146: However, just a bit more pythonic way: `if '.md#' in uri`

klihub: Tnaks, I'll update.

kad (approved):

PR #596, opened 2021-01-15
server,resource-manager: flush logs after every request/event processed.

klihub:
Explicitly flush logs after every CRI request and resource-manager event processed.

### DIFF
--- a/pkg/cri/resource-manager/events.go
+++ b/pkg/cri/resource-manager/events.go
@@ -78,6 +78,7 @@ func (m *resmgr) startEventProcessing() error {
 					evtlog.Error("rebalancing failed: %v", err)
 				}
 			}
+			logger.Flush()
 		}
 	}()
 
--- a/pkg/cri/server/server.go
+++ b/pkg/cri/server/server.go
@@ -342,6 +342,7 @@ func (s *server) intercept(ctx context.Context, req interface{},
 	}
 
 	s.collectStatistics(kind, name, start, send, recv, end)
+	logger.Flush()
 
 	return rpl, err
 }


pkg/cri/resource-manager/events.go
```diff
@@ -79,6 +79,7 @@ func (m *resmgr) startEventProcessing() error {
 				}
 			}
 		}
+		logger.Flush()
```
marquiz, line 82: Should this be inside the for loop(?)

klihub: Of course. Argh.

pkg/cri/server/server.go
```diff
@@ -342,6 +342,7 @@ func (s *server) intercept(ctx context.Context, req interface{},
 	}
 
 	s.collectStatistics(kind, name, start, send, recv, end)
+	logger.Flush()
```
marquiz, line 345: Hmm, should be `defer logger.Flush()` instead? Would flush in case of errors etc.

klihub: That's what I head initially. But because the whole function is obviously deliberately written so that it never return anywhere else than in the end, I changed it to the current form.

klihub: But maybe it would be less clobbersome if we move the log flush to the end of server.interceptRequest() (the caller of inercept), or to the beginning of that with a deferred call. WDYT?

marquiz: > the whole function is obviously deliberately written so that it never returns anywhere else than in the end

Fair enough, makes sense. Just a smartass note that this might be changed by a future patch 😉 



> But maybe it would be less clobbersome if we move the log flush to the end of server.interceptRequest() (the caller of inercept), or to the beginning of that with a deferred call.

Dunno, I'm really undecided on that. I think the current location is good but don't object if you want to move it

marquiz (approved):
LGTM

askervin (approved):
LGTM

PR #597, opened 2021-01-20
resource-manager: rename 'memtier' policy to 'topology-aware'.

klihub:
This PR replaces the original `topology-aware` policy implementation with its forked
and updated `memtier` version. In particular the PR
 - removes the original `topology-aware` policy
 - renames the `memtier` policy to `topology-aware`
 - registers `memtier` as a policy and configuration alias for `topology-aware`
 - replaces all references to `memtier` with `topology-aware` in the docs
 - replaces all references to `memtier` with `topology-aware` in tests and benchmarks


kad:
After looking at this, maybe it would be more logical to do opposite way -> move all memtier as "topology-aware", and then do memtier as alias to new "topology-aware" ?
In the end of the day, the memtier initially supposed to be only development variant of topology-aware?

klihub:
> After looking at this, maybe it would be more logical to do opposite way -> move all memtier as "topology-aware", and then do memtier as alias to new "topology-aware" ?
> In the end of the day, the memtier initially supposed to be only development variant of topology-aware?

Fair enough. Reworked the PR with that in mind.

askervin:
Should we drop `docs/policy-memtier.md` at this point, too? It has a broken link to `/pkg/cri/resource-manager/policy/builtin/memtier/README.md`.

This tiny comment might be less confusing if changed to topology-aware:
`pkg/cri/resource-manager/control/page-migrate/page-migrate.go:  demoter    *demoter              // demoter adopted from memtier policy`

How would you prefer to deal with `test/e2e/benchmarks.test-suite/memtier_benchmark` with a few points to memtier. Handle in this patch or later?

Otherwise looks good!

klihub:
> Should we drop `docs/policy-memtier.md` at this point, too? It has a broken link to `/pkg/cri/resource-manager/policy/builtin/memtier/README.md`.
> 
> This tiny comment might be less confusing if changed to topology-aware:
> `pkg/cri/resource-manager/control/page-migrate/page-migrate.go: demoter *demoter // demoter adopted from memtier policy`

Yes, I was not sure about this... but you're right it should refer to the renamed policy, so memtier.

> 
> How would you prefer to deal with `test/e2e/benchmarks.test-suite/memtier_benchmark` with a few points to memtier. Handle in this patch or later?

Neither. Handle it by leaving it completely alone and untouched. AFAICT/IMO, that `memtier` comes
from is the name of the `redis`/`memcached` benchmark test client, `memtier_benchmark`. Not really
related to/affiliated with our former `memtier` policy in any other way than us using it for testing.

klihub:
> > How would you prefer to deal with `test/e2e/benchmarks.test-suite/memtier_benchmark` with a few points to memtier. Handle in this patch or later?
> 
> Neither. Handle it by leaving it completely alone and untouched. AFAICT/IMO, that `memtier` comes
> from the name of the `redis`/`memcached` benchmark test client, `memtier_benchmark`. Not really
> related to/affiliated with our former `memtier` policy in any other way than us using it for testing.

But TBH, you should know better because you wrote those bits. So if you disagree or think I overlooked
there something which does refer to the renamed memtier policy, please let me know and I'll update the
PR accordingly.

askervin:
> Neither. Handle it by leaving it completely alone and untouched. AFAICT/IMO, that `memtier` comes
> from the name of the `redis`/`memcached` benchmark test client, `memtier_benchmark`.

Definitely so. I already forgot this name collision, thanks for reminding me! =D

...and you already had updated the essential line there in the `cri-resmgr.cfg`.

With your most recent change this looks good to me!

askervin (dismissed):
LGTM

pkg/cri/resource-manager/policy/builtin/topology-aware/topology-aware-policy.go
```diff
@@ -280,23 +286,203 @@ func (p *policy) Introspect(state *introspect.State) {
...
+
+	p.root.Dump("<post-realloc>")
+
+	return nil
+}
+
 func (p *policy) configNotify(event config.Event, source config.Source) error {
-	log.Info("configuration %s:", event)
+	policyName := PolicyName
+	if !p.isAlias {
```
klihub, line 377: This test is the wrong way around...

askervin (approved):
LGTM

marquiz (approved):
Looks good 👍
LGTM

PR #598, opened 2021-01-22
podpools: policy for pod-granularity workload placement

askervin:


pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,694 @@
...
+func (p *podpools) validatePodCPU(pod cache.Pod, pool *Pool) {
+	// Log pod configuration error if a pool had fixed amount of
+	// CPUs per, pod but the pod failed to request the correct
+	// amount.
+	podID := pod.GetID()
+	if podmCPU, ok := p.podMaxMilliCPU[podID]; ok {
+		if pool.Def.MaxPods > 0 {
+			poolmCPUperPod := int64(pool.CPUs.Size() * 1000 / pool.Def.MaxPods)
+			mCPUerr := podmCPU - poolmCPUperPod
+			if mCPUerr < -podMilliCPUErrorMargin || mCPUerr > podMilliCPUErrorMargin {
```
klihub, line 353: Is the idea here that somehow slight over- and undersubscriptions will even/cancel each other out, or why do we allow both ?

sample-configs/podpools-policy.cfg
```diff
@@ -0,0 +1,101 @@
...
+    # In this example we use 14 CPUs, excluding CPUs #0 and #1
+    # (hyperthreads of core 0).
+    CPU: cpuset:2-15
+
+  # ReservedResources specifies CPU(s) that the policy dedicates for
+  # running kube-system pods.
+  ReservedResources:
+    # Here we dedicate CPU #15 for these pods.
+    # This leaves 13 out of 14 available CPUs unallocated.
+    CPU: cpuset:15
```
klihub, line 28: Maybe we could mention here what are the correct corresponding system- and kube-(CPU-)reservation flags to be passed to `kubelet`.

klihub (commented):
Overall looks very good to me. I spotted a few typos, and have some suggested additions to the comments. And a question about the reason for allowing both over- and undersubscription.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,694 @@
...
+	// Check overbooking
+	if cpuAvail := p.availableMilliCPUs(pool); cpuAvail < 0 {
+		log.Error("overbooked pool %q, cpuset:%s: %dm / %dm CPUs used, %d mCPU available", pool.PrettyName(), pool.CPUs, pool.CPUs.Size()*1000-int(cpuAvail), pool.CPUs.Size()*1000, cpuAvail)
+	}
+}
+
+// validatePodCPU compares max CPU requests against pool CPU capacity per pod.
+func (p *podpools) validatePodCPU(pod cache.Pod, pool *Pool) {
+	// Log pod configuration error if a pool had fixed amount of
+	// CPUs per, pod but the pod failed to request the correct
```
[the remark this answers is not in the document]
askervin: Fixed.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,694 @@
...
+	case is == "":
+		instances = 1
+	case strings.HasSuffix(is, "%"):
+		tis := strings.TrimSpace(strings.TrimSuffix(is, "%"))
+		i64, err := strconv.ParseInt(tis, 0, 32)
+		if err != nil || i64 < 0 {
+			return 0, 0, podpoolsError("invalid Instances: %q", is)
+		}
+		instances = freeCpus * int(i64) / 100 / cpusPerPool
+	case strings.HasSuffix(strings.ToLower(is), "cpus"):
```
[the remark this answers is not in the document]
askervin: Fixed, allowing also "N cpu" for N > 1. Some unit tests added.

docs/policy/podpools.md
```diff
@@ -0,0 +1,97 @@
...
+The policy is configured using the yaml-based configuration system of CRI-RM.
+See [setup and usage](../setup.md#setting-up-cri-resource-manager) for more
+details on managing the configuration.
+
+At minimum, you need to specify the active policy in the
+configuration, and define at least one pod pool. For example, the
+following configuration dedicates 95 % of non-reserved CPUs on the
+node to be used by `dualcpu` pools. Every pool instance (`dualcpu[0]`,
+`dualcpu[1]`, ...) contains two exclusive CPUs and has a capacity
+(`MaxPods`) of one pod. The CPUs are used by only by containers of pods
```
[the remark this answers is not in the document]
askervin: Fixed.

pkg/cri/resource-manager/policy/builtin/podpools/flags.go
```diff
@@ -0,0 +1,121 @@
...
+	return fmt.Sprintf("#UNNAMED-FILLORDER(%d)", int(fo))
+}
+
+// MarshalJSON marshals a FillOrder as a quoted json string
+func (fo FillOrder) MarshalJSON() ([]byte, error) {
+	buffer := bytes.NewBufferString(fmt.Sprintf("%q", fo))
+	return buffer.Bytes(), nil
+}
+
+// UnmarshalJSON unmashals a FillOrder quoted json string to the enum value
```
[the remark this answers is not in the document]
askervin: Fixed.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,694 @@
...
+	// PodIDs maps pod ID to list of container IDs.
+	// - len(PodIDs) is the number of pods in the pool.
+	// - len(PodIDs[podID]) is the number of containers of podID
+	//   currently assigned to the pool.
+	// - Def.MaxPods - len(PodIDs) is free pod capacity.
+	PodIDs map[string][]string
+}
+
+var log logger.Logger = logger.NewLogger("policy")
+var _ policy.Backend = &podpools{}
```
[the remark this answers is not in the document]
askervin: Fixed here. Build errors appear without it, too (`podpools does not implement policy.Backend (missing Description method)`).

Didn't want to touch other policies in this patch.  Let's nuke these in a separate cleanup commit:

```
none/none-policy.go:var _ policy.Backend = &none{}
--
static/static-policy.go:var _ policy.Backend = &static{}
--
static-plus/static-plus-policy.go:var _ policy.Backend = &staticplus{}
--
static-pools/stp-policy.go:var _ policy.Backend = &stp{}
--
topology-aware/topology-aware-policy.go:var _ policyapi.Backend = &policy{}
```

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,694 @@
...
+
+// String is a stringer for a pool.
+func (pool Pool) String() string {
+	podCount := len(pool.PodIDs)
+	contCount := 0
+	for _, contIDs := range pool.PodIDs {
+		contCount += len(contIDs)
+	}
+	s := fmt.Sprintf("%s{cpus:%s, mems:%s, pods:%d/%d, containers:%d}",
+		pool.PrettyName(), pool.CPUs, pool.Mems, podCount, pool.Def.MaxPods, contCount)
```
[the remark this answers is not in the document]
askervin: Added dumpPool(), and using it on container allocation/release when policy debugging is enabled.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
askervin: The purpose of this margin is to allow rounding errors in the magnitude of 0.5 mCPU/container up to 20 containers/pod - and allow the error in rounding up and down. (Therefore both over and undersubscription.)

sample-configs/podpools-policy.cfg
askervin: Added.

askervin:
@marquiz , this is now good for review, most @klihub's comments resolved. Two left open (a partial fix and an answer to question rather than a fix).

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
klihub: Could we then spell that out explicitly with a comment ?

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,727 @@
...
+	delete(p.podMaxMilliCPU, podID)
+}
+
+// trackPodCPU keeps track on pod's CPU requests.
+func (p *podpools) trackPodCPU(pod cache.Pod, pool *Pool) {
+	podID := pod.GetID()
+	current := p.getPodMilliCPU(podID)
+	if max, ok := p.podMaxMilliCPU[podID]; ok {
+		if max < current {
+			p.podMaxMilliCPU[podID] = current
```
marquiz, line 360: Hmm, what is the purpose of this `podMaxMilliCPU` map? Wouldn't `getPodMilliCPU()` be enough? AFAIU the resource allocation (requests or limits) may not change or no new containers may be added after the pod object has been created. I'm probably missing something here 🤔

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,727 @@
...
+			}
+		}
+	}
+	return cpuRequested
+}
+
+// configNotify applies new configuration.
+func (p *podpools) configNotify(event pkgcfg.Event, source pkgcfg.Source) error {
+	log.Info("configuration %s", event)
+	if err := p.setConfig(podpoolsOptions); err != nil {
```
marquiz, line 408: Stupid question: what happens to the already running pods? Are they re-allocated somewhere/somehow?

marquiz (commented):
Looks really good! 👍  And the tests are extensive - feeling guilty when looking at some of my own PRs 🙄  Just a few nitpicks, comments and stupid questions.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
klihub: Without an mutating webhook, there is no way to make sure that the resource request for any pool-annotated Pod/continer will be consistent with the configuration of that pool type so that parallel accounting in the scheduler does not go haywire. The next best thing one can do is to detect clearly inconsistent/incorrect configuration/usage and log a diagnostic error message about it.

However, (again without forcing the policy to run with cri-resmgr-webhook or something similar) you cannot detect such misconfiguration when the pod is created. You can tell it once all the containers in the pod have been created. With the amount of stock metadata in a pod or a container, you don't know when that moment in time comes. The easiest thing to do is to record the largest ever concurrent allocation within the pod, then cross-check this with the configuration when the pod is deleted. AFAICT, this is the recording part, and the cross-checking counterpart is `validatePodCPU`.

marquiz: Ah yes, of course, makes perfectly sense. Tend to forget that resource accounting problem...

It might be beneficial for future readers to describe that^ in the docstring/comment of `trackPodCPU()`. WDYT?

klihub: Yes, I think that would be a good idea.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
askervin: Clarifying comments added here and next to declaring the constant `podMilliCPUErrorMargin`.

docs/policy/podpools.md
```diff
@@ -0,0 +1,97 @@
...
+    command:
+      - "sh"
+      - "-c"
+      - "while :; do grep _allowed_list /proc/self/status; sleep 5; done"
+  - name: testcont1
+    image: busybox
+    command:
+      - "sh"
+      - "-c"
+      - "while :; do grep _allowed_list /proc/self/status; sleep 5; done"
```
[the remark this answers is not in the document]
askervin: Definitely. Added with explanations.

pkg/cri/resource-manager/policy/builtin/podpools/flags.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
```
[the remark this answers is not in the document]
askervin: Sedited `s/2020/2020-2021/g`.

pkg/cri/resource-manager/policy/builtin/podpools/flags.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// PoolDef contains a pool definition.
+type PoolDef struct {
+	// Name is the name of the pool, or name prefix of
+	// multi-instance pools.
+	Name string `json:"Name"`
+	// CPU specifies the number of CPUs exclusively usable by
+	// pods in the pool.
+	CPU string `json:"CPU,omitempty"`
```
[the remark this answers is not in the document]
askervin: Thanks, `omitempty` removed.

pkg/cri/resource-manager/policy/builtin/podpools/flags.go
```diff
@@ -0,0 +1,121 @@
...
+	// MaxPods specifies the maximum number of pods assigned to
+	// the pool. 0 (the default) means unlimited. -1 means no
+	// pods.
+	MaxPods int `json:"MaxPods"`
+	// Instances specifies the number of multi-instance pools,
+	// either directly or as CPU (count/percentage) reserved for
+	// instances.
+	Instances string `json:"Instances,omitempty"`
```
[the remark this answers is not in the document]
askervin: Added the default being 1 into the comment and kept `omitempty`.

sample-configs/podpools-policy.cfg
```diff
@@ -0,0 +1,108 @@
+# This example demonstrates pod-based CPU and memory pinning.
+# All containers of a pod run in the same CPU/memory pool.
+# The capacity of a pool is defined as a number of pods it can
+# contain.
+#
+# You can launch a pod to a pod pool by annotating it:
```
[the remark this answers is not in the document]
askervin: Great, added!

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -0,0 +1,727 @@
...
+	// first pool in the list is the preferred one.
+	switch poolDef.FillOrder {
+	case FillBalanced:
+		sort.Slice(pools, func(i, j int) bool {
+			return len(pools[i].PodIDs) < len(pools[j].PodIDs)
+		})
+	case FillPacked:
+		sort.Slice(pools, func(i, j int) bool {
+			return len(pools[i].PodIDs) > len(pools[j].PodIDs)
+		})
```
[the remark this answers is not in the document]
askervin: That's very correct. I considered logging a warning or an error message, but finally ended up rejecting the pool configuration with "_n_ pool(s) unreachable..." error message. Still feeling a bit uncertain as I tend to avoid writing `if`s that only prevent behavior, even shooting oneself in the foot. Maybe this time it's not just my lack of imagination that I can't come up with a use case for unreachable pools... :)

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
askervin: Explained this in new comment in `trackPodCPU()`.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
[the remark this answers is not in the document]
marquiz: > Still feeling a bit uncertain as I tend to avoid writing `if`s that only prevent behavior, even shooting oneself in the foot

Yeah, that is a good point and a good principle – too "smart" software can be a real pita. I'm counting on your judgement here 😉

klihub (commented):
So what about reconfiguration, the point @marquiz raised ?

We can leave this as a future improvement but then IMO, the
policy should always reject all reconfiguration attempts, either
unconditionally or if it has any current pod assignments in any
of the pools.

AFAICT, the current code does not handle it properly. It tries to
recreate the pools according to the new configuration but loses
all container assignments regardless whether this fails or not.
In the failure case, the configuration will be rolled back by the
infra, but that will still leave the policy with an empty container
set.

pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
askervin: @klihub, working on this...

askervin: Reconfiguration implemented.

askervin:
> So what about reconfiguration, the point @marquiz raised ?

Ended up implementing reconfiguration as follows:
- If new configuration is valid and pool constructions succeeds, the configuration is accepted. Otherwise it is rejected.
- Running containers are added to new pools. This is now done without trying to be smart about the memory node the container was previously using. (Possible future work, but probably not too high priority.)

For a bit cleaner configuration rejection, I changed `applyPoolDef()` not to modify the podpools object but only update a list of allocated pools. Thus `setConfig()` really sets the config only if there were no errors in the configuration. Before this change it left the object to an undefined state in case of an error (which was not very nice thing to do).

I also noticed that there was `Sync()` call missing in `Start()`. Running containers did not get assigned to any pools after `cri-resmgr` restart. Added that `Sync()`, too.

Added an E2E test for dynamic reconfiguration through the agent. At the same time this is the first E2E test where we use and update cluster-managed cri-resmgr configuration (ConfigMap). Worked fine. :)

This patch is ready for re-review/merging now.

marquiz:
> * If new configuration is valid and pool constructions succeeds, the configuration is accepted. Otherwise it is rejected.
> * Running containers are added to new pools. This is now done without trying to be smart about the memory node the container was previously using. (Possible future work, but probably not too high priority.)

What should we do if
- existing containers do not fit into the set of new pools (i.e. number of pool instances in the new config is too small)? 
- a pool that containers were assigned is removed/renamed?

Should we at least print a warning in those cases, WDYT?

askervin:
> Should we at least print a warning in those cases, WDYT?

Thanks for pointing this out! Currently there are warnings, like

```
pod "pod2" pool "singlecpu" does not match any pool definition, falling back to "default"
cannot find free "dualcpu" pool for pod "pod3"
```

but these situations are not handled in consistently with other issues that arise from resource configuration errors, for instance the CPU overbooking error.

I will change these warnings into log error messages and fix the latter so that it will use the default pool as a fallback, too.

Current error policy will be "relaxed" for pod/pool resource misconfiguration: log errors and use the default pool as a fallback. We can add configuration option for "strict" error policy later on, if necessary.

marquiz:
> I will change these warnings into log error messages and fix the latter so that it will use the default pool as a fallback, too.
> 
> Current error policy will be "relaxed" for pod/pool resource misconfiguration: log errors and use the default pool as a fallback. We can add configuration option for "strict" error policy later on, if necessary.

Good, I think this will be good enough for now. And, as you pointed out, we can improve later

klihub (approved):
I think if all those comments are now addressed, then this LGTM.

marquiz (changes requested):
> I think if all those comments are now addressed, then this LGTM.

I think this is still unaddressed:
*I will change these warnings into log error messages and fix the latter so that it will use the default pool as a fallback, too.*

askervin:
@marquiz, changed warnings to errors, and fixed a skipped pod placement in case of insufficient pool capacity by placing the pod to the default pool. Added E2E test to make sure this works as intended.

marquiz (approved):
I would be ready to merge this now 👍 

Really good work @askervin ! The code was well structured and commented and generally really pleasant to review. And extensive testing ❤️ 😄

klihub (approved):

PR #601, opened 2021-01-27
topology-aware: improved topology tree/node dump

klihub:
When dumping the tree of pools,
  - dump grants and association types with all associated nodes (cpu+mem, cpuonly, memonly)
  - omit dumping empty node memory sets
  - omit dumping empty child lists


kad (approved):
lgtm

marquiz (approved):
lgtm

Issue #191, opened 2020-03-06
cpuallocator: improve allocation heuristics


marquiz:
Bring more NUMA-awareness to the cpuallocator (implemented in `pkg/cpuallocator/`). Discussing/reviewing the cpu allocation logic with @klihub we realized that the allocator is too simple, resulting in clearly non-optimal results. This concerns especially takeIdleCores() which *should* (we think) try to more aggressively and intelligently pack workloads onto a single numa-node.

klihub:
It should also do [these things](https://github.com/intel/cri-resource-manager/pull/602#pullrequestreview-577805992), too.
```
It would also need to be improved with additional tightest-fit allocation rules
beyond the current topology socket/core/thread hierarchy to get to a more
realistic socket/die/NUMA node/core/thread hierachy:
  - try allocating a full die if the number of requested cpus matches exactly
  - try allocating a full NUMA node if the number of cpus matches exactly
  - only then give up and try allocating mere full cores or threads, and also with these
    - try taking sub-NUMA node number of cores/threads from a single NUMA node,
    - try taking sub-die number of cores/threads from a single die
```

marquiz:
> It should also do [these things](https://github.com/intel/cri-resource-manager/pull/602#pullrequestreview-577805992), too

I added these to the issue description

marquiz:
Had a brainstorming session with @klihub . I think we have an agreement that a full rewrite would be the right thing to do. Some ideas we came up with:
- use a tree structure to present the cpu hierarchy, somewhat similar to what the topology-aware policy has
- make it possible to override functionality (function overrides)
- configurability in the default implementation(s), too,

PR #603, opened 2021-01-28
cpuallocator: implement package prioritization

marquiz:
Try allocate all cpus from a single package, if possible. For each
allocation request calculate priority order between cpu packages, based
on requested number of cpus and the cpu priority preference.

### DIFF
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -43,14 +43,15 @@ const (
 
 // allocatorHelper encapsulates state for allocating CPUs.
 type allocatorHelper struct {
-	logger.Logger               // allocatorHelper logger instance
-	sys           sysfs.System  // sysfs CPU and topology information
-	topology      topologyCache // cached topology information
-	flags         AllocFlag     // allocation preferences
-	from          cpuset.CPUSet // set of CPUs to allocate from
-	prefer        CPUPriority   // CPU priority to prefer
-	cnt           int           // number of CPUs to allocate
-	result        cpuset.CPUSet // set of CPUs allocated
+	logger.Logger                    // allocatorHelper logger instance
+	sys             sysfs.System     // sysfs CPU and topology information
+	topology        topologyCache    // cached topology information
+	flags           AllocFlag        // allocation preferences
+	from            cpuset.CPUSet    // set of CPUs to allocate from
+	prefer          CPUPriority      // CPU priority to prefer
+	packagePriority map[sysfs.ID]int // CPU package priority order (lower is higher priority)
+	cnt             int              // number of CPUs to allocate
+	result          cpuset.CPUSet    // set of CPUs allocated
 
 	pkgs []sysfs.CPUPackage // physical CPU packages, sorted by preference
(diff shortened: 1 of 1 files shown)


kad:
@marquiz Package is not enough. we need to check whole hierarchy: cluster->die->package

marquiz:
> @marquiz Package is not enough. we need to check whole hierarchy: cluster->die->package

I know but I would leave that as a future improvement. We can discuss that in #191.

This PR is supposed to fix the greatest shortcoming in the current cpuallocator.

askervin:
This change breaks a couple of e2e tests. I'm verifying if the new allocation order makes sense in these cases. For instance
`policies.test-suite/podpools/n4c16/test01-basic-placement`

(Thanks for very good debug prints. ;) )

pkg/cpuallocator/allocator.go
```diff
@@ -343,6 +354,97 @@ func (a *allocatorHelper) allocate() cpuset.CPUSet {
...
+	pkgs := a.sys.PackageIDs()
+	offline := a.sys.Offlined()
+
+	pkgAvail := make(map[sysfs.ID]int, len(pkgs))
+	pkgPrioAvail := make(map[sysfs.ID][NumCPUPriorities]int, len(pkgs))
+	for _, pkg := range pkgs {
+		avail := a.topology.pkg[pkg].Difference(offline).Intersection(a.from)
+		prioAvail := [NumCPUPriorities]int{}
+		for prio := PriorityHigh; prio < NumCPUPriorities; prio++ {
+			prioAvail[prio] = avail.Intersection(a.topology.cpuPriorities[prio]).Size()
```
askervin, line 369: On e2e test VM (topology n4c16), I'm getting `a.topology.cpuPriorities[prio] == [empty, empty, empty]`. Therefore `iTotal` and `jTotal` are always zero. Even in the "stealing" case. As a result, this function ends up sorting packages based on the number of available cores (line 434).

Yet this still results in very nice CPU allocation for instance in the podpools e2e tests, I'd like to make sure that happens also in the case when not ending up (unintentionally?) to this fallback branch.

askervin:
@marquiz, suggesting changes/fixes.

Make sure `a.topology.cpuPriorities` is non-empty. It seems like it would be empty on any platform without this:
```
diff --git a/pkg/cpuallocator/allocator.go b/pkg/cpuallocator/allocator.go
index 4f9b74e0..daf71b4a 100644
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -527,6 +527,7 @@ func (c *topologyCache) discoverCPUPriorities(sys sysfs.System) {
                        prio[p] = prio[p].Union(cset)
                }
        }
+       c.cpuPriorities = prio
 }
```

If the platform doesn't have cpufreq data, do you think all CPUs could be considered of Normal priority? Like adding following to `discoverSstCPUPriority()` (near `// Finally, determine priority of each CPU`):
```
if len(eppList) == 0 {
        if cpus, ok := epps[sysfs.EPPUnknown]; ok {
                prios[PriorityNormal] = cpus
        }
}
```

(Thanks to @klihub for spotting this!)

PR #606, opened 2021-01-29
Extended detection of Intel Speed Selection Technology (SST)

marquiz:
This PR implements detection of Intel Speed Select Technology (SST) features through the Linux kernel ioctl device interface and discovery of CPU prioritization based on the detected SST configuration.

The first two patches implement detection of SST configuration in `pkg/sysfs`
- pkg/sysfs: implement adaptation for SST
  Add rudimentary read-only support for interfacing with the kernel
  control interface of Intel Speed Select Technology (SST).

  This functionality is not using the sysfs interface but the sysfs is the
  most convenient location for the code because of the high dependency to
  topology detection features.

  The implementation of the ioctl interface is based on Linux kernel
  headers, the source code of the intel-speed-select tool under Linux
  kernel source tree [[1]] and SST mailbox interface description [[2]].
- pkg/sysfs: wire sst support into system discovery
  Enable automatic detection of Intel Speed Select Technologies (SST) via
  CRI-RM system discovery.

The third patch implements SST-based CPU priority discovery in the cpuallocator:
- cpuallocator: support SST
  Discover CPU priority set via Intel Speed Select Technologies (SST). SST
  Turbo Frequency (SST-TF), Core Power (SST-CP) and Base Frequency
  (SST-BF) are supported. The heuristics of determining CPU priority are
  simple and we do not try to analyze the configuration parameters deeper.
  That could be improved in the future even though this path is covered
  with potholes and pitfalls because of large configuratio space, overlap
  with HWP (khobs in cpufreq/intel_pstate driver) and platform
  differences.

  The current implementation only uses one source of information (per CPU
  package) for determining CPU priority: first SST if that has been
  enabled, falling back to cpufreq if SST is not supported or is disabled.

Based on #602 

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/tools/power/x86/intel-speed-select?h=v5.10.11
[2]: https://github.com/intel/CommsPowerManagement/blob/5155b11e2ea66c8989a8368ec671af9243568e7c/intel_sst_os_interface/mailbox.md


pkg/sysfs/sst.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// IOCTL numbers
+// Derived from kernel headers linux/isst_if.h
+const (
+	ISST_IF_GET_PHY_ID   = 0x000000c008fe01
+	ISST_IF_IO_CMD       = 0x0000004008fe02
+	ISST_IF_MBOX_COMMAND = 0x000000c008fe03
+)
+
```
kad, line 220: those would be good to move to separate file and use cgo generation, so those constants can be easily updated from linux header files. also, would be good to split sst low level things into separate files and mark them buildable only on linux/x86-64 arch, as ioctl numbers are arch dependant.

marquiz: > those would be good to move to separate file and use cgo generation, so those constants can be easily updated from linux header files. also, would be good to split sst low level things into separate files and mark them buildable only on linux/x86-64 arch, as ioctl numbers are arch dependant.

I originally had the lower and higher level interface in separate files but joined them when moving the code over to `pkg/sysfs`. But I now agree that it's cleaner to keep separate. Changed back to two separate files.

Good point about the ioctl numbers - I didn't realize that they may vary between architectures. I now implemented auto-generation for the definitions and types (structs) derived from the Linux kernel source tree.

pkg/sysfs/sst_if.go
```diff
@@ -0,0 +1,113 @@
...
+			{Logical_cpu: uint32(cpu)},
+		},
+	}
+
+	if err := isstIoctl(ISST_IF_GET_PHY_ID, uintptr(unsafe.Pointer(&req))); err != nil {
+		return -1, fmt.Errorf("failed to get CPU mapping for cpu %d: %v", cpu, err)
+	}
+
+	return ID(req.Cpu_map[0].Physical_cpu), nil
+}
```
klihub, line 70: nitpick: isn't this package-specific `PUNIT CPU number` the same as already dug out and published by the kernel as `/sys/devices/system/cpu/cpu[0-9]*/topology/core_id` ? If it is the same and we seem to start needing it, wouldn't it be better to add `sysfs.CPU.CoreID() ID` and add a one-liner to pkg/sysfs to read it out during CPU discovery ?

pkg/sysfs/system.go
```diff
@@ -941,6 +961,36 @@ func (sys *system) discoverPackages() error {
...
 }
 
+func (sys *system) discoverSst() error {
+	if !SstSupported() {
+		sys.Info("Speed Select Technology (SST) support not detected")
+		return nil
+	}
+
+	for _, pkg := range sys.packages {
+
```
klihub, line 971: nitpick: is this blank line intentional ?

docs/policy/cpu-allocator.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. If SST-TF has been enabled, all CPUs prioritized by SST-TF are flagged as
+   high priority.
+1. If SST-CP is enabled but SST-TF disabled, the CPU allocator examines the
+   active CLOSes and their parameters. CPUs associated with the highest priority
+   CLOS will be flagged as high priority, lowest priority CLOS will be flagged
+   as low priority and possible "middle priority" CLOS as normal priority.
+1. If SST-BF has been enabled and SST-TF and SST-CP are inactive, all BF high
+   priority cores (having higher guaranteed base frequency) will be flagged
+   as high priority.
```
klihub, line 44: nitpick: all list items are numbered `1.`

klihub (commented):
A few nitpicks. Otherwise looks good to me. I don't have an informed opinion about the SST detection bits as I don't know anything about the SST details.

pkg/cpuallocator/allocator.go
```diff
@@ -446,5 +399,255 @@ func newTopologyCache(sys sysfs.System) topologyCache {
...
+		return "normal"
+	case PriorityLow:
+		return "low"
+	}
+	return "none"
+}
+
+// cmpCPUSet compares two cpusets in terms of preferred cpu priority. Returns:
+//   > 0 if cpuset A is preferred
+//   < 0 cpuset B is preferred
```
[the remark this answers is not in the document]
marquiz: Fixed in the underlying PR (#602)

pkg/sysfs/sst_if.go
marquiz: That is a good question I pondered myself at some point. I don't think it necessarily is. This punit core id solving is done this way in the `intel-speed-select` tool in the Linux kernel sources so I think they know better and it's better to follow their footsteps. I'd leave this now as is. I'll ping the kernel developers and see what they say. If they confirm your suspicion then we can change this (and I'll possibly able to write a kernel patch 😄 ) WDYT?

pkg/sysfs/system.go
marquiz: Nope. Removed

docs/policy/cpu-allocator.md
marquiz: Yeah, that's the markdown standard. Linter nitpicks otherwise

docs/policy/cpu-allocator.md
```diff
@@ -0,0 +1,59 @@
...
+(SST) features, i.e. Speed Select Technology Performance Profile (SST-PP), Base
+Frequency (SST-BF), Turbo Frequency (SST-TF) and Core Power (SST-CP).
+
+CPU prioritization is based on detection of the currently active SST features
+and their parameterization:
+
+1. If SST-TF has been enabled, all CPUs prioritized by SST-TF are flagged as
+   high priority.
+1. If SST-CP is enabled but SST-TF disabled, the CPU allocator examines the
+   active CLOSes and their parameters. CPUs associated with the highest priority
```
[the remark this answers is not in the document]
marquiz: Good suggestion. Fixed

pkg/sysfs/system.go
```diff
@@ -664,6 +678,12 @@ func (c *cpu) Isolated() bool {
 	return c.isolated
 }
 
+// SstClos returns the Spped Select Core Ppwer CLOS number assigned to the CPU
```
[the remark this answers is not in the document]
marquiz: Thanks. Fixed

marquiz:
Rebased and addressed comments.

Code generation changed so that the actual golang code blocks were moved out of the generator script to their own files. It actually turned out nicer this way

pkg/sysfs/sst_if.go
klihub: Sure. That's fine. This is not really a concern, more like wondering...

pkg/sysfs/system.go
```diff
@@ -664,6 +678,12 @@ func (c *cpu) Isolated() bool {
 	return c.isolated
 }
 
+// SstClos returns the Speed Select Core Ppwer CLOS number assigned to the CPU
```
klihub, line 681: typo: still one more typo on that line... s/Ppwer/Power/

marquiz: Thanks. Fixed

klihub (approved):
LGTM.

askervin (approved):
LGTM.
All E2E tests pass.

marquiz:
Check #602 first

PR #607, opened 2021-02-01
klog: skip headers for journald by default.

klihub:
When logging to a journald stream default to logging
without headers to avoid logging date and time stamp
twice.


marquiz (approved):
👍  lgtm

PR #608, opened 2021-02-01
cri-resmgr: increase allowed service journal log bursts.

klihub:
Increate journald log bursts to avoid losing messages
when running with extensive debugging enabled.


marquiz (approved):
lgtm

PR #609, opened 2021-02-01
fixes: topology-aware policy cpu/memory accounting fixes.

klihub:
**Note:** This is a stacked PR. You probably want to review the following underlying PRs first:
- [x] [#612: topology-aware: force reserved/kube-system containers to the root](https://github.com/intel/cri-resource-manager/pull/612)
- [x] [#613: e2e: add tests for memset expansion in topology-aware policy](https://github.com/intel/cri-resource-manager/pull/613)
- [x] [#614: resource-manager,dump: dump extra data for message disambiguation](https://github.com/intel/cri-resource-manager/pull/614)
- [x] [#616: topology-aware: memory accounting and memset expansion fixes](https://github.com/intel/cri-resource-manager/pull/616)
- [x] [#617: resource-manager: catch containers earlier when they are gone](https://github.com/intel/cri-resource-manager/pull/617)

This full stack of PRs fixes a number of CPU and memory accounting bugs in the topology-aware
policy. In particular it
- fix the logic for expanding container cpuset.mems for OOM-avoidance
- fixes CPU allocation/accounting leaks for failed allocations
- fixes memory allocation/accounting leaks for failed allocations
- fixes pools being double-charged for already allocated memory

Additionally, the stack of PRs
- intercepts CRI `StopPodSandbox` and `ListContainer` messages to detect gone containers earlier
- forces reserved/kube-system containers to the root pool
- improves disambiguation of CRI request/response pairs in message dumps
- improves both the extend and readability of debug logs in the topology-aware policy
- adds end-to-end tests for memset expansion to avoid container OOM

### DIFF
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -264,6 +264,8 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
 func (m *resmgr) StopPod(ctx context.Context, method string, request interface{},
 	handler server.Handler) (interface{}, error) {
 
+	reply, rqerr := handler(ctx, request)
+
 	m.Lock()
 	defer m.Unlock()
 
@@ -272,13 +274,6 @@ func (m *resmgr) StopPod(ctx context.Context, method string, request interface{}
 
 	if !ok {
 		m.Warn("%s: failed to look up pod %s, just passing request through", method, podID)
-	} else {
-		m.Info("%s: stopping pod %s (%s)...", method, pod.GetName(), podID)
-	}
-
-	reply, rqerr := handler(ctx, request)
-
-	if !ok {
 		return reply, rqerr
 	}
 
@@ -286,6 +281,8 @@ func (m *resmgr) StopPod(ctx context.Context, method string, request interface{}
 		m.Error("%s: failed to stop pod %s: %v", method, podID, rqerr)
 		return reply, rqerr
 	}
+
+	m.Info("%s: stopped pod %s (%s)...", method, pod.GetName(), podID)
 
 	released := []cache.Container{}
 	for _, c := range pod.GetInitContainers() {
@@ -319,6 +316,8 @@ func (m *resmgr) StopPod(ctx context.Context, method string, request interface{}
 func (m *resmgr) RemovePod(ctx context.Context, method string, request interface{},
 	handler server.Handler) (interface{}, error) {
 
+	reply, rqerr := handler(ctx, request)
+
 	m.Lock()
 	defer m.Unlock()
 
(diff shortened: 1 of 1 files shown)


kad:
As all pre-PRs got merged, maybe it is time to get that one in?

klihub:
> As all pre-PRs got merged, maybe it is time to get that one in?

@marquiz, @askervin Do you spot any problems or otherwise oppose the two
commits inlined in this PR ? If not, then we could kick this forward.

One thing I'd like to ideally add on top of this is administering separately the
(internal) resource assignment state (assigned/released) from the container
state. That would allow us to do only a single internal resource release cycle
during a regularly performed StopC-RemoveC-StopP-RemoveP wind down
sequence. But that can be dealt with in a separate commit/PR.

pkg/cri/resource-manager/requests.go
```diff
@@ -566,7 +566,7 @@ func (m *resmgr) RemoveContainer(ctx context.Context, method string, request int
 			method, container.PrettyName(), err)
 	}
 
-	m.cache.DeleteContainer(container.GetCacheID())
+	container.UpdateState(cache.ContainerStateStale)
```
askervin, line 569: It seems to me that elsewhere the order is
1. `ReleaseResources(container)`
2. `container.UpdateState(stale)`
3. `runPostReleaseHooks(container)`

but here in `RemoveContainer()` the order (2) and (3) go the other way around.

askervin (commented):
Now that a container is in the stale state if and only if its resources have been released,  do you think we could make `policy` interface more convenient by calling `ReleaseResources` for each container at most once?

pkg/cri/resource-manager/requests.go
klihub: Please, read my comment regarding this above, in the very same message where I pinged you and marquiz...

klihub: > but here in `RemoveContainer()` the order of (2) and (3) is the other way around.

Hmm... I'll need to recheck that. That would leave the container lingering in the cache
in a stale state and while it should have no effect on resource allocations we still don't
want that.

askervin: @klihub , thanks, I'm completely fine with separate single-`ReleaseResources` patch to make the release sequence nice and clean in policy perspective. Sorry about reading only the code and forgetting the latter part of the message.

klihub: @askervin Now updating the state and running the hooks should be more consistent.

klihub: @askervin The lingering containers were a thinko introduced by the initial version of that 'don't jump the gun' commit: neither `RemoveContainer()` nor `runPostReleaseHooks()` was removing the directly passed stale removed container (released). That should be fixed now.

klihub: > @klihub , thanks, I'm completely fine with separate single-`ReleaseResources` patch to make the release sequence nice and clean in policy perspective. Sorry about reading only the code and forgetting the latter part of the message.

No problemo. It is only good that you raised it, because it could have been as well so that I haven't had considered it. This time it just happened to be so that I have.

marquiz (approved):
lgtm

PR #610, opened 2021-02-04
resource-manager: force full reallocation when switching policies.

klihub:
Normally it is enough to release stale containers and reallocate newly discovered
ones when we do our initial sync with the runtime during startup. However, when
switching the active policy is allowed and it does happen during startup, we need
to force reallocating everything. Otherwise containers already known and present
in the cache would not get properly updated by the newly activated policy.

### DIFF
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -60,6 +60,20 @@ func (m *resmgr) startRequestProcessing() error {
 		return err
 	}
 
+	//
+	// Notes:
+	//   While normally it is enough to release stale containers and allocate
+	//   newly discovered ones, if we are switching policies we need to force
+	//   reallocating everything. Otherwise containers already present in the
+	//   cache would not get properly updated by the new policy.
+	//
+	if m.policySwitch {
+		containers := m.cache.GetContainers()
+		cache.SortContainers(containers)
+		add, del = containers, containers
+		m.policySwitch = false
+	}
+
 	if err := m.policy.Start(add, del); err != nil {
 		return resmgrError("failed to start policy %s: %v", policy.ActivePolicy(), err)
 	}
--- a/pkg/cri/resource-manager/resource-manager.go
+++ b/pkg/cri/resource-manager/resource-manager.go
@@ -61,6 +61,7 @@ type resmgr struct {
 	relay        relay.Relay        // our CRI relay
 	cache        cache.Cache        // cached state
 	policy       policy.Policy      // resource manager policy
+	policySwitch bool               // active policy is being switched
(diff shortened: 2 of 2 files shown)


marquiz (dismissed):
Thanks @klihub ! Looks good and I think we can merge this 👍 

One question I started to wonder (I don't remember the code on top of my head) is that do we add the new containers in random or some meaningful order? If it's not already that way, would it be easy sort the containers by Pod id (could be left as a future improvement)?

pkg/cri/resource-manager/requests.go
```diff
@@ -60,6 +60,19 @@ func (m *resmgr) startRequestProcessing() error {
...
+	//   While normally it is enough to release stale containers and allocate
+	//   newly discovered ones, if we are switching policies we need to force
+	//   reallocating everything. Otherwise containers already present in the
+	//   cache would not get properly updated by the new policy.
+	//
+	if m.policySwitch {
+		containers := m.cache.GetContainers()
+		cache.SortContainers(containers)
+		add, del = containers, containers
+	}
```
askervin, line 74: Would the if branch be good place to add
```
m.policySwitch = false
```
...some other place might do as well. But I'm somehow missing that line...

askervin (commented):
A question on resetting `m.policySwitch`.

pkg/cri/resource-manager/requests.go
klihub: > One question I started to wonder (I don't remember the code on top of my head) is that do we add the new containers in random or some meaningful order? If it's not already that way, would it be easy sort the containers by Pod id (could be left as a future improvement)?

Since we disrupt containers here we might as well try to optimize their placement. Therefore this patch
strongly suggests for the active policy to reallocate containers in most-important-largest-first order by
sorting the container slice by QoS/memory/CPU before passing it on to the policy.

klihub: > Would the if branch be good place to add
> 
> ```
> m.policySwitch = false
> ```
> 
> ...some other place might do as well. But I'm somehow missing that line...

Thanks for reminding me. it was supposed to be there for clarity and regularity, regardless
of this code never being run twice, but it slipped my mind.

marquiz: > Since we disrupt containers here we might as well try to optimize their placement. Therefore this patch
> strongly suggests for the active policy to reallocate containers in most-important-largest-first order by
> sorting the container slice by QoS/memory/CPU before passing it on to the policy.

Ah, sorry, I somehow missed the `SortContainers()` in this patch 😬 All good 😄

askervin (approved):
LGTM!

PR #614, opened 2021-02-07
resource-manager,dump: dump extra data for message disambiguation.

klihub:
**Note:** This PR is also merged into stacked [PR #609](https://github.com/intel/cri-resource-manager/pull/609).

This PR adds extra data to dumped `CRI` messages to help disambiguate corresponding pairs of request/response messages.

### DIFF
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -51,6 +51,75 @@ func (m *resmgr) setupRequestProcessing() error {
 	return nil
 }
 
+// disambiguate produces disambiguation context for a request/reply dump.
+func (m *resmgr) disambiguate(msg interface{}) string {
+	var qualifier string
+
+	m.RLock()
+	defer m.RUnlock()
+
+	switch req := msg.(type) {
+	case *criapi.RunPodSandboxRequest:
+		if req.Config != nil && req.Config.Metadata != nil {
+			qualifier = req.Config.Metadata.Name
+		}
+	case *criapi.StopPodSandboxRequest:
+		if pod, ok := m.cache.LookupPod(req.PodSandboxId); ok {
+			qualifier = pod.GetName()
+		} else {
+			qualifier = "unknown pod " + req.PodSandboxId
+		}
+	case *criapi.RemovePodSandboxRequest:
+		if pod, ok := m.cache.LookupPod(req.PodSandboxId); ok {
+			qualifier = pod.GetName()
+		} else {
+			qualifier = "unknown pod " + req.PodSandboxId
+		}
+
+	case *criapi.CreateContainerRequest:
+		switch {
+		case req.SandboxConfig == nil || req.SandboxConfig.Metadata == nil:
+			qualifier = "missing pod metadata in request"
+		case req.Config == nil || req.Config.Metadata == nil:
+			qualifier = "missing metadata in request"
+		default:
+			qualifier = req.SandboxConfig.Metadata.Name + ":" + req.Config.Metadata.Name
+		}
+
+	case *criapi.StartContainerRequest:
+		if container, ok := m.cache.LookupContainer(req.ContainerId); ok {
+			qualifier = container.PrettyName()
+		} else {
+			qualifier = "unknown container " + req.ContainerId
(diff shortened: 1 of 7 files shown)


pkg/cri/resource-manager/requests.go
```diff
@@ -51,6 +51,75 @@ func (m *resmgr) setupRequestProcessing() error {
...
+			qualifier = pod.GetName()
+		} else {
+			qualifier = "unknown pod " + req.PodSandboxId
+		}
+	case *criapi.RemovePodSandboxRequest:
+		if pod, ok := m.cache.LookupPod(req.PodSandboxId); ok {
+			qualifier = pod.GetName()
+		} else {
+			qualifier = "unknown pod " + req.PodSandboxId
+		}
```
marquiz, line 77: Could these be combined with `fallthrough`(?)

pkg/cri/resource-manager/requests.go
```diff
@@ -51,6 +51,75 @@ func (m *resmgr) setupRequestProcessing() error {
...
+		} else {
+			qualifier = "unknown container " + req.ContainerId
+		}
+
+	case *criapi.UpdateContainerResourcesRequest:
+		if container, ok := m.cache.LookupContainer(req.ContainerId); ok {
+			qualifier = container.PrettyName()
+		} else {
+			qualifier = "unknown container " + req.ContainerId
+		}
```
marquiz, line 113: Also these could be combined with `fallthrough` (?)

pkg/dump/dump.go
```diff
@@ -123,6 +125,7 @@ func ReplyMessage(kind, name string, rpl interface{}, latency time.Duration, syn
 			dir:     reply,
 			kind:    kind,
 			method:  name,
+			qualif:  qualif,
```
marquiz, line 128: super nit: "qualif" just looks very odd abbreviation, "qualifier" would not be that much longer and would be more descriptive

pkg/cri/resource-manager/requests.go
klihub: At least I can't see how. At least not in a way that'd give reasonable reduction of (syntactic) code duplication. And the reason I don't see a way is the type switch. For this most obvious try:
```
	switch req := msg.(type) {
...
	case *criapi.StopPodSandboxRequest,	*criapi.RemovePodSandboxRequest:
		if pod, ok := m.cache.LookupPod(req.PodSandboxId); ok {
			qualifier = pod.GetName()
		} else {
			qualifier = "unknown pod " + req.PodSandboxId
		}
		if pod, ok := m.cache.LookupPod(req.PodSandboxId); ok {
			qualifier = pod.GetName()
		} else {
			qualifier = "unknown pod " + req.PodSandboxId
		}
```
the compiler seems to generate code where `req` in that combined case has the generic `interface{}` type, so derefering fields don't work. You'd need to switch on the type again which destroys the whole point.. Golang does not have real duck-typing even if it syntactically might occasionally seem so.

But of course I might miss something mode obvious here...

pkg/cri/resource-manager/requests.go
klihub: Not that I can quickly see... See my previous comment above.

pkg/cri/resource-manager/requests.go
marquiz: Ach, you're correct. I was reading through the code too quickly.

pkg/cri/resource-manager/requests.go
marquiz: Sorry for the noist

pkg/dump/dump.go
klihub: Updated, changed to `qualifier`.

marquiz (approved):
Thanks @klihub, and, apologies for the noisy review

LGTM

klihub:
> Thanks @klihub, and, apologies for the noisy review

No problemo. Those were valid comments/questions. And neither did it occur to me originally to try your suggested combined type switch nor did I know, once you suggested it, whether it would work/compile. So I tried it to find it out. 
 
> LGTM

PR #615, opened 2021-02-07
topology-aware: better and more readable logs.

klihub:
**Note:** This PR is also merged into stacked [PR #609](https://github.com/intel/cri-resource-manager/pull/609).

This PR improves debug logging in the topology aware policy. In praticular, it
- changes debug logs to use container `PrettyName()`s
- produces more readable dumps of `memoryMaps`
- dumps the memory accounting state of pool tree nodes

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
@@ -391,7 +391,11 @@ func (p *policy) allocatePool(container cache.Container, poolHint string) (Grant
 			request, supply.DumpAllocatable(), err)
 	}
 
-	log.Debug("allocated req '%s' to memory node '%s' (memset %s,%s)", container.GetCacheID(), grant.GetMemoryNode().Name(), grant.GetMemoryNode().GetMemset(memoryDRAM), grant.GetMemoryNode().GetMemset(memoryPMEM))
+	log.Debug("allocated req '%s' to memory node '%s' (memset %s,%s,%s)",
+		container.PrettyName(), grant.GetMemoryNode().Name(),
+		grant.GetMemoryNode().GetMemset(memoryDRAM),
+		grant.GetMemoryNode().GetMemset(memoryPMEM),
+		grant.GetMemoryNode().GetMemset(memoryHBM))
 
 	// In case the workload is assigned to a memory node with multiple
 	// child nodes, there is no guarantee that the workload will
@@ -563,7 +567,8 @@ func (p *policy) allocatePool(container cache.Container, poolHint string) (Grant
 					return nil, err
 				}
 				if changed {
-					log.Debug("* moved container %s upward to node %s to guarantee memory", oldGrant.GetContainer().GetCacheID(), oldGrant.GetMemoryNode().Name())
+					log.Debug("* moved container %s upward to node %s to guarantee memory",
+						oldGrant.GetContainer().PrettyName(), oldGrant.GetMemoryNode().Name())
 					break
 				}
 			}
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
@@ -82,6 +82,8 @@ type Supply interface {
 	DumpCapacity() string
 	// DumpAllocatable returns a printable representation of the supply's alloctable resources.
 	DumpAllocatable() string
+	// DumpMemoryState dumps the state of the available and allocated memory.
+	DumpMemoryState(string)
 }
 
 // Request represents CPU and memory resources requested by a container.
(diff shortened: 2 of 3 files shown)


pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
```diff
@@ -391,7 +391,10 @@ func (p *policy) allocatePool(container cache.Container, poolHint string) (Grant
 			request, supply.DumpAllocatable(), err)
 	}
 
-	log.Debug("allocated req '%s' to memory node '%s' (memset %s,%s)", container.GetCacheID(), grant.GetMemoryNode().Name(), grant.GetMemoryNode().GetMemset(memoryDRAM), grant.GetMemoryNode().GetMemset(memoryPMEM))
+	log.Debug("allocated req '%s' to memory node '%s' (memset %s,%s)",
```
marquiz, line 394: Should we be prepared for HBM, too(?)

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -830,6 +839,73 @@ func (cs *supply) DumpAllocatable() string {
...
+// prettyMem formats the given amount as k, M, G, or T units.
+func prettyMem(amount uint64) string {
+	units := []string{"k", "M", "G", "T"}
+	a := float64(amount)
+	f := a
+	u := ""
+	for _, unit := range units {
+		if f /= float64(1024); f < 1 {
+			break
+		}
```
marquiz, line 851: Do I read this correctly that now we only get pretty values less than 1. I think we should check for <1024 so that we'd get values between [0, 1024[ (?)

Also a super nit^10 that does not really matter: instead of consecutive floating point divisions I'd just compare integer values and only do the division when we find the limit

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -830,6 +839,73 @@ func (cs *supply) DumpAllocatable() string {
...
+	memTypes := []memoryType{memoryDRAM, memoryPMEM, memoryHBM}
+	totalFree := uint64(0)
+	totalGranted := uint64(0)
+	for _, kind := range memTypes {
+		free := cs.mem[kind]
+		granted := cs.grantedMem[kind]
+		if free != 0 || granted != 0 {
+			log.Debug(prefix+"- %s: free: %s, granted %s",
+				kind, prettyMem(free), prettyMem(granted))
+		}
```
marquiz, line 869: `free == 0 || granted != 0` would be impossible or a bug(?) Should we print something in that case?

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -318,18 +320,25 @@ func (m memoryMap) AddHBM(hbm uint64) {
...
+			types++
 		}
 		if hbm > 0 {
-			mem += sep + "hbm:" + strconv.FormatUint(hbm, 10)
+			mem += sep + "HBM " + prettyMem(hbm)
+			types++
+		}
+		if types > 1 {
+			mem += sep + "total " + prettyMem(pmem+dram+hbm)
+			types++
```
askervin, line 341: Could we drop line 341 and leave `types` to the number of found memory types?

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
klihub: > Do I read this correctly that now we only get pretty values less than 1. I think we should check for <1024 so that we'd get values between [0, 1024[ (?)

No, that's exactly the range. We stop dividing once the result goes below 1, but the value used is from the previous iteration of the loop. You get logs like this:
```
D: [     policy     ] <post-alloc>      - DRAM: free: 108.58G, granted 80.00G
D: [     policy     ] <post-alloc>      - total free: 108.58G, total granted 80.00G
D: [     policy     ] <post-alloc>      - extra reservations:
D: [     policy     ] <post-alloc>        - coredns-f9fd979d6-w4qpw:coredns: 170.00M (DRAM 170.00M)
D: [     policy     ] <post-alloc>        - coredns-f9fd979d6-4g982:coredns: 170.00M (DRAM 170.00M)
```

askervin (changes requested):
One request in code + `s/Prerry/Pretty/` in the commit message.

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
marquiz: True, I prolly misread the `'f'` below in `FormatFloat` as `f` or smth

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
klihub: I'm not sure what you mean... Did you mean `free == 0 && granted != 0` ?

Anyway, the invariant wrt. free `(cs.mem[kind])` and granted (`cs.grantedMem[kind]`) for any node is that `free + granted == total available`. Whenever an allocation is done, that's worth is moved from free to granted (this was exactly one of the bugs, the inconsistent use of free) and the other way around when it is released.

Now with that in mind, pretty much any combinations of free and granted are possible (maybe with the exception of `free == 0 && granted == 0` since we don't have memory-less nodes in the tree).

pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
klihub: Yes, since the code does have preliminary support for HBM, maybe we could dump that here, too...

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
marquiz: Argh, sorry, again I was reading too quickly. Or, maybe I should get a pair of glasses 😬   I was reading `&&` instead of `||`, sorry about that...

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
klihub: Sure. Updated.

pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
klihub: Added.

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
klihub: Updated to do the conversion with comparisons and a single float64 division.

askervin (approved):
LGTM

marquiz (approved):
LGTM 👍 

Love the integer arithmetics 😉

PR #616, opened 2021-02-07
topology-aware: memory accounting and memset expansion fixes.

klihub:
**Note 1:** This PR is stacked on [#615](https://github.com/intel/cri-resource-manager/pull/615). You probably want to review that separately first.

**Note 2:** This PR is also merged into stacked [PR #609](https://github.com/intel/cri-resource-manager/pull/609).

This PR
- fixes the memset expansion logic in `ExpandMemset()`
- fixes accounting to not leak resources on failed CPU allocation
- fixes accounting to not leak memory on (partially) failed memory allocation
- fixes a bug that double-charges pool tree nodes for already allocated memory
- adds debug logs to the memory accounting code

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
@@ -1390,7 +1390,6 @@ func (cg *grant) RestoreMemset() {
 
 func (cg *grant) ExpandMemset() (bool, error) {
 	supply := cg.GetMemoryNode().FreeSupply()
-	mems := cg.MemLimit()
 	node := cg.GetMemoryNode()
 	parent := node.Parent()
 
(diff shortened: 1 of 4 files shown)


pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -586,10 +617,14 @@ func (cs *supply) Allocate(r Request) (Grant, error) {
 			cs.node.Name(), full, cs.sharable, cs.AllocatableSharedCPU())
 	}
 
+	grant := newGrant(cs.node, cr.GetContainer(), cr.cpuType, exclusive, 0, 0, nil, 0)
+	grant.AccountAllocateCPU()
```
marquiz, line 621: I'm really not that familiar with the code but would it make the code simpler if we moved this after the fractional allocation below - would remove the need for the release calls there (?)

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -473,7 +475,7 @@ func (cs *supply) allocateMemory(cr *request) (memoryMap, error) {
 		available := cs.mem[memoryPMEM] - cs.grantedMem[memoryPMEM]
 
 		log.Debug("%s: trying %s from PMEM, available %s",
-			cr.GetContainer().PrettyName(),
+			r.GetContainer().PrettyName(),
```
askervin, line 478: This is a matter of taste, but I'd think this code would be less fragile if we'd have the same reservation code routine for every memory type:

```
for mem in [memoryPMEM, memoryDRAM, memoryHBM]:
    if remaining == 0: break
    if remaining < available:
        ...
    else:
        ...
```

Sure, it'll need the coldstart check after PMEM reservation inside the loop, but I'd still think it would make this easier to read and test.

WDYT?

klihub: Yes, I agree. Especially, since related code we've fixed or added in the scope of this and other #609-related PRs were updated and written like that. I'll update this one, too.

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
klihub: It wouldn't make to code any simpler IMO. After a failure you'd still need to return the exclusively allocated CPUs to the to where they were carved out from (call ReleaseCPU(), because takeCPU() has already removed them). AccountAllocate was deliberately put where it is so that after that point we are in fully internally consistent state: exclusive part of the allocation both done and accounted for.

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
klihub: Refactored `allocateMemory()`. Still passes unit tests.

marquiz (commented):
I think this is good to go. Let's just wait for #615, first

marquiz (approved):
✅

askervin (approved):
LGTM. And allocateMemory is now better than ever.

PR #617, opened 2021-02-07
resource-manager: catch containers earlier when they are gone.

klihub:
**Note 1:** This a stacked PR. You probably want to review the following underlying PRs separately first:
- [x] [#616: topology-aware: memory accounting and memset expansion fixes](https://github.com/intel/cri-resource-manager/pull/616)

**Note 2:** This PR is also merged into stacked [PR #609](https://github.com/intel/cri-resource-manager/pull/609).

This PR adds code to intercept CRI `StopPodSandbox()` and `ListContainers()`
requests to detect better/earlier when a container is not longer running and to
release the resources of such a container.

### DIFF
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -448,6 +448,7 @@ func (m *resmgr) StopContainer(ctx context.Context, method string, request inter
 
 	if rqerr != nil {
 		m.Error("%s: failed to stop container %s: %v", method, container.PrettyName(), rqerr)
+		return reply, rqerr
 	}
 
 	// Notes:
--- a/pkg/cri/resource-manager/requests.go
+++ b/pkg/cri/resource-manager/requests.go
@@ -39,6 +39,7 @@ func (m *resmgr) setupRequestProcessing() error {
 		"StartContainer":  m.StartContainer,
 		"StopContainer":   m.StopContainer,
 		"RemoveContainer": m.RemoveContainer,
+		"ListContainers":  m.ListContainers,
 
 		"UpdateContainerResources": m.UpdateContainer,
 	}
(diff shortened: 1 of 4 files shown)


pkg/cri/resource-manager/requests.go
```diff
@@ -189,6 +190,60 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
...
+		released = append(released, c)
+	}
+	for _, c := range pod.GetContainers() {
+		m.Info("%s: releasing resources for container %s...", method, c.PrettyName())
+		if err := m.policy.ReleaseResources(c); err != nil {
+			m.Warn("%s: failed to release container %s: %v", method, c.PrettyName(), err)
+		}
+		c.UpdateState(cache.ContainerStateExited)
+		released = append(released, c)
+	}
```
marquiz, line 236: Just wondered is it safe to release if the request fails or would it be better to wait for re-try? 🤔  E.g. the container runtime was not running at that very moment and didn't get the request. Kubelet tries again later at which point we've already lost track of the containers.

pkg/cri/resource-manager/requests.go
```diff
@@ -512,6 +513,61 @@ func (m *resmgr) RemoveContainer(ctx context.Context, method string, request int
...
+	if f := request.(*criapi.ListContainersRequest).Filter; f != nil {
+		switch {
+		case f.Id != "":
+			return reply, nil
+		case f.State != nil:
+			return reply, nil
+		case f.PodSandboxId != "":
+			return reply, nil
+		case len(f.LabelSelector) > 0:
+			return reply, nil
```
marquiz, line 535: I guess here we could use `fallthrough`(?) 😄

pkg/cri/resource-manager/requests.go
```diff
@@ -512,6 +513,61 @@ func (m *resmgr) RemoveContainer(ctx context.Context, method string, request int
...
+		}
```
marquiz, line 536: I guess here one could use fallthrough
```
		switch {
		case f.Id != "",
		case f.State != nil,
		case f.PodSandboxId != "",
		case len(f.LabelSelector) > 0:
			return reply, nil
		}
```

but I don't know how much difference it really makes

klihub: Hmm... I was under the impression that a case list is just syntactic sugar and it compiles to a similar/exactly same if-else if-else if construct than a normal switch without case lists.
 
And I guess this is waaay off-topic, but is that really called a `fallthrough` ? I was under the impression that a fallthrough refers to the explicit usage of the `fallthrough` keyword in golang, since unlike in C, by default the cases in a switch construct break instead of falling through.

But your suggestion has the benefit of emphasizing the identical return values in all those cases, so it's probably better in this case.

pkg/cri/resource-manager/requests.go
klihub: Hmm... It's a valid question and I'm not sure which is a/the better way to go.

My approach was simply based on the observation that kubelet will not try to stop a pod unless it thinks all the containers in the pod are already stopped/have exited. Hence, it seemed/seems like a perfect synchronization point to make sure that we are of the same opinion.

Or, to look at it form an alternative perspective, we really don't know why the runtime failed to stop the pod. If it did because the pod does not exist any more, but both the kubelet and us still beg to differ, then ATM we have an effective allocation of resources for something which is not there. That allocation must eventually be (internally) released to prevent it from becoming a resource leak. And if the pod has multiple containers, we potentially have a bunch of those at this point.

I can change this to bail out on error without releasing the containers. Then (assuming that kubelet will not try to indefinitely stop/remove pods if those operations keep failing in the runtime) we have to use intercepted  `ListContainers()` with more complex logic than the current one (which simply releases containers that has recently exited). For some reason I just have a very much Matt Kowalski / Mardi Gras '87 feeling about that...

pkg/cri/resource-manager/requests.go
marquiz: > Hmm... I was under the impression that a case list is just syntactic sugar and it compiles to a similar/exactly same if-else if-else if construct than a normal switch without case lists.

Yes, I think this is the way it is. The idea behind my suggestion was just eliminating identical returns
 
> And I guess this is waaay off-topic, but is that really called a `fallthrough` ? I was under the impression that a fallthrough refers to the explicit usage of the `fallthrough` keyword in golang, since unlike in C, by default the cases in a switch construct break instead of falling through.

Yup, you're correct: fall-through only is the explicit use of `fallthrough` in one case, not combining multiple ifs in one case as a list. My terminology was bad, thanks for correcting me. 

> But your suggestion has the benefit of emphasizing the identical return values in all those cases, so it's probably better in this case.

This was the purpose of my nitty-gritty-suggestion, improving readability 😺  I wasn't sure if was good or bad, though.

klihub: Okay, so that is syntactically incorrect go. What you can do is this:
```
	switch {
	case f.Id != "",
		f.State != nil,
		f.PodSandboxId != "",
		len(f.LabelSelector) > 0:
		    return reply, nil
	}
```
but that gets turned by the tabulation and formatting nazis into this:
```
	switch {
	case f.Id != "",
		f.State != nil,
		f.PodSandboxId != "",
		len(f.LabelSelector) > 0:
		return reply, nil
	}
```
which is not very readable. So I gave in a replaced it with an `if`.

marquiz: Woops, yes, my code example was borked 😬  And, nice work gofmt 😢  Maybe an if is actually the best here, agree

pkg/cri/resource-manager/requests.go
marquiz: I fully trust your judgement and reasoning here. Let's keep it the way you wrote it. Sounds safer to me that way. Anyway, my question was just something that occurred to me while reading the code

marquiz:
LGTM 👍  But only after the dependencies have been merged...

pkg/cri/resource-manager/requests.go
klihub: Well, I already went ahead and updated it to not release containers on a failed pod stop request and added nuking to the processing of intercepted list request. Needs to bang that for a while to see if my worst fears are eased enough or if I need to roll back to the original simpler implementation.

pkg/cri/resource-manager/requests.go
```diff
@@ -189,6 +191,61 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
...
+		m.Info("%s: stopping pod %s (%s)...", method, pod.GetName(), podID)
+	}
+
+	reply, rqerr := handler(ctx, request)
+
+	if !ok {
+		return reply, rqerr
+	}
+
+	if rqerr != nil {
```
kad, line 216: should it handle errors like "already stopped" or similar conditions?

klihub: If you mean if handling errors returned by the runtime like 'already stopped' for StopPodSandbox or StopContainer with a stopped pod's of container's ID, then there are no errors like that. The runtime will reply with success if you try to stop an already stopped container or pod.

askervin (approved):
LGTM

marquiz (approved):
lgtm

PR #621, opened 2021-02-09
e2e: fuzz topology-aware

askervin:


klihub:
That looks really nice !

I was trying to hack together something very similar using pure bash with random-generated sequence of actions of start, stop, restart, sleep N, fail C, exit C, and check (start/stop/restart the pod, sleep for N, stop container C with a simulated failure/success (exit 1/exit 0), checking actual vs. expected state). Instead of signals I was controlling the containers with a per-pod-per-container hostPath mounted /ctrl/ctrl file system entry which was read with a tail -F loop for commands (exit, fail)...

But I'll abort that exercise, this looks so much better, especially I imagine the verification part should be easier.

askervin:
- Removed generated test code.
- Added big fat warning on skipping the test when generated tests are not found.
- Added help on generating tests.
- Added help on building fMBT (test generator) docker image if not found.
- Production Dockerfiles (fmbt-gui and fmbt-cli) added to fMBT devel branch for more decently sized images (0.4 GB vs 0.8 GB vs. 2.5 GB)

klihub (approved):
LGTM

PR #622, opened 2021-02-10
topology-aware: use normal as fallback for reserved.

klihub:
Fall back to using normal CPUs if we have insufficient
reserved CPU capacity to allocate from. Give a warning
about potentially incorrect configuration whenever this
happens.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
@@ -554,10 +554,21 @@ func (cs *supply) AllocateCPU(r Request) (Grant, error) {
 	full := cr.full
 	fraction := cr.fraction
 
-	if cr.cpuType == cpuReserved && full > 0 {
+	cpuType := cr.cpuType
+
+	if cpuType == cpuReserved && full > 0 {
 		log.Warn("exclusive reserved CPUs not supported, allocating %d full CPUs as fractions", full)
 		fraction += full * 1000
 		full = 0
+	}
+
+	if cpuType == cpuReserved && fraction > 0 && cs.AllocatableReservedCPU() < fraction {
+		log.Warn("possible misconfiguration of reserved resources:")
+		log.Warn("  %s: allocatable %s", cs.GetNode().Name(), cs.DumpAllocatable())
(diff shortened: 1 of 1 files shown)


kad:
Should we have a test case for that situation?

marquiz (approved):
From what I understand, lgtm

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -554,12 +554,23 @@ func (cs *supply) AllocateCPU(r Request) (Grant, error) {
...
+	if cpuType == cpuReserved && full > 0 {
 		log.Warn("exclusive reserved CPUs not supported, allocating %d full CPUs as fractions", full)
 		fraction += full * 1000
 		full = 0
 	}
 
+	if cpuType == cpuReserved && fraction > 0 && cs.AllocatableReservedCPU() < fraction {
+		log.Warn("possible misconfiguration of reserved resources:")
+		log.Warn("  %s: allocatable %s", cs.GetNode().Name(), cs.DumpAllocatable())
+		log.Warn("  %s: needs %d reserved, only %d available",
```
askervin, line 568:
```
I0215 09:39:28.755486 3062426 resources.go:948] pod5:pod5c0: CPU preferences: cpuType=reserved, full=0, fraction=2000, isolate=false
W0215 09:39:28.755549 3062426 resources.go:566] possible misconfiguration of reserved resources:
W0215 09:39:28.755616 3062426 resources.go:567]   root: allocatable <root allocatable: CPU: reserved:7,11 (allocatable: 1204m), grantedReserved:796m, sharable:4-6,8-10,12-13 (allocatable:8000m), MemLimit: DRAM 7.41G>
W0215 09:39:28.755640 3062426 resources.go:568]   pod5:pod5c0: needs 2000 reserved, only 1204 available
W0215 09:39:28.755659 3062426 resources.go:570]   falling back to using normal unreserved CPUs instead...
```

This is complete nit-pick and this looks good to me already, but maybe `s/%d/%dm/g` on line 568 would align the units with the root node capacity in the warning better - especially when numbers happen to have only 1-3 digits in some situation.

askervin (approved):
This is very good.

I'll add updated tests.

PR #625, opened 2021-02-15
cache: (over)write cache file atomically.

klihub:


pkg/cri/resource-manager/cache/cache.go
```diff
@@ -1526,8 +1527,13 @@ func (cch *cache) Save() error {
 		return cacheError("failed to save cache: %v", err)
 	}
 
-	if err = ioutil.WriteFile(cch.filePath, data, cacheFilePerm.prefer); err != nil {
-		return cacheError("failed to write cache to file '%s': %v", cch.filePath, err)
+	tmpPath := cch.filePath + ".saving"
+	if err = ioutil.WriteFile(tmpPath, data, cacheFilePerm.prefer); err != nil {
+		return cacheError("failed to write cache to file %q: %v", tmpPath, err)
+	}
+	if err := syscall.Rename(tmpPath, cch.filePath); err != nil {
```
kad, line 1534: em. better to use `os.Rename()`. it will call the same syscall underneath, but will check that there is no problems with source/destinations/links/etc.

kad (changes requested):

pkg/cri/resource-manager/cache/cache.go
klihub: Updated.

kad (approved):
lgtm

PR #626, opened 2021-02-15
resource-manager: save cache at startup.

klihub:
Save the cache as the last action in a successful startup sequence.


askervin (approved):
LGTM

kad (approved):
+1

PR #629, opened 2021-02-16
topology-aware: fix filtering out nodes with insufficient memory

askervin:


askervin:
Unlike the fixed unit test required, container c3 cannot be placed to a leaf node because its parent node (socket 0) gets container c2, and memory required by c2 exceeds available memory in the leaf node. Expected behavior and observed behavior were incorrect, and both are fixed by this commit.

klihub:
LGTM, but I suspect github waits for the conflicts to get resolved before it kicks the checking/testing-workflow forward. Let's do that and merge this if all passes.

klihub (approved):

PR #634, opened 2021-02-17
sample-configs: make cri-resmgr-configmap.example.yaml usable

marquiz:
Comments starting from the beginning of line inside the configmap data
broke the yaml format.


askervin (approved):
LGTM

PR #641, opened 2021-03-05
docs: update Pygments requirements

marquiz:
Sync Pygments version between local and containerized builds. Fixes
possible "lexer name 'toml' is not known" error in local builds.


klihub (approved):

marquiz:
Hold your horses, I'll try to investigate the cause for ci not catching the error

marquiz:
OK, the problem was caused by CI changing from Ubuntu 18.04 to 20.04. With the latter, actually an older version of Pygments will be installed

PR #642, opened 2021-03-10
fixes: fixes for fedora 33

klihub:
This PR switches the fedora cloud image to 33 and the corresponding bootstrap commands
to use grubby for switching systemd to a non-unified v1/v2 cgroupfs mount.

We expect fedora users to have switched by now now to the latest which is version 33.
Since that is glibc-incompatible with the previous ones, switching the cloud image
as well allows running on fedora 33 hosts e2e tests in fedora 33 VMs.


askervin (approved):
LGTM!

PR #646, opened 2021-03-22
cache: add a directory to findContainerDir search path

askervin:
This change enables finding proper container cgroup directory with `container.GetCgroupDir()` on e2e/demo ubuntu/debian VMs.

The problem was found when trying to use future `blockio` from `goresctrl` like this:
```
blockio.SetContainerClass(c.GetCgroupDir(), class)
```

### DIFF
--- a/pkg/cri/resource-manager/cache/utils.go
+++ b/pkg/cri/resource-manager/cache/utils.go
@@ -267,6 +267,7 @@ func findContainerDir(podCgroupDir, podID, ID string) string {
 	cpusetDir := cgroups.Cpuset.Path()
 
 	dirs = []string{
+		path.Join(cpusetDir, podCgroupDir, ID),
 		// containerd, systemd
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/cache/utils.go
```diff
@@ -267,6 +267,7 @@ func findContainerDir(podCgroupDir, podID, ID string) string {
 	cpusetDir := cgroups.Cpuset.Path()
 
 	dirs = []string{
+		path.Join(cpusetDir, podCgroupDir, ID),
```
klihub, line 270: What is the kubelet and containerd configuration wrt. the cgroup driver in this case ? IOW, is this the real correct `containerd, cgroupfs` in all cases ? If it is then is the 3rd entry after inserting this wrong and unnecessary ? If not, when is the 3rd entry correct instead of this ?

klihub (approved):
Ok, let's get this in. We can fix/worry about any unnecessary entries.

PR #650, opened 2021-03-22
blockio: start using blockio from the goresctrl library

askervin:


askervin:
goresctrl logging was changed to klog/logrus-style, causing cri-rm logger interface incompatibility in `blockio.SetLogger(logger.Get("blockio"))`.

This patchset includes commit f76bcdc8362294be4df98b9cff8ea5c594ac9667 that does similar change in cri-rm logger to fix the incompatibity (without a log wrapper).

klihub:
We should update these bits, but I think the right course of action is, once the corresponding containerd bits get merged, to turn cri-resmgr into a postman which merely conveys information to/remote controls goresctrl using the dedicated/associated annotations.

marquiz:
The caveat in this approach is that we don't have any visibility on whether blockio (or rdt) support has actually been enabled in the container runtime. Blindly setting the annotations may cause surprises (unexpected failures)

klihub:
> The caveat in this approach is that we don't have any visibility on whether blockio (or rdt) support has actually been enabled in the container runtime. Blindly setting the annotations may cause surprises (unexpected failures)

Yes that is true. But this is hardly a `cri-resmgr`-specific problem. Isn't the same true for user provided pod-level RDT annotations, and especially potential device plugin injected container level RDT annotations, if they are present in a PodSpec while RDT-support is disabled in the runtime. As long as RDT support can be disabled at all (in the runtime, or in the kernel), the only partial remedy I can see would be to update the runtime integration bits to give an error, or log a warning, when RDT is disabled yet annotations are present. Ditto for Block I/O.

marquiz:
With the current implementation I see the problem in implicitly setting RDT class to pod QoS class.

PR #654, opened 2021-03-31
scripts: kube-cgroups prints cgroup entries per pod/container

askervin:
I've been using this script for testing blkio cgroup parameters. I think it could be useful for you too, as it shows pretty clearly cpuset.cpus/mems for every pod/container, too.


klihub (approved):
LGTM.

PR #656, opened 2021-04-08
e2e: add support for k8s=X.Y.Z to set Kubernetes version

askervin:
- Supported initially on ubuntu-* and debian-* distros.
- Update help to include k8s and distro.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -235,13 +235,17 @@ debian-install-crio-pre() {
 }
 
 debian-install-k8s() {
+    local k8sverparam
     debian-refresh-pkg-db
     debian-install-pkg apt-transport-https curl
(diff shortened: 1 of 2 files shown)


askervin:
This helps answering questions like "does feature X work with k8s version Y".

klihub (commented):
@askervin Any plans to add similar support for CentOS, fedora and SUSE ?

askervin:
> Any plans to add similar support for CentOS, fedora and SUSE ?

@klihub, it was very good idea to take a look how would it work. Wasn't too complex after all. All distros are now supported. This only works a bit differently on SUSE: k8s=X.Y.Z and k8s=X.Y.Å will both install X.Y, because only the latest X.Y.* is available through snap.

Rebased.

klihub:
Probably the current snap-based on is cleaner and good enough.

klihub (approved):

PR #663, opened 2021-06-28
e2e: lib: Use proper locale for bc to work

jukkar:
The current time returned by bash variable EPOCHREALTIME can contain
either "." or "," separator between seconds and microseconds (depending
on used locale), and because that value is feed directly to bc, then it
is possible that bc command will not understand the value and can return
"(standard_in) 1: syntax error"
if the current locale uses "," separator.
    
In order to fix this issue, set LC_NUMERIC=C so that bash uses always "."
when feeding the data to bc command.

### DIFF
--- a/demo/lib/command.bash
+++ b/demo/lib/command.bash
@@ -10,6 +10,8 @@
 # command-start and command-end set environment variables:
 # COMMAND, COMMAND_STATUS, COMMAND_OUTPUT
 
+export LC_NUMERIC=C
+
 # These exports force ssh-* to fail instead of prompting for a passphrase.
 export DISPLAY=bogus-none
 export SSH_ASKPASS=/bin/false


jukkar:
I need to modify this slightly, it works ok in Fedora but when run in Ubuntu I still get the `syntax error` print.

jukkar:
New version avoids using `bc` because it does not understand float numbers that are formatted according to the locale.

askervin:
Currently - with a tested locale setting - this produces timestamps in `output/commands/*` files like:
```
# start time: 7.222846
# command: dnf install -y runc conmon
Last metadata expiration check: 0:11:20 ago on Tue 29 Jun 2021 07:11:36 AM UTC.
Package runc-2:1.0.0-378.rc95.fc33.x86_64 is already installed.
Package conmon-2:2.0.27-2.fc33.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
# exit status: 0
# end time: 10.352635
```

That is, you can find command, output, exit status and start/end time from its own file for each host-command and vm-command executed in the script.

I'm afraid this change would make the timestamps harder to read.

Do you think we could still try `export LC_NUMERIC=C` and, if we want to get rid of `bc`, then maybe use `awk` for the computation. It's a host dependency anyway. And more often available than bc.

jukkar:
> Do you think we could still try `export LC_NUMERIC=C`

This seems to be the simplest solution. I did some experiments yesterday with various LC_ vars but had some mixed results between Fedora and Ubuntu. But now when trying it again it worked just fine, I probably did not try the LC_NUMERIC var earlier.

askervin (approved):
LGTM. Thanks Jukka!

klihub (approved):

PR #673, opened 2021-07-02
e2e: enable running without govm if VM_IP is set

askervin:
If VM_IP is set, check that we can run commands as root on the host.

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -43,6 +43,19 @@ vm-ssh-user() {
 }
 
 vm-check-env() {
+    # If VM IP address is already defined, govm is not needed.
+    if [ -n "$VM_IP" ]; then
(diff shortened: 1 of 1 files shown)


klihub (commented):
nitpick: the commit message is a bit iffy/out of context IMO. We're merely checking that we have sudo rights here if $VM_IP is set.

askervin:
Thanks, @klihub! Commit message rephrased. Also note the `return`s inside the "if VM__IP" branch, that cause skipping govm check.

klihub:
> Thanks, @klihub! Commit message rephrased. Also note the `return`s inside the "if VM__IP" branch, that cause skipping govm check.

Ah, true. Also the govm check is skipped.

klihub (approved):

PR #675, opened 2021-07-02
e2e: add vm-install-runc

askervin:
vm-install-runc supports installing runc from a package and from the
${runc_src} directory where it is created by "make".

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -607,6 +607,22 @@ vm-install-golang() {
     distro-install-golang
 }
 
+vm-install-runc() {
+    local host_runc="$runc_src/runc"
+    local vm_runc="/usr/sbin/runc"
+    if [ -n "$runc_src" ]; then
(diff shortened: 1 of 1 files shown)


klihub (commented):
nitpick: since related PRs enable running the tests on a real existing host as opposed to spinning up a throwaway test host using govm, maybe things that disciplined folks usually try to avoid (for instance overwriting binaries in system packages with quasi-random self-compiled binaries) should be put behind a 'allow-big-hammer' flag, which then should default to false.

askervin:
> 'allow-big-hammer' flag, which then should default to false.

What's your gut feeling, should this flag & default protect all hosts, or non-govm-managed only?

klihub:
> > 'allow-big-hammer' flag, which then should default to false.
> 
> What's your gut feeling, should this flag & default protect all hosts, or non-govm-managed only?

Only non-govm-managed hosts.

klihub (approved):

PR #677, opened 2021-07-02
e2e: support annotations in common pod templates

askervin:
Annotations can be added to besteffort, burstable and guaranteed pods
through ANN[0-9] environment variables. Example:
CPU=500m MEM=64M ANN0='my.key: "myvalue"' create guaranteed

### DIFF
--- a/test/e2e/besteffort.yaml.in
+++ b/test/e2e/besteffort.yaml.in
@@ -4,6 +4,15 @@ metadata:
   name: ${NAME}
   labels:
     app: ${NAME}
+    $(for lbl in ${!LABEL*}; do [[ "$lbl" == LABEL[0-9]* ]] && echo "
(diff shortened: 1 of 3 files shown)


klihub (commented):
LGTM. Maybe we should add support for labels in the same commit while we're at it...

askervin:
Good idea. Support for LABEL0=... LABEL1=.... included to this patch, commit message rephrased.

klihub (approved):

marquiz (commented):
Looks good. Just a small note about the naming. I'd prefer using a non-abbreviated name `ANNOTATION0` instead of `ANN0`. Especially as we now have `LABEL0` (and not `LAB0` for example).

I can live with the abbreviated naming too, if you want to merge this in it's current form, though 😄

Issue #686, opened 2021-07-09
e2e-test / openSUSE / amazon instances: missing package hub repositories


klihub:
**Describe the bug**
On openSUSE on amazon instances the package hub repository is not enabled by default. This prevents the scripts from successfully bootstrapping a working cluster. Also, after manually enabling the repository and rerunning the scripts, the failed step of pulling in the necessary packages is not retried.

**Expected behavior**
Successful bootstrapping of a working K8s cluster on amazon instances running openSUSE.

**To Reproduce**
Run `vm=mytestvm distro=opensuse VM_IP=$mytestvm_ADDRESS VM_SSH_USER=ec2-user test/e2e/run.sh` to try bootstrapping a cluster and running some tests.

klihub:
Making sure the package hub repo is reported to fix this problem.

Additionally, it might be an idea worth considering to add a flag akin to `reinstall_crirm` & co and allow users forcing the bootstrapping process to be retried all the way from the beginning (pulling in the necessary packages). Now rerunning the script after a failure with package hub enabled does not retry that failed step.

PR #688, opened 2021-08-04
e2e: add support for distro=sles

askervin:
Fixes AWS SLES VM support as required in issue #686.

Also enables running SLES on local VMs if the user has downloaded SLES cloud image (requires registration).


askervin:
Implements better Amazon support as requested in issue #686.

Also enables running SLES on local VMs if the user has downloaded SLES OpenStack VM image (requires registration).

klihub (approved):
Looks good to me.

PR #689, opened 2021-08-05
e2e: enable reinstalling pretty much everything on VMs

askervin:
- If a reinstall variable is set for any package, it will be forced to reinstall on the system.
- OpenSUSE / SLES install runc from Virtualization_containers.
- Reinstalling k8s means scrapping and recreating the cluster on VM -- in addition to reinstalling kubeadm, kubectl and kubelet.
- Refactor: get rid of screen-install-k8s.

Implemented on top of #688.

Implements suggested reinstall feature in issue #686 .


klihub (approved):
LGTM

PR #690, opened 2021-08-09
e2e: distro=sles uses official package repositories

askervin:


askervin:
@klihub, tested now with local and aws SLES vms. ready for review.

demo/lib/distro.bash
```diff
@@ -192,7 +195,18 @@ debian-install-pkg() {
...
+
+    # Add apt-get option "--reinstall" if any environment variable
+    # reinstall_<pkg>=1
+    local pkg
+    local opts=""
+    for pkg in "$@"; do
+        if [ "$(eval echo \$reinstall_$pkg)" == "1" ]; then
+            opts="$opts --reinstall"
+            break
+        fi
```
klihub, line 207: nitpick: would this fail sometimes (or work incorrectly) if you'd unconditionally always pass `--reinstall` as an option ?

askervin: The purpose is to protect whatever pkg/binary (like crio, containerd, runc) happens to be present in the vm unless reinstall_<pkg> is explicitly requested.

For instance, the binary may be have been copied to the vm from host if `crio_src`, `containerd_src` or `runc_src` has been set when the binary was (re)installed last time.  If user is not asking for reinstall, then the script tries to keep it that way.

klihub (approved):

PR #692, opened 2021-08-16
Expose the cpu usage of pools to Promethues.

huangrui666:
1. Implement the introspect function of podpools-policy.
2. Caculate cpu utilization of cores and pools by using
the statistics from /proc/stat.
3. Expose the cpu usage metrics to Prometheus.

Signed-off-by: Huang Rui <rui1.huang@intel.com>


klihub:
I have a few quick comments after a first skim-through of the PR/code.

So far we tried to keep the introspection interface very generic. For instance, the only (placement) data we exposed about containers were related to hardware (NUMA nodes, CPU cores), not to any internal software abstractions (for instance pools). One of your additions (to the introspection abstraction) is to associate a container with a pool. This brings in a related, fairly strong implicit assumption that all policies are pool-based. There is nothing wrong with that assumption per se, as long as it is a conscious explicit choice that we are able and willing to make and not just an accidental side-effect (of exposing policy data to Prometheus by first pulling it through introspection).

Pulling metrics data through introspection introduces an implicit dependency of the Prometheus exporter on the enabled state of the introspection interface (which is a built-in HTTP server and for security reasons off by default). Whereas these two were orthogonal before, after this change if you want to expose (podpools) policy data to Prometheus you need to enable the introspection interface. An alternative to this would be to separate the two/keep the two separated and add a metrics-pulling interface to the policy API, similar to the introspection one but using whatever data/format the policy wants to expose instead of trying to impose a generic data format on all policies.

huangrui666:
hi @klihub , thank you for your modification suggestions. After read your comment, I realized that I misunderstood the usage of the introspection. As you suggested, a new API MetricsPulling has been added into the pkg/cri/resource-manager/policy/policy.go which is similar to the introspection one but different kinds of policies can define its own policy metrics for monitoring. I think this is a better design. 
My implementation method is to define a struct PolicyMetrics in pkg/metrics/metrics.go that has Policy and Data fields. The filed Policy is used to keep the policy name and the filed Data is used to keep the policy metrices string from different policies.  For now it only exposes the metrics of podpools-policy which I am mainly working on. And the MetricsPulling API has been left to other policies, which could be fulfilled in the future by anyone else.

pkg/cri/resource-manager/metrics/metrics.go
```diff
@@ -193,3 +196,16 @@ func dump(prefix string, f *model.MetricFamily) {
 func metricsError(format string, args ...interface{}) error {
 	return fmt.Errorf("metrics: "+format, args...)
 }
+
+// set the updated policy metrics.
+func (m *Metrics) Set(pm *metrics.PolicyMetrics) {
+	m.Lock()
+	defer m.Unlock()
+	m.policyMetrics = pm
+	data, err := json.Marshal(m.policyMetrics)
```
klihub, line 205: nitpick: I think there is no reason to do the heavy lifting of `json.Marshal()` with that lock held. You could first marshal, then take the lock only for updating the cached data.

pkg/cri/resource-manager/metrics/metrics.go
```diff
@@ -193,3 +196,16 @@ func dump(prefix string, f *model.MetricFamily) {
...
+	if err != nil {
+		err = metricsError("failed to marshal policy metrics for monitor: %v", err)
+		return
+	}
+	metrics.Set(string(data))
```
klihub, line 210: nitpick: if you'd store the (locally) opaque policy metrics as `[]byte` instead of a `string` you could avoid the double-conversion to string here and later from string in the collector.

pkg/cri/resource-manager/policy/policy.go
```diff
@@ -127,6 +129,8 @@ type Backend interface {
 	ExportResourceData(cache.Container) map[string]string
 	// Introspect provides data for external introspection.
 	Introspect(*introspect.State)
+	// MetricsPulling provides policy metrics for monitoring.
+	MetricsPulling(*metrics.PolicyMetrics)
```
klihub, line 133: nitpick: I think it would sound more natural to call this `PullMetrics()` or `PollMetrics()`.

pkg/metrics/metrics.go
```diff
@@ -11,11 +12,26 @@ var (
 	registeredCollectors  = []prometheus.Collector{}
 	initializedCollectors = make(map[string]struct{})
 	log                   = logger.NewLogger("collectors")
+	policyMetrics         string
```
klihub, line 15: nitpick: regarding my earlier comment about avoiding double-conversion, this could be stored as `[]byte` instead.

klihub:
> a new API MetricsPulling has been added into the pkg/cri/resource-manager/policy/policy.go which is similar to the introspection one but different kinds of policies can define its own policy metrics for monitoring. I think this is a better design.

Thanks for the updated PR. I also think it is better that way.

There is one larger/structural thing that still bothers me with the updated PR: the implementation adds an explicit dependency from the collector to the actual policy implementations. As far as I can tell, the only reason for this is so `updatePoolCpuUsageMetric()` can emit pool-based prometheus metrics for the `podpools` policy.

While this dependency is currently limited to `podpools`, this is only because `podpools` is the only policy that provides real metrics data. If we implemented the metrics data interface for other policies, we'd end up adding similar dependencies on all of them, similar policy-specific branches in that `switch` for all of them, and policy-specific metrics emission code for all of them.

It looks to me that the collector merely acts here as a postmen between the policy and prometheus. It pulls cached metrics data from/for the active policy, converts it to pool-based metrics and pushes that to prometheus. Therefore, I don't see why the code/interfaces/abstractions couldn't be restructured so, that all the policy-specific code would reside in the policy itself. The rest of the infrastructure would only take care of declaring the metrics, pulling, and caching the metrics data, and calling in to the policy to generate the policy-specific metric bits from cached data, if and when necessary.

What do you think ? @askervin, any opinion about this ?

For example, wouldn't it be quite a natural spit of duties to try adding something like this to the `Policy` interface and the specific policy implementations:
```
type Policy interface {
    ...
    // DescribeMetrics generates policy-specific prometheus metrics data descriptors. `nil` implies no policy-specific metrics.
    DescribeMetrics() []*prometheus.Desc
    // PollMetrics polls policy-specific metrics data.
    PollMetrics() Metrics
    // CollectMetrics generates prometheus metrics from cached/polled policys-specific metrics data.
    CollectMetrics(Metrics) ([]prometheus.Metric, error)
}

// Metrics is (opaque) policy-specific metrics data stored in a policy-specific format.
// There are good and bad points with such a generic Metrics type. It avoids any
// need for an internal data conversion between the pulling/caching/emitting phases.
// But the data can't be interpreted outside of the policies. As long as the data can
// be treated internally as truly opaque (no other parts than the policy itself needs to
// understand it) this should be find.
//
// If we can come up with a working common format other just an interface, that could
// be better. However that format should not assume too much about the internals/logic
// of the policy, so it's difficult to see what else could be generic enough but less opaque.
//
type Metrics interface{}
```

The current code then would need to be reworked accordingly, moving the podpools-specific metrics description and generation code to `podpools`, changing the metrics, collector etc, bits so that they get the necessary descriptors and data to emit using the generic policy interfaces also correctly handling policies that do not generate metrics (provide a `nil` descriptor slice). There might be a few other tricky bits, for instance it might be necessary to delay the metrics/collector initialization until the active policy is known, and other similar things...

askervin:
@huangrui666, this is a good improvement. I agree with @klihub, that it will be further improved by putting all podpools-specific (metrics) code inside the `podpools` package. Calling active policy's new functions for polling/collecting metrics is a nice way to get rid of `switch-case` of policies. That will keep implementing new policies as simple as possible. Policy interface extension suggested by @klihub looks good to me.

huangrui666:
hi @klihub  , I tried to add the APIs as you suggested. But it seems to be hard to implement the API 
`CollectMetrics(Metrics) ([]prometheus.Metric, error)` in the policy module.  From my understanding, `resource-manager/policy` 
can only update its metrics data to `resource-manager/metrics` by resource-manager pushing. For example:

> // updatePolicyMetrics pushes updated policy metrics for monitoring·
> func (m *resmgr) updatePolicyMetrics() {
>       m.metrics.SetPolicyMetrics(m.policy.PollMetrics())
> }

Therefore, it’s able to cache the updated metrics whenever containers status change and set the policy descriptors easily.
However, if it registers a new collector which is used to collect the policy-specific metrics in `pkg/policyMetrics/collector.go`:

> err := metrics.RegisterCollector("policyMetrics", NewCollector)

The collector will generate prometheus data and calculate the result in API `Collect` at every metrics intervals. While it implements the `CollectMetrics(Metrics) ([]prometheus.Metric, error)` in policy module, the only solution I can think of is to update the collected Prometheus metrics to pkg/metrics at every metrics intervals as well… This implementation will need resource manager to invoke and call periodically. I don't think this is a good solution. Therefore my thoughts have been stuck. Have I missed something? Any suggestions from you will be appreciated! Thanks a lot.

huangrui666:
What if adding a new API 'NewPolicyCollector()' in `policy.go`. Each policy implements its own collector.
> // Return policy specific collector. 
>  NewPolicyCollector() (prometheus.Collector, error)

And register the policy-specific collector to metrics when a policy has been created:

> err := metrics.RegisterCollector("policyMetrics", NewPolicyCollector())

PR #694, opened 2021-08-18
scripts: add CRI-O support to kube-cgroups

askervin:
kube-cgroups prints pods/containers with their cgroups configuration.
Before this change it found the configurations only for containers
managed by containerd. This change makes it work similarly with CRI-O
containers, too.

This change is needed for running
test/e2e$ k8scri=crio ./run_tests.sh blockio.test-suite


klihub (approved):
LGTM.

PR #696, opened 2021-08-23
e2e: wait for cloud-init to finish during VM bootstrap.

klihub:
Currently the test framework does not wait for `cloud-init` to finish when bootstrapping a VM. This can (and often does) break bootstrapping any distro that requires triggering a reboot from within `cloud-init`. One such distro is fedora, where we change cgroupfs mount options.

The breakage happens if
- `host-wait-vm-ssh-server` gets into the VM on the first boot after ssh has started but before `cloud-init` has rebooted, and
- the first subsequent `vm-command` runs after `cloud-init` has triggered a reboot but before the VM is up again

In such cases `vm-command` aborts with a (rather misleading) error about command failure. The fix is to wait for `cloud-init` to finish once the ssh server on the host is up.

### DIFF
--- a/demo/lib/host.bash
+++ b/demo/lib/host.bash
@@ -199,6 +199,7 @@ host-create-vm() {
     echo "# VM Qemu monitor: docker exec -it $VM_CONTAINER_ID nc local:/data/monitor"
     VM_MONITOR="docker exec -i $VM_CONTAINER_ID nc local:/data/monitor"
     host-wait-vm-ssh-server
+    host-wait-cloud-init
 }
 
 host-wait-vm-ssh-server() {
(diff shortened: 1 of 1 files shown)


demo/lib/host.bash
```diff
@@ -234,6 +235,23 @@ host-wait-vm-ssh-server() {
     [ "$retries" == "$retries_left" ] || echo ""
 }
 
+host-wait-cloud-init() {
+    retries=60
+    retries_left=$retries
+    while ! $SSH ${VM_SSH_USER}@${VM_IP} -o ConnectTimeout=2 cloud-init status --wait 2>/dev/null; do
```
askervin, line 241: Fix: on `distro=centos-7` running `cloud-init status` as a user fails to a permission issue. I'd suggest adding `sudo`. That works on all distros.

Nitpick: suggestion on style: maybe ConnectTimeout could be more clearly a SSH parameter.
```
$SSH ${VM_SSH_USER}@${VM_IP} -o ConnectTimeout=2 COMMAND
->
$SSH -o ConnectTimeout=2 ${VM_SSH_USER}@${VM_IP} COMMAND
```

askervin (commented):
This turns out to be more complicated than what I expected...

`centos-7` issue can be fixed easily (by adding `sudo`), but this will break `distro=opensuse`. Its cloud-init hangs on running zypper, possibly due to bad proxy settings at that point.

demo/lib/host.bash
klihub: Updated according to those suggestions.

askervin (approved):
LGTM.

PR #697, opened 2021-08-23
e2e: allow installing cri-o from distro repos.

klihub:
Don't insist installing cri-o from sources. Add support for
installing cri-o from distro repos for Fedora and CentOS-8.


askervin (approved):
LGTM. Works fine. Thanks!

PR #699, opened 2021-08-24
e2e: follow HTTP redirects when fetching apt repo keys.

klihub:
Simply fetch keys with `curl -L` instead of plain `curl`.


askervin (approved):
LGTM. Works fine.

Note (to self, at least): there's a warning at least on `distro=debian-sid`:
```
root@vm> curl -L -s https://download.docker.com/linux/debian/gpg | apt-key add -
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
```

...but that's not a problem of this patch.

PR #702, opened 2021-08-30
Monitor the pool CPU usage for podpools-policy.

huangrui666:
1. Add the API PollMetrics in Policy and Backend interface.
It aims to define the diverse metrics for different
policies and expose these policy metrics for monitoring.
2. Add the API NewPolicyCollector() in Policy and Backend
interface. It aims to generate policy-specific prometheus
collector for different policies. If a policy has defined
its own collector, register it to metrics when the new
policy has been created.
2. Calculate CPU utilization of cores and pools by using
the statistics from /proc/stat for podpools policy. And
expose the CPU usage metrics to Prometheus.

Signed-off-by: Huang Rui <rui1.huang@intel.com>

### DIFF
--- a/scripts/testing/kube-cgroups
+++ b/scripts/testing/kube-cgroups
@@ -1,0 +1,137 @@
+#!/bin/bash
+
+usage() {
+    cat <<EOF
+Usage: kube-cgroups [options]
+
+Options:
+  -g CGDIR            print cgroup data under CGDIR.
+                      The default is /sys/fs/cgroup.
+  -E                  print also empty files.
+                      The default is to print non-empty files only.
+  -F                  print full cgroup filename.
+                      The default is basename only.
+  Filtering options:
+  -n NS_REGEXP        print only pods in namespaces matching NS_REGEXP
+  -p POD_REGEXP       print only pods matching POD_REGEXP
+  -c CNTR_REGEXP      print only containers matching CNTR_REGEXP
+  -f CGFILE_REGEXP    print only cgroup files matching CGFILE_REGEXP
+
+Examples:
+
+  # print cgroup information of pods in any namespace
+  kube-cgroups -n .
+
+  # print read bps and iops throttling of containers in mypod
+  kube-cgroups -g /sys/fs/cgroup/blkio -p mypod -f read
+EOF
+}
+
+error() {
+    echo "kube-cgroups: $*" >&2
+    exit 1
+}
+
+full_filename=0
+empty_files=0
+ns_regexp="default" # regexp matching
+pod_regexp="." # regexp matching any pod name
+cntr_regexp="." # regexp matching any container line
+cgfile_regexp="cpuset.cpus|cpuset.mems|blkio.throttle.*_device" # regexp matching any cgroup file
+
+cg_controller_dir=/sys/fs/cgroup
+
+while getopts "hg:EFn:p:c:f:" OPTION; do
+    case $OPTION in
+        h)
+            usage
+            exit 0
+            ;;
+        g)
+            cg_controller_dir="$OPTARG"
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/requests.go
```diff
@@ -236,6 +236,7 @@ func (m *resmgr) RunPod(ctx context.Context, method string, request interface{},
 
 	pod := m.cache.InsertPod(podID, request, nil)
 	m.updateIntrospection()
+	m.updatePolicyMetrics()
```
klihub, line 239: In my opinion, it is not a good idea to trigger (and wait) for a full collection cycle with every single CRI operation we intercept.

Granted we do the same for introspection data generation, but there is a huge difference. That one is a simple conversion of existing data already present in the policy to a canonical format. It does not need to reach out and collect an unknown amount of data from external sources of unknown latency. The `podpools` policy collector here will loop trough a bunch of cgroups and other data in `procfs`. We wait for that synchronously here and that adds that much latency to the processing of every request.

This comes to another point with the original design sketch up. Although it was not explicitly state there, our idea was to decouple policy-specific/`podpools` metrics collection from polling. IOW, to use an independent asynchronous goroutine within the (`podpools`) policy to periodically collect metrics with a configurable frequency/sampling rate and cache it internally. Then `podpools.PollMetrics()` would simply (read-lock the podpools policy) and return the already collected metrics data.

This would also have the additional benefit that your metrics sampling rate would be independent of the prometheus collection/sampling rate which would be a very good thing. Otherwise if you try to collect some more complex statistics you'll miss things as the externally clocked/triggered (relatively seldom) collection sampling rate will act as a low-pass filter on your data. As a stupid artificial example,  for instance if you tried to collect info about the CPU-'burstiness' of containers in a pool, then with a 10 second externally triggered collection you miss containers that steadily alternate between 1 second of extreme load and 1 second of idleness.

klihub:
Thanks for the updated PR. Much better in my opinion. We're progressing nicely towards a first mergable version, although we're not quite there yet, IMO.

@askervin, @marquiz Want to chime in with additional comments ?

pkg/cri/resource-manager/policy/builtin/podpools/collector.go
```diff
@@ -0,0 +1,331 @@
...
+type Metrics struct {
+	PoolMetrics map[string]*PoolMetrics
+}
+
+// PoolMetrics defines the metrics from pool level.
+type PoolMetrics struct {
+	DefName        string
+	PrettyName     string
+	CPUs           string
+	CPUMiliSize    string
```
askervin, line 42: Typo: `s/Mili/Milli/g`

pkg/cri/resource-manager/policy/builtin/podpools/collector.go
```diff
@@ -0,0 +1,331 @@
...
+			"CPUs",
+			"memory",
+			"pool_size",
+			"pod_name",
+			"container_name",
+		}, nil,
+	),
+}
+
+// core time statistic is used to calculate the CPU usage.
```
askervin, line 79: I'd suggest following the godoc convention here, too: "...begins with the name of the element it describes".

```
// coreTimeStat stores stats and usage per CPU core.
```

askervin (commented):
Minor notes to the code added.

I do share @klihub's concern about CPU usage data update interval. Do you think it could be updated only when gathered?  That is, data refresh rate would be completely controlled by metrics gathering refresh rate?

If this doesn't sound good, then we should probably launch a data gatherer worker goroutine, but not unless it is really needed. That is, the data is really polled. The problem would be that we probably do not know for sure when it is no more needed and could be stopped.

PR #703, opened 2021-09-01
e2e: support running tests with CRI-O and cri-resmgr in NRI mode

askervin:
Add support for k8scri='crio&cri-resmgr'. In this CRI pipe kubelet
will connect to CRI-O, and cri-resmgr runs as an NRI plugin.

Example:
crio_src=/path/to/CRI-O-with-NRI-support \
crirm_src=/path/to/cri-resource-manager-with-NRI-support \
k8scri='crio&cri-resmgr' \
./run_tests.sh policies.test-suite/topology-aware


klihub (approved):
LGTM

PR #704, opened 2021-09-02
e2e: test container exit

askervin:


klihub (approved):
If that 'killing the parent' works as expected/described then this LGTM.

PR #707, opened 2021-09-14
e2e: fix sysctl settings that break cilium CNI on Fedora

askervin:
Tested to work on `distro=fedora` and `distro=centos-8`.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -744,6 +744,12 @@ touch /etc/sysctl.d/k8s.conf
 echo "net.bridge.bridge-nf-call-ip6tables = 1" >> /etc/sysctl.d/k8s.conf
 echo "net.bridge.bridge-nf-call-iptables = 1" >> /etc/sysctl.d/k8s.conf
 echo "net.ipv4.ip_forward = 1" >> /etc/sysctl.d/k8s.conf
+
+# rp_filter (partially) mitigates DDOS attacks with spoofed IP addresses
+# by dropping packages with non-routable (unanswerable) source addresses.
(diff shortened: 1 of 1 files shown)


jukkar (dismissed):

demo/lib/distro.bash
```diff
@@ -744,6 +744,7 @@ touch /etc/sysctl.d/k8s.conf
 echo "net.bridge.bridge-nf-call-ip6tables = 1" >> /etc/sysctl.d/k8s.conf
 echo "net.bridge.bridge-nf-call-iptables = 1" >> /etc/sysctl.d/k8s.conf
 echo "net.ipv4.ip_forward = 1" >> /etc/sysctl.d/k8s.conf
+echo "net.ipv4.conf.*.rp_filter = 0" >> /etc/sysctl.d/k8s.conf
```
marquiz, line 747: Even narrower, limited to lxc interfaces should be enough:
```
net.ipv4.conf.lxc*.rp_filter = 0
```

marquiz: @askervin: what do you think about this? In our use case only the lxc interfaces need this. I'm fine with the little bit bigger hammer, too

marquiz (approved):
LGTM

demo/lib/distro.bash
askervin: @marquiz , in this case bigger hammer feels better to me. I'm more concerned on hard-to-debug `rp_filter`ing issues on various CNI and container runtime implementations than malicious IP spoofing in our e2e VMs.

(Even if this is now documented and spelled out loud, someone will anyway copy-paste the setup routine to a life-critical cluster setup that will then be DDOSsed, but that's how things go because of  Murphy's law...)

PR #711, opened 2021-09-22
Integrate with goresctrl

jukkar:
~This depends on the changes in https://github.com/intel/goresctrl/pull/75~


jukkar:
* fixed gofmt errors
* removed remaining sst files

jukkar:
* Fixed various linter errors
* Use goresctrl v0.2.0

pkg/sysfs/utils.go
```diff
@@ -101,13 +103,13 @@ func writeSysfsEntry(base, entry string, val, oldp interface{}, args ...interfac
 	path := filepath.Join(base, entry)
 
 	switch val.(type) {
-	case string, ID, int, uint, int8, uint8, int16, uint16, int32, uint32, int64, uint64:
```
klihub, line 104: Hmm, don't we need to handle an `idset.ID` here to keep the code functionally identical ?

pkg/sysfs/utils.go
```diff
@@ -152,15 +154,13 @@ func parseValue(str string, value interface{}) error {
 	case *string:
 		*value.(*string) = str
 
-	case *ID, *int, *int8, *int16, *int32, *int64:
+	case *int, *int8, *int16, *int32, *int64:
```
klihub, line 157: Ditto here... no need to handle an `idset.ID` for fully identical behavior ?

pkg/sysfs/utils.go
```diff
@@ -152,15 +154,13 @@ func parseValue(str string, value interface{}) error {
...
 		v, err := strconv.ParseInt(str, 0, 0)
 		if err != nil {
 			return fmt.Errorf("invalid entry '%s': %v", str, err)
 		}
 
 		switch value.(type) {
-		case *ID:
```
klihub, line 162: ditto

pkg/sysfs/utils.go
```diff
@@ -324,7 +324,7 @@ func formatValue(value interface{}) (string, error) {
 	switch value.(type) {
 	case string:
 		return value.(string), nil
-	case ID, int, uint, int8, uint8, int16, uint16, int32, uint32, int64, uint64:
```
klihub, line 327: ditto

klihub (commented):
I have a few spotted I few differences I don't understand, wrt. fully identical code before vs. after these changes. Apart from those/once I understand them, this LGTM.

pkg/sysfs/utils.go
jukkar: Note that commit https://github.com/intel/goresctrl/commit/7d764492dc971b5704626dc50527ebfa6ef8cb38 in goresctrl changed the type of ID
```
-type ID int
+type ID = int
```
so ID is now int which means that there cannot be duplicate int type checks various switches.

marquiz: Yeah, I defined ID that way in goresctrl in order to reduce the need for pointless (IMO) type casts

klihub: Ah, ok. I missed that completely.

klihub (approved):
LGTM.

pkg/sysfs/system.go
```diff
@@ -962,26 +964,28 @@ func (sys *system) discoverPackages() error {
...
-			for i, id := range ids {
-				sys.cpus[id].sstClos = clos[i]
+
+			for _, id := range ids {
+				clos, err := sst.GetCPUClosID(id)
+				if err != nil {
+					return fmt.Errorf("failed to get SST-CP clos ids for package %d: %v", pkg.id, err)
+				}
+
+				sys.cpus[id].sstClos = clos
```
jukkar, line 988: @klihub / @marquiz can you check that this looks ok to you, the Clos get function was changed when code was moved to goresctrl so I had to change this part a bit more

pkg/sysfs/system.go
```diff
@@ -962,26 +964,28 @@ func (sys *system) discoverPackages() error {
...
-			if err != nil {
-				return fmt.Errorf("failed to get SST-CP clos ids for package %d: %v", pkg.id, err)
-			}
-			for i, id := range ids {
-				sys.cpus[id].sstClos = clos[i]
+
+			for _, id := range ids {
+				clos, err := sst.GetCPUClosID(id)
+				if err != nil {
+					return fmt.Errorf("failed to get SST-CP clos ids for package %d: %v", pkg.id, err)
```
marquiz, line 985: nit:
```suggestion
					return fmt.Errorf("failed to get SST-CP clos id for cpu %d: %v", id, err)
```

pkg/sysfs/system.go
marquiz: Looks good 👍

marquiz (changes requested):
Quite a diffstat 😅  One small suggestion, otherwise looks good.

Also, I'd suggest to squash the changes to `go.{mod,sum}` into one single commit.

jukkar:
Updated according to comments.

marquiz (approved):
lgtm

PR #716, opened 2021-10-01
e2e: add distro-install-kernel-dev for distro=opensuse

askervin:
distro-install-kernel-dev installs sources and build tools for
distribution-specific kernels.


klihub (approved):
LGTM

PR #719, opened 2021-10-03
distro: force non-interactive 'apt-get install'.

klihub:
Explicitly force apt-get (install) into non-interactive mode.
This should fix bootstrapping a cluster on debian sid.


marquiz (approved):
lgtm

PR #726, opened 2021-10-04
0.6.0 release preparation: use latest fedora image for cross-build.

klihub:
dockerfiles: use fedora:latest for cross-builds.

### DIFF
--- a/dockerfiles/cross-build/Dockerfile.fedora
+++ b/dockerfiles/cross-build/Dockerfile.fedora
@@ -1,5 +1,5 @@
 # pull in base + a minimal set of useful packages
-FROM fedora as fedora-build
+FROM fedora:latest as fedora-build
 
 ARG GOLANG_VERSION=1.16.8
 ARG GOLANG_URLDIR=https://dl.google.com/go


askervin:
Just to understand consequences... is it possible that this change results in a dependency to so new libc version that prevents running cri-resmgr on some Fedora versions that would have been supported otherwise?

marquiz (approved):
lgtm

marquiz:
> Just to understand consequences... is it possible that this change results in a dependency to so new libc version that prevents running cri-resmgr on some Fedora versions that would have been supported otherwise?

Isn't `fedora` and `fedora:latest` basically the same thing? 🧐

klihub:
> > Just to understand consequences... is it possible that this change results in a dependency to so new libc version that prevents running cri-resmgr on some Fedora versions that would have been supported otherwise?
> 
> Isn't `fedora` and `fedora:latest` basically the same thing? monocle_face

Beats me... I thought it should be, but apparently not always. Without adding that explicit `latest` tag, it was always building for me with `fedora:32`, which was the image I had readily available on my machine. That is why I added the explicit tag.

askervin (approved):
LGTM

klihub:
> Just to understand consequences... is it possible that this change results in a dependency to so new libc version that prevents running cri-resmgr on some Fedora versions that would have been supported otherwise?

Yes, of course that is possible. This is for cross-building against the latest released stable fedora cloud image, not any fixed version (hence the plain `fedora` suffix). If we wanted to build for a particular fixed version, we'd need to add versioned cross-build and other related collateral for that.

marquiz (approved):
> Beats me... I thought it should be, but apparently not always. Without adding that explicit latest tag, it was always building for me with fedora:32,

Ah, yes it behaves that way
lgtm

PR #730, opened 2021-10-05
0.6.0 release preparation: always try to enable 'SystemdCgroup = true' for tests with containerd.

klihub:
distro: always try setting 'SystemdCgroup = true' for containerd.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -892,9 +892,17 @@ default-install-containerd() {
 }
 
 default-config-containerd() {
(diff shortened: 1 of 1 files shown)


marquiz (approved):
lgtm, if it works.

In retrospect we really should've used ansible or smth for these 😇

klihub:
> lgtm, if it works.
> 
> In retrospect we really should've used ansible or smth for these

Exactly what I said to you guys a month or so ago during one of our office days. So I can't claim to disagree...

PR #732, opened 2021-10-11
draft: bin/memtierd: proof-of-concept memory moving daemon

askervin:
This patch introduces a memory moving daemon. The lower levels of the daemon are based on the topology aware memory demoter with few optimizations.


askervin:
Abandoning this draft PR. memtierd can be found in https://github.com/intel/memtierd

PR #734, opened 2021-11-17
e2e: Enable cri-o in centos-8

jukkar:
If user has not specified cri-o sources, then system specified
cri-o package is used but it is not enabled by default in centos-8.
This means that in second boot of the VM, the kube environment does
not start properly because cri-o is not started automatically.
Fix this by enabling cri-o in systemd so that it is started at boot.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -328,6 +328,15 @@ centos-8-install-crio-pre() {
     fi
 }
 
+centos-8-install-crio() {
+    if [ -n "$crio_src" ]; then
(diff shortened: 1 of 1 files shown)


klihub (approved):
LGTM

askervin (approved):
Thanks @jukkar, LGTM!

klihub:
@jukkar Can you rebase this on top of latest master (1c0071c3daea544ce1e33b90ad0e5602450e1eb1) ? That would get us rid of the failing (doc) build test and allow merging.

PR #736, opened 2021-11-18
scripts/build/get-buildid: return unknown version

marquiz:
Otherwise you might get version numbers like
"failed to determine version from any source" in the Makefile which
breaks things.


klihub (approved):
LGTM

PR #737, opened 2021-11-19
e2e: Add support for Fedora 35 images

jukkar:
Update F33 to F35 when creating Fedora images for testing.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -413,13 +413,17 @@ fedora-install-golang() {
     from-tarball-install-golang
 }
 
-fedora-install-crio-pre() {
+fedora-install-crio-version() {
     distro-install-pkg runc conmon
     vm-command "ln -sf /usr/lib64/libdevmapper.so.1.02 /usr/lib64/libdevmapper.so.1.02.1" || true
 
     if [ -z "$crio_src" ]; then
-        vm-command "dnf -y module enable cri-o:${crio_version:-1.21}"
+        vm-command "dnf -y module enable cri-o:${crio_version:-$1}"
     fi
+}
+
+fedora-install-crio-pre() {
+    fedora-install-crio-version 1.21
 }
 
 fedora-install-crio() {
@@ -494,6 +498,14 @@ if grep -q NAME=Fedora /etc/os-release; then
 fi
 EOF
     fi
+}
+
+fedora-33-image-url() {
+    echo "https://download.fedoraproject.org/pub/fedora/linux/releases/33/Cloud/x86_64/images/Fedora-Cloud-Base-33-1.2.x86_64.qcow2"
+}
+
+fedora-33-install-crio-pre() {
+    fedora-install-crio-version 1.20
 }
 
 ###########################################################################
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -352,7 +352,7 @@ centos-install-golang() {
 }
 
 fedora-image-url() {
-    echo "https://download.fedoraproject.org/pub/fedora/linux/releases/33/Cloud/x86_64/images/Fedora-Cloud-Base-33-1.2.x86_64.qcow2"
+    echo "https://download.fedoraproject.org/pub/fedora/linux/releases/35/Cloud/x86_64/images/Fedora-Cloud-Base-35-1.2.x86_64.qcow2"
 }
 
 fedora-ssh-user() {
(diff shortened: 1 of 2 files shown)


jukkar:
I am making this a draft as this requires a bit more testing.

askervin (dismissed):
LGTM.
Works fine!

demo/lib/distro.bash
```diff
@@ -466,6 +466,9 @@ fedora-install-k8s() {
     local yumkey="https://packages.cloud.google.com/yum/doc/yum-key.gpg"
     local rpmkey="https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg"
 
+    vm-command "( lsmod | grep -q br_netfilter ) || { echo br_netfilter > /etc/modules-load.d/50-br_netfilter.conf; modprobe br_netfilter; }"
+    vm-command "echo 1 > /proc/sys/net/ipv4/ip_forward"
```
jukkar, line 470: I just noticed that we need to save the setting to a sysctl config file that gets loaded during boot in order this to work as expected. I will send a fixed version of this one.

klihub: I think this probably could/should be done with(in) fedora-(35-)specific `bootstrap-commands()`...

demo/lib/distro.bash
```diff
@@ -445,12 +447,27 @@ fedora-config-containerd-post() {
...
+    if vm-command "swapon --show | grep -q partition"; then
+	echo "Disabling swap"
+	vm-sed-file /etc/fstab '/^\/.*[[:space:]]swap[[:space:]].*$/d'
+	vm-command "systemctl --type swap"
+	vm-command "for swp in \`systemctl --type swap | awk '/\\.swap/ { print \$1 }'\`; do systemctl stop \"\$swp\"; systemctl mask \"\$swp\"; done"
+	vm-command "swapoff --all"
+	vm-reboot
+    fi
+}
+
```
klihub, line 462: nitpick: couldn't/shouldn't this be better done in `fedora-bootstrap-commands*()` which is always run on the first boot ?

jukkar: Indeed, the `default-bootstrap-commands()` even sets the ip_forward already. I am just not seeing it set in my test setup, need to investigate why the default func is not run.

demo/lib/distro.bash
jukkar: Indeed, the default-bootstrap-commands() even sets the ip_forward already. I am just not seeing it set in my test setup, need to investigate why the default func is not run.

demo/lib/distro.bash
jukkar: Yep, will do

jukkar:
I will rework the swap disabling so that it will be part of #741

jukkar:
Requires that #741 is merged first.

klihub (approved):
LGTM.

I just wonder if it made sense to keep F33 support in there for a while, too.

The reason I'm asking is that F34 broke backwards compatibility with glibc in F33. That incompatibility creeps into `cri-resmgr` as well, since it gets linked against glibc. Based hands-on experience I know that this can be a major PITA if your systems still run F33 (regardless of whether that is a good idea or not).

With a it of luck, this could be a simple as adding a `fedora-33-image-url()`, if the updated `cri-o` package version is also available for F33 (since `iproute-tc` looks to be the correct package providing `/usr/sbin/tc` in F33 as well).

jukkar:
> With a bit of luck, this could be a simple as adding a `fedora-33-image-url()`, provided that the updated `cri-o` package version is also available for F33 (since `iproute-tc` looks to be the correct package providing `/usr/sbin/tc` in F33 as well).

It wasn't as simple as that, but still quite simple. I will send a version with F33 support shortly.

jukkar:
This one contains support also for F33

demo/lib/distro.bash
```diff
@@ -494,6 +500,66 @@ EOF
...
+    fedora-config-containerd-post
+}
+
+fedora33-install-k8s() {
+    fedora-install-k8s $*
+}
+
+fedora33-bootstrap-commands-post() {
+    fedora-bootstrap-commands-post $*
+}
```
klihub, line 561: nitpick: I think we should name support for F33 `fedora-33` instead of `fedora33`. Then of these functions we could leave out those which are just tautologic fallbacks to the unversioned fedora ones, because `distro-resolve-fn()` has built-in support for that: it always looks for `distro-function()` next if `distro-function-XY()` is not defined.

klihub (commented):
Looks good. Only one comment related to existing built-in logic in `distro-resolve-fn()` which should allow us to leave out much of the hand-crafted tautologic `fedora33-*` -> `fedora-*` fallbacks, if we changed the `fedora33` name to `fedora-33` instead.

demo/lib/distro.bash
jukkar: > distro-resolve-fn() has built-in support for that: it always looks for distro-function() next if distro-X-function() is not defined.

Ah, I missed that. Indeed that change makes sense, I will send a new version.

askervin (approved):
LGTM

Checked that `iproute-tc` is available in Fedora 33, too. Works fine.

klihub (approved):
LGTM

PR #738, opened 2021-11-22
e2e: Do not try to use govm if user has specified the VM address

jukkar:
If user has specified VM_IP when starting run.sh, then there is
no need to try to resolve the IP address of the VM. This allows
the script to be used to configure non govm based hosts.


askervin (approved):
Looks good, works fine. Thanks @jukkar!

PR #739, opened 2021-11-23
broken/rfc: e2e: opensuse 15.3

askervin:
In order to work, this patch needs either of the following:
1. govm update that enables writing cloud-init iso image so that
   it has "NoCloud" cloud-init files. Currently govm adds only
   ConfigDrive cloud-init.
2. govm update that enables modifying qcow2 images before launching
   Qemu inside the govm docker container. Like:
     1. qemu-nbd --connect=/dev/nbd0 /PATH/TO/opensuse153.qcow2
     2. mount /dev/nbd0 /mnt/vm-vda
     3. run user image modifying script in /mnt/vm-vda
   (Would this be possible with reasonable security on host?)
   Having this in place, our new distro-modify-image() could
   add ConfigDrive to /etc/cloud/cloud.cfg datasource_list, and
   opensuse 15.3 NoCloud image would start working.
3. Find an opensuse 15.3 openstack build somewhere else.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -559,6 +559,10 @@ opensuse-image-url() {
     echo "https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.2/images/openSUSE-Leap-15.2-OpenStack.x86_64-0.0.4-Build8.25.qcow2"
 }
 
+opensuse-15_3-image-url() {
+    echo "https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.3/images/openSUSE-Leap-15.3.x86_64-1.0.1-NoCloud-Build2.14.qcow2"
+}
+
 opensuse-tumbleweed-image-url() {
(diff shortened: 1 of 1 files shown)


askervin:
@kad , @klihub , @obedmr, I'd like to hear your views on enabling opensuse 15.3 on govm. As far as I can see, opensuse does not provide anymore cloud-init ConfigDrive option by default on their qcow2 cloud image. If we can't get that back to opensuse builds, we need to do something in govm to support opensuse 15.3. The options that I could come up with are trying to support both ConfigDrive and NoCloud init files on the iso image generated by govm, or enabling qcow2 image modifications before launching Qemu on that image.

klihub:
I really don't know what to do with this. So what is the problem/goes wrong in OpenSUSE ? The trusted keys don't get installed, or something else ? Is this related to these [OpenSUSE](https://bugs.launchpad.net/cloud-init/+bug/1779139) [bugs](https://bugs.launchpad.net/cloud-init/+bug/1733226) ?

askervin:
Now OpenSUSE 15.3 OpenStack Cloud image is available. We can forget about this. Created PR #751.

PR #741, opened 2021-11-24
e2e: Fixes for non govm usage

jukkar:
These fix couple of issues when the k8s is installed to a non VM host.
* The proxies can be setup automatically if `setup_proxies=1` is set when invoking `run.sh`
* Setup pre-flight reqs like ip_forward setting in the host

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -472,8 +472,9 @@ EOF
         command-error "failed to enable kubelet"
 }
 
-fedora-bootstrap-commands-pre() {
+fedora-bootstrap-commands-post() {
     cat <<EOF
+reboot_needed=0
 mkdir -p /etc/sudoers.d
 echo 'Defaults !requiretty' > /etc/sudoers.d/10-norequiretty
 
@@ -487,11 +488,29 @@ EOF
 if grep -q NAME=Fedora /etc/os-release; then
     if ! grep -q systemd.unified_cgroup_hierarchy=0 /proc/cmdline; then
         sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
-        shutdown -r now
+        reboot_needed=1
     fi
 fi
 EOF
     fi
+
+    # Using swapoff is not enough as we also need to disable the swap from systemd
+    # and then reboot the VM
+    cat <<EOF
+if swapon --show | grep -q partition; then
+    sed -E -i '/^\\/.*[[:space:]]swap[[:space:]].*\$/d' /etc/fstab
+    systemctl --type swap
+    for swp in \`systemctl --type swap | awk '/\\.swap/ { print \$1 }'\`; do systemctl stop "\$swp"; systemctl mask "\$swp"; done
+    swapoff --all
+    reboot_needed=1
+fi
+EOF
+
+    cat <<EOF
+if [ "\$reboot_needed" == "1" ]; then
+   shutdown -r now
+fi
+EOF
 }
 
 ###########################################################################
@@ -806,7 +825,7 @@ rpm-refresh-pkg-db() {
 
 default-bootstrap-commands() {
     cat <<EOF
-touch /etc/modules-load.d/k8s.conf
+rm -f /etc/modules-load.d/k8s.conf; touch /etc/modules-load.d/k8s.conf
 modprobe bridge && echo bridge >> /etc/modules-load.d/k8s.conf || :
 modprobe nf-tables-bridge && echo nf-tables-bridge >> /etc/modules-load.d/k8s.conf || :
 modprobe br_netfilter && echo br_netfilter >> /etc/modules-load.d/k8s.conf || :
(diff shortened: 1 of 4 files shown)


demo/lib/vm.bash
```diff
@@ -30,6 +30,10 @@ vms:
         grep -E -v '^ *$'
 }
 
+vm-bootstrap-commands() {
```
askervin, line 33: Before this patch the `distro-bootstrap-commands` function only prints commands to be executed when bootstrapping the system. Usually these are executed by cloud-init. I hesitate a bit making it a dual-operation function: printing or executing.

Do you think we could keep it only a printing function? Maybe new `vm-bootstrap` could store the bootstrap commands printed by `distro-bootstrap-commands` on vm and runs them as root? I mean like:
```
vm-bootstrap() {
    distro-bootstrap-commands | vm-pipe-to-file "./e2e-bootstrap.sh"
    vm-command "sh ./e2e-bootstrap.sh"
}
```

(Sure this needs revisiting the bootstrap commands so that they would not add duplicate lines to the same .conf files every time executed.)

test/e2e/run.sh
```diff
@@ -1199,6 +1203,12 @@ host-get-vm-config "$vm" || host-set-vm-config "$vm" "$distro" "$cri"
 
 if [ -z "$VM_IP" ] || [ -z "$VM_SSH_USER" ]; then
     screen-create-vm
+else
```
askervin, line 1206: I'd like to avoid running bootstrap commands on every launch --- and would actually try to avoid running it automatically when using an existing (not created by run.sh) host unless explicitly requested.

Maybe we could add `vm-reinit()` that would run bootstrap commands and then setup proxies? (First bootstrap then proxies would be the order in which these commands are executed for new vm's created by run.sh, so we might be a bit more safe if we'd execute them in this order in reinit, too.)

An existing way to execute arbitrary test script (including calling `vm-setup-proxies` and/or `vm-reinit`) are the hooks. Like:
```
$ on_vm_online='vm-bootstrap; vm-setup-proxies' ./run.sh interactive
```

But I'm fine having a new variable in `run.sh help` to assist new user initializing an existing host.

demo/lib/vm.bash
jukkar: Sure, that can be done like it. I was not sure why those commands were not immediately executed so did it like that.

test/e2e/run.sh
jukkar: Sure, I am fine with that. The command line gets quite complex with different options and combinations, just wondering if user is able to memorize all options that would need to be set.

jukkar:
Updated according to comments. Now the target bootstrap + proxy setup is done if `reinstall_bootstrap=1` and `VM_IP` is set in command line.
Example usage:
```
distro=fedora k8scri=crio reinstall_bootstrap=1 VM_IP=target.host VM_SSH_USER=root ./run.sh interactive
```

demo/lib/vm.bash
```diff
@@ -30,6 +30,12 @@ vms:
         grep -E -v '^ *$'
 }
 
+vm-bootstrap() {
+    distro-bootstrap-commands | vm-pipe-to-file "./e2e-bootstrap.sh"
```
klihub, line 34: I'm not able to emulate all the shell levels involved in my head to follow/tell, so it's better if I ask: does `distro-bootstrap-commands` put in enough protection for the recently added code with shell variable evaluation, so that the variables get evaluated only in `e2e-bootstrap.sh` on the host and not before ?

demo/lib/distro.bash
```diff
@@ -487,11 +488,29 @@ EOF
...
 fi
 EOF
     fi
+
+    # Using swapoff is not enough as we also need to disable the swap from systemd
+    # and then reboot the VM
+    cat <<EOF
+if swapon --show | grep -q partition; then
+    sed -E -i '/^\\/.*[[:space:]]swap[[:space:]].*\$/d' /etc/fstab
+    systemctl --type swap
```
klihub, line 502: nitpick: what is the purpose of this `systemctl --type swap` here ?

klihub (commented):
I have two clarifying questions. Apart from those this LGTM.

demo/lib/distro.bash
jukkar: I added it there in order to see what the system thinks about the swap devices, so just a debugging aid if something goes wrong. Can be removed if you think that is not needed.

demo/lib/vm.bash
jukkar: Yes, the variables are evaluated only inside the e2e-bootstrap.sh script. But that just relies to the fact that the user escapes the relevant shell variables properly.

demo/lib/distro.bash
klihub: No need to remove it if you consider it useful. I just wanted to make sure it is not an accidental leftover.

demo/lib/vm.bash
klihub: Okay, so I guess that means that in the current (fedora) `distro-bootstrap-commands()` the 'user' had escaped the variables correctly... ;-)

klihub (dismissed):

jukkar:
Do not merge yet, I noticed one issue and need to do more testing.

askervin (approved):
Looks Great! Thanks @jukkar!

klihub (approved):
LGTM

PR #744, opened 2021-12-20
e2e: benchmark: Workaround for dbus error

jukkar:
Container creation in benchmark script could fail with this error:

	dbus: connection closed by user

Add a hackish workaround for this issue by restarting dbus and kubelet
before creating redis container.

### DIFF
--- a/test/e2e/benchmarks.test-suite/memtier_benchmark/n4c16/test01-memtier-stress-ng/code.var.sh
+++ b/test/e2e/benchmarks.test-suite/memtier_benchmark/n4c16/test01-memtier-stress-ng/code.var.sh
@@ -35,6 +35,14 @@ BM_MEMTIER="memtier-benchmark --server=redis-service --authenticate=$REDIS_PASS"
 # Clean up
 vm-command "kubectl delete jobs --all --now; kubectl delete deployment redis; kubectl delete service redis-service; kubectl delete secret redis; kubectl delete pods --all --now; true"
 
+# redis pod creation might fail with this error
+#    Warning FailedCreatePodContainer 5s (x4 over 45s) kubelet unable to ensure pod container exists: failed to create container for [kubepods burstable podfbd7f451-4b49-4deb-8b49-40ea97685afa] : error while starting unit "kubepods-burstable-podfbd7f451_4b49_4deb_8b49_40ea97685afa.slice" with properties [{Name:Description Value:"libcontainer container kubepods-burstable-podfbd7f451_4b49_4deb_8b49_40ea97685afa.slice"} {Name:Wants Value:["kubepods-burstable.slice"]} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:TasksAccounting Value:true} {Name:DefaultDependencies Value:false}]: dbus: connection closed by user
+# Currently only solution seems to be to restart dbus and kubelet so do it here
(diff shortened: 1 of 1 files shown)


klihub:
I'm not against merging this workaround per se. I'm just wondering if this is/will be the only (test) case that we're interested in running in an environment which exhibits such erratic behavior... If we suspect there might be others, then it would be better to move this workaround to the test infra itself so it could be easily reused in other test cases as well. At least as a single function that would check for that error and try to initiate the fixup...

Of course, the best would be if we could automatically detect that error and trigger the workaround, preferably automagically unbeknownst to any test case. We do have that `is-hooked()/run-hook()` scriptlet infra that remotely looks like it could be suitable for this (without my glasses on), but I'm not sure if beating that to obedience for this would really be possible with reasonable effort. Mostly due to the fact that many of the test cases go `vm-command kubectl ...`, which does not make hooking in anything very easy.

**If** could hook in a post-check somewhere somehow, then I think the detection and fixup attempt could be relatively easy with something as simple a puke as `vm-command-q 'journalctl -u kubelet --since "15 seconds ago" | grep -q "dbus: connection closed by user" && systemctl restart kubelet dbus'`...

jukkar:
Yes, the proposal is a dirty hack and I "fixed" the issue only in the benchmark script as it was the only place I encountered it. I investigated a way to detect the issue but vm-command will exit if the command fails so it requires a bit work to make usable.
I will check the hook thingy if we could utilize that.

klihub:
> Yes, the proposal is a dirty hack and I "fixed" the issue only in the benchmark script as it was the only place I encountered it. I investigated a way to detect the issue but vm-command will exit if the command fails so it requires a bit work to make usable. I will check the hook thingy if we could utilize that.

Putting my glasses one, that hook infra looks to be at the wrong layer, so at least that would need to be moved from `test/e2e/run.sh` down to `demo/lib/...`. But even then, I suspect that `vm-command` would need quite ugly plastic surgery to hide this logic there. It is way too generic, it can be run both synchronously and asynchronously, etc.

I think it would be way cleaner to provide a dedicated wrapper for running kubect on the VM/test host, maybe something like `vm-kubectl` (something that in hindsight we should have probably done in the first place) and then try injecting error detection/fixup there. But it's still not straightforward at all to do if we keep vm-kubectl a generic wrapper.... and I'm not sure if it is really worth the effort. Maybe the cost/benefit ratio is more acceptable if we simply try providing a combined error detection+workaround-triggering function and then those test cases can use it to deal with this cr@p as necessary.

jukkar:
After pondering this PR a bit more, I am actually not sure if we should apply it at all. I have seen it in one Fedora 34 host (although it was happening consistently) so it seems to be somewhat rare. If there are reports of this kind of issue in the future, then it should be easy to continue further. Opinions?

klihub:
> After pondering this PR a bit more, I am actually not sure if we should apply it at all. I have seen it in one Fedora 34 host (although it was happening consistently) so it seems to be somewhat rare. If there are reports of this kind of issue in the future, then it should be easy to continue further. Opinions?

I'm fine with that.

askervin:
I tried what happens on Fedora 34 on VM (PR #745). The dbus problem doesn't seem to manifest itself there. So postponing merging this makes sense - the other Fedora 34 environment might suffer from some other (hopefully rare) issue with dbus/kubelet.

jukkar:
Closing for now, can be revisited later if needed.

PR #746, opened 2021-12-27
e2e: add vm-install-glibc, vm-set-glibc and on_*_install hooks

askervin:
- Enable running a binary on VM, even if the binary would have been
  built with later glibc than available in the VM.
- Example: installing a locally built containerd to a distro with
  an old glibc:
  containerd_src=.../github.com/containerd/containerd \
  on_cri_install='vm-install-glibc 2.34
                  vm-set-glibc 2.34 /usr/bin/containerd
                  vm-command "systemctl restart containerd"'

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -92,6 +92,8 @@ usage() {
     echo "             mode if a verification fails: on_verify_fail=interactive"
     echo "    on_verify, on_create, on_launch: code to be executed every time"
     echo "             after verify/create/launch function"
+    echo "    on_{cri,runc,k8s}_install: code to be executed right after installing"
+    echo "             these components."
     echo ""
     echo "  VM configuration VARs: (effective when vm is not already configured)"
     echo "    topology: JSON to override NUMA node list used in tests."
@@ -1231,6 +1233,7 @@ fi
 
 if [ "$reinstall_containerd" == "1" ] || [ "$reinstall_crio" == "1" ] || ! vm-command-q "( type -p containerd || type -p crio ) >/dev/null"; then
     vm-install-cri
+    is-hooked on_cri_install && run-hook on_cri_install
 fi
 
 # runc is installed as a dependency of containerd and crio.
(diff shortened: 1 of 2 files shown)


askervin:
This patch helps testing the same (local) build of runc, crio, containerd and cri-resmgr on Linux distributions that have older glibc versions that the local build.

test/e2e/run.sh
```diff
@@ -1231,6 +1233,7 @@ fi
 
 if [ "$reinstall_containerd" == "1" ] || [ "$reinstall_crio" == "1" ] || ! vm-command-q "( type -p containerd || type -p crio ) >/dev/null"; then
     vm-install-cri
+    is-hooked "on_cri_install" && run-hook "on_cri_install"
```
klihub, line 1236: nitpick 1: why to quote `on_cri_install` here ?

nitpick 2: I've always wondered why isn't the `is-hooked` test folded into `run-hook` ? Then we could always unconditionally go `run-hook foo` instead of the now idiomatic/repetitive `is-hooked foo && run-hook foo`...

test/e2e/run.sh
```diff
@@ -1239,10 +1242,12 @@ fi
 # a custom locally built runc may be overridden from packages.
 if [ "$reinstall_runc" == "1" ] || ! vm-command-q "type -p runc >/dev/null"; then
     vm-install-runc
+    is-hooked "on_runc_install" && run-hook "on_runc_install"
```
klihub, line 1245: ditto...

test/e2e/run.sh
```diff
@@ -1239,10 +1242,12 @@ fi
...
 fi
 
 if [ "$reinstall_k8s" == "1" ] || ! vm-command-q "type -p kubelet >/dev/null"; then
     vm-install-k8s
+    is-hooked "on_k8s_install" && run-hook "on_k8s_install"
```
klihub, line 1250: ditto...

klihub (commented):
I'm fine with this, although I think if one is unfortunate enough to have to go this route, then it would be *significantly* faster to compile/install `glibc` on the host to a temporary location and just copy the whole shebang to the VM (which of course would assume `gcc` to be compatible between the host and the VM)...

I had a few real nitpicks, though...

test/e2e/run.sh
askervin: Quotations fixed. I agree with integrating is-hooked into run-hook. I do not remember anymore why-oh-why did I split it in two. It may have been a thought failure in thinking how to return correct exit status from a function.

I'd like to write another PR that removes calling is-hooked before run-hook.

test/e2e/run.sh
askervin: Fixed.

askervin:
Thanks @klihub! I added support for the `glibc_src` environment variable, quite similarly to `containerd_src`, `crio_src`, `runc_src` and `cri_resmgr_src`. If that is set, `vm-install-glibc` installs glibc to the vm from the host directory pointed by `glibc_src` . This is definitely is much better in my own setup, too! But I still kept the download-build-install-on-vm routine (when `glibc_src` is not set), too, as in case where a single-shot-solution is fine, it works without any extra dependencies on host.

klihub (approved):
LGTM

PR #747, opened 2021-12-27
e2e: support k8scni=<cilium|weavenet|flannel>

askervin:
Similarly to the `k8scri` parameter, allow user to define which CNI to install on new cluster by setting `k8scni`.

This patch also adds support to the flannel CNI.

Setting `reinstall_bootstrap` is added to keep our shellcheck (git hook) happy.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -2,7 +2,6 @@
 GO_URLDIR=https://golang.org/dl
 GO_VERSION=1.16.8
 GOLANG_URL=$GO_URLDIR/go$GO_VERSION.linux-amd64.tar.gz
-CNI_SUBNET=10.217.0.0/16
 
 ###########################################################################
 
@@ -39,6 +38,7 @@ distro-restart-crio()       { distro-resolve "$@"; }
 distro-install-k8s()        { distro-resolve "$@"; }
 distro-install-kernel-dev() { distro-resolve "$@"; }
 distro-k8s-cni()            { distro-resolve "$@"; }
+distro-k8s-cni-subnet()     { distro-resolve "$@"; }
 distro-set-kernel-cmdline() { distro-resolve "$@"; }
 distro-bootstrap-commands() { distro-resolve "$@"; }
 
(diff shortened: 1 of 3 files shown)


demo/lib/distro.bash
```diff
@@ -941,7 +941,15 @@ default-install-utils() {
...
 default-k8s-cni() {
-    echo cilium
+    echo ${k8scni:-cilium}
+}
+
+default-k8s-cni-subnet() {
+    if [ "$(distro-k8s-cni)" == "flannel" ]; then
+        echo 10.244.0.0/16
+    else
+        echo 10.217.0.0/16
```
klihub, line 951: Is this special-case required by `flannel` ?

test/e2e/run.sh
```diff
@@ -1059,6 +1061,7 @@ cri_resmgr_extra_args=${cri_resmgr_extra_args:-""}
 cri_resmgr_agent_extra_args=${cri_resmgr_agent_extra_args:-""}
 cleanup=${cleanup:-0}
 reinstall_all=${reinstall_all:-0}
+reinstall_bootstrap=${reinstall_bootstrap:-0}
```
klihub, line 1064: nitpick: I guess this is a(n orthogonal) fixup for the origin commit introducing the `reinstall_bootstrap` flag.

klihub (approved):
LGTM.

I just have a small question and a nitpick (that we can very well leave unaddressed, no need to split it out to a commit/PR of its own).

test/e2e/run.sh
askervin: Yeah, your guess is right.

Our shellcheck git hook bit me when I was trying to create a commit that modifies run.sh.  Fixing the shellcheck issue here was the least time consuming but the second ugliest choice. It definitely would have been nicer to disable the shellcheck hook for some time and create two PRs. And it would have been even more evil to leave this unfixed and let it bite someone again in the future. So this was a non-optimal compromise that won't make anyone very happy...

Encouraged by your comment of "no need to split", I could leave this as is and feel the sting I deserved. :)

PR #749, opened 2021-12-29
e2e: vm-install-cri replaces original containerd/crio binary

askervin:
Before this patch locally built containerd and crio were always
installed to /usr/bin on vm. This caused problems in distros where
packages install these runtimes somewhere else, for instance under
/sbin.

This patch verifies which containerd and crio binaries are used on the
vm and replaces those with local builds when calling vm-install-cri
with containerd_src or crio_src environment variables set.

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -699,24 +699,32 @@ vm-install-runc() {
 }
 
 vm-install-cri() {
+    local vm_cri_dir="/usr/bin"
     distro-install-"$VM_CRI"
     distro-config-"$VM_CRI"
     if [ "$VM_CRI" == "containerd" ]; then
         if [ -n "$containerd_src" ]; then
             vm-command "systemctl stop containerd"
+            vm-command 'command -v containerd'
+            if [ -n "$COMMAND_OUTPUT" ] && [ "x$COMMAND_STATUS" == "x0" ]; then
+                vm_cri_dir="${COMMAND_OUTPUT%/*}"
+            fi
             for f in ctr containerd containerd-stress containerd-shim containerd-shim-runc-v1 containerd-shim-runc-v2; do
-                vm-put-file "$containerd_src/bin/$f" "/usr/bin/$f"
+                vm-put-file "$containerd_src/bin/$f" "$vm_cri_dir/$f"
             done
-            vm-command "systemctl start containerd"
+            vm-command "systemctl enable --now containerd"
         fi
     elif [ "$VM_CRI" == "crio" ]; then
         if [ -n "$crio_src" ]; then
             vm-command "systemctl stop crio"
+            vm-command 'command -v crio'
+            if [ -n "$COMMAND_OUTPUT" ] && [ "x$COMMAND_STATUS" == "x0" ]; then
+                vm_cri_dir="${COMMAND_OUTPUT%/*}"
+            fi
             for f in crio crio-status pinns; do
-                vm-put-file "$crio_src/bin/$f" "/usr/bin/$f"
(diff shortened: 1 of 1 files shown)


klihub (commented):
I am fine with getting this merged...

That said, since since these fragments depend on the CRI runtime being managed by `systemd`, (and since the target machine can nowadays be other than a test VM), I wonder if it really would be that much more difficult to always put the runtime compiled on the host under `/usr/local/bin` then override the runtime's `systemd` configuration to look for the binary there.

I think (maybe) this could be fairly easily doable with something like:

```
echo "[Service]" | EDITOR="tee" systemctl edit $VM_CRI
xeq=$(systemctl cat $VM_CRI | grep ExecStart= | tail -1)
echo "ExecStart=" | EDITOR="tee -a" systemctl edit $VM_CRI
echo "$xeq" | sed "s#${vm_cri_dir}/#/usr/local/bin/#" | EDITOR="tee -a" systemctl edit $VM_CRI
```

Or something to this effect if it is impossible to take out enough artificial stupidity from `systemctl edit` by overriding `EDITOR`.

askervin:
I would gladly start installing locally built binaries into /usr/local/bin on vm, if only the daemon binaries would not have any dependencies to executables. However, as this is not the case, I'm afraid that installing another set of binaries would result in running wrong versions of `crio+pinns+crio-status` and `containerd+ctr+*shims*` combos (which are nasty to remember and debug), and wondering whether or not a locally patched version of all binaries were actually executed on the system. Changing the system setup (like removing crio/containerd packages or renaming original binaries) to prevent running wrong binaries would bring their own problems.

So far I've considered replacing the original binaries to be the easiest way to get a custom CRI installed, and `reinstall_{containerd,crio}=1 vm-install-cri` to rollback to the distro default version without fear of having anything left behind from the custom CRI.

That said, it would still be nice to have a cleaner solution to this problem. For instance building a custom crio/containerd package that would install everything (man pages, etc.). But having that for all the distros we support today sounds like quite a lot of work...

klihub:
> I would gladly start installing locally built binaries into /usr/local/bin on vm, if only the daemon binaries would not have any dependencies to executables. However, as this is not the case, I'm afraid that installing another set of binaries would result in running wrong versions of `crio+pinns+crio-status` and `containerd+ctr+*shims*` combos (which are nasty to remember and debug), and wondering whether or not a locally patched version of all binaries were actually executed on the system.

Hmm... That's nasty. So how simply overwriting only the original runtime binary with the one compiled by us (this PR) avoids hitting this very same implicit version/cross-binary dependency problem ?

demo/lib/vm.bash
```diff
@@ -699,21 +699,31 @@ vm-install-runc() {
...
+            vm-command 'command -v containerd'
+            if [ -n "$COMMAND_OUTPUT" ] && [ "x$COMMAND_STATUS" == "x0" ]; then
+                vm_cri_dir="${COMMAND_OUTPUT%/*}"
+            fi
             for f in ctr containerd containerd-stress containerd-shim containerd-shim-runc-v1 containerd-shim-runc-v2; do
-                vm-put-file "$containerd_src/bin/$f" "/usr/bin/$f"
+                vm-put-file "$containerd_src/bin/$f" "$vm_cri_dir/$f"
             done
+            vm-command "systemctl enable containerd"
             vm-command "systemctl start containerd"
```
jukkar, line 716: Could be combined to one command like this
```
vm-command "systemctl enable --now containerd"
```

askervin: Thanks @jukkar, changed to `enable --now` on both containerd and crio.

jukkar (approved):

PR #751, opened 2021-12-30
e2e: update distro=opensuse to 15.3

askervin:
- Add distro=opensuse-15.2 for testing older environments.
- Fix missing pidof by installing sysvinit-tools. This is necessary on
  15.3 and tumbleweed, yet it used to be tumbleweed-only.
- Change opensuse-install-pkg to not give an error if a package is
  already installed. This is aligned with debian-install-pkg. Without
  this change opensuse-15.2 and opensuse-15.3 behaviors would differ.


klihub (approved):
LGTM

PR #753, opened 2021-12-31
e2e: containernetworkplugins in /opt/cni/bin in fedora

askervin:
- Bind-mount /opt/cni/bin in Fedora also when installing crio, in
  addition to containerd.
- Fixes crio CNI detection issue in Fedora.


klihub (approved):
LGTM

PR #754, opened 2021-12-31
e2e: detect cri-resmgr(-agent) with fuser/pidfile instead of pidof

askervin:
- pidof cri-resmgr does not work when cri-resmgr runs with custom
  glibc versions.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -517,7 +517,7 @@ launch() { # script API
             vm-command "cat $(basename "$cri_resmgr_cfg")"
             if [[ "$k8scri" == cri-resmgr* ]]; then
                 # launch cri-resmgr as the top element in the k8s container runtime stack
-                cri_resmgr_mode="-relay-socket /var/run/cri-resmgr/cri-resmgr.sock -runtime-socket $cri_sock -image-socket $cri_sock"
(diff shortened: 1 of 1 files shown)


askervin:
Marking this as draft, because this should be aligned with "cri-resmgr as NRI plugin usecase". What would be the best routine to find out the PID of cri-resmgr and signal (kill) it in all cases?

klihub:
> Marking this as draft, because this should be aligned with "cri-resmgr as NRI plugin usecase". What would be the best routine to find out the PID of cri-resmgr and signal (kill) it in all cases?

We could change it to write a pid-file under `/run/cri-resmgr` once it has taken the socket... We could even have it register a simple gRPC-protocol of its own with a `Shutdown()` (and anything else we might see fit)  and the corresponding command line option `--shutdown` which would connect to the socket and do the gRPC `Shutdown()` client call. I don't know what would be the best.

marquiz:
A pid file wounds sounds like an easy-ish "standard" solution to me

askervin:
Good, sounds like a pid file then. :)

Opinions, shall I close this PR, or should we keep it in the list as a reminder. In the latter case I'll update this with s/fuser/pidfile when we have it.

klihub (approved):
LGTM

PR #755, opened 2022-01-03
e2e: add a test for distro/cni/cri combinations

askervin:
- Test all distro, k8scni and k8scri values.

- Testing all supported combinations of values is very
expensive (takes several days), but testing every value only in a
single case gives poor coverage. Testing full pairwise coverage in all
values is a reasonable compromise. For instance, if there is a CNI
that does not work in a distro, we well catch that error.

- The number of tests to be executed will grow slowly even if new
variables are added. In case of all combinations this number would
grow exponentially.

### DIFF
--- a/scripts/testing/pairwise
+++ b/scripts/testing/pairwise
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+
+"""pairwise - print var-value combinations that cover all value pairs
+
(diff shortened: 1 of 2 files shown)


jukkar (dismissed):

jukkar (approved):

klihub (approved):
Should be safe to merge.

PR #756, opened 2022-01-03
resmgr: write a PID file upon successful startup.

klihub:
Implement basic PID file handling. Write a PID file upon successful startup. If we fail to start up due to an existing active socket, try to read the process ID of the running instance from the PID file. If this succeeds write an diagnostic/error message with the read process ID.

### DIFF
--- a/pkg/cri/resource-manager/resource-manager.go
+++ b/pkg/cri/resource-manager/resource-manager.go
@@ -36,6 +36,7 @@ import (
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/visualizer"
 	"github.com/intel/cri-resource-manager/pkg/instrumentation"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
+	"github.com/intel/cri-resource-manager/pkg/pidfile"
 
 	policyCollector "github.com/intel/cri-resource-manager/pkg/policycollector"
 	"github.com/intel/cri-resource-manager/pkg/utils"
@@ -118,6 +119,10 @@ func NewResourceManager() (ResourceManager, error) {
 	}
 
 	if err := m.setupRelay(); err != nil {
+		pid, _ := pidfile.OwnerPid()
+		if pid > 0 {
+			m.Error("looks like we're already running as pid %d...", pid)
+		}
 		return nil, err
 	}
 
@@ -163,6 +168,13 @@ func (m *resmgr) Start() error {
 
 	if err := m.relay.Start(); err != nil {
 		return resmgrError("failed to start CRI relay: %v", err)
+	}
+
+	if err := pidfile.Remove(); err != nil {
+		return resmgrError("failed to remove stale/old PID file: %v", err)
+	}
+	if err := pidfile.Write(); err != nil {
+		return resmgrError("failed to write PID file: %v", err)
 	}
 
 	if opt.ForceConfig == "" {
--- a/pkg/pidfile/pidfile.go
+++ b/pkg/pidfile/pidfile.go
@@ -1,0 +1,165 @@
+// Copyright 2022 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pidfile
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	pidFilePath = defaultPath()
+	pidFile     *os.File
+)
+
+// GetPath returns the current pidfile path.
+func GetPath() string {
+	return pidFilePath
+}
+
+// SetPath sets the pidfile path to the given one.
+func SetPath(path string) {
+	close()
+	pidFilePath = path
+}
+
+// Write opens the PID file and writes os.Getpid() to it. If the PID file already
+// exists Write() fails with an error. On successful completion, Write keeps the
+// PID file open.
+func Write() error {
+	if pidFile != nil {
+		return nil
+	}
+
+	err := os.MkdirAll(filepath.Dir(pidFilePath), 0755)
+	if err != nil {
+		return errors.Wrap(err, "failed to create PID file")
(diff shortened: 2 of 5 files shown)


klihub:
@askervin, @marquiz, @jukkar If this looks remotely sane, I'll take another look at it, clean it up if/as necessary then remove the draft status.

klihub:
As a side-note, I did look for a sane-looking and maintained existing pidfile pkg. I only found the facebook abandonware and another one which among other things lacked any test cases. So I decided we'll roll our own.

askervin:
Do you think we should/could follow [the pattern](https://man7.org/tlpi/code/online/dist/filelock/create_pid_file.c.html) that leaves pidfile open and locked for the daemon when it runs? It seems that 100 % of pid files under my /var/run have the same `fuser` pid as in the file (and 50 % of the files include linefeed as the last character).

Recognizing running cri-resmgr by the lock would be safer than relying on the pid written to the file. Current logic might encourage killing an unlucky processes has received the pid of a former cri-resmgr, in case `setupRelay()` fails.

klihub:
> Do you think we should/could follow [the pattern](https://man7.org/tlpi/code/online/dist/filelock/create_pid_file.c.html) that leaves pidfile open and locked for the daemon when it runs? It seems that 100 % of pid files under my /var/run have the same `fuser` pid as in the file (and 50 % of the files include linefeed as the last character).

Keeping the PID file open should be fine and I can add it. I really wouldn't add locking it... at least not before [portable file locking](https://github.com/golang/go/issues/33974) lands in a golang release. AFAICT, this shouldn't be a problem since anyway we don't/won't use the lockfile to protect against multiple concurrently running instances (we use the relay socket for that). And keeping the file open alone should satisfy your need for the workaround to kill existing instances using `fuser` + `kill` install of `pkill`.

pkg/cri/resource-manager/resource-manager.go
```diff
@@ -118,6 +119,10 @@ func NewResourceManager() (ResourceManager, error) {
 	}
 
 	if err := m.setupRelay(); err != nil {
+		pid, _ := pidfile.Read()
```
marquiz, line 122: Hmm, shouldn't we do `ProcessExists()` here instead`

marquiz (commented):
Looks basically good to me 👌 I wouldn't add locking, either, but keeping the file open sounds fine

pkg/cri/resource-manager/resource-manager.go
klihub: Hmm... not sure, but maybe. Probably. The idea here was to try reading out the process idea from the PID file and print it out if we fail to start up. But it is true that it does not make much sense to print the process ID of a process if it does not exist, so maybe it should indeed be verified.

klihub: Updated to keep PID file open and to use (`OwnerPid()`, the updated equivalent of) `ProcessExists()` to read the PID upon failure to start up.

pkg/pidfile/pidfile.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Write opens the PID file and writes os.Getpid() to it. If the PID file already
+// exists, Write() fails with an error. On successful completion, Write keeps the
+// PID file open.
+func Write() error {
+	if pidFile != nil {
+		return nil
+	}
+
+	err := os.MkdirAll(filepath.Dir(pidFilePath), 0744)
```
jukkar, line 52: Hmm, should the directory permission be 0755 (so rwxr-xr-x)?

klihub:
@askervin Actually... do we really need this whole thing ? You can already do an `fuser /var/run/cri-resmgr/cri-resmgr.sock` to get the pid of the `cri-resmgr`. Granted, it should also give any CRI client being proxied, but that is only kubelet if any so disambiguation shouldn't be too difficult.

jukkar (dismissed):

askervin:
@klihub:
> You can already do an fuser /var/run/cri-resmgr/cri-resmgr.sock to get the pid of cri-resmgr.

Yeah, this is what I did in the draft PR #754, though it doesn't yet take the disambiguation (kubelet) into account.

I kept the PR as draft, as I had a doubt that it will not work anymore in the case where `cri-resmgr` runs as a NRI plugin. The pid file sounded - and still sounds - a nice and standardish solution, which could be easily extended to `cri-resmgr-agent`, too.

With this reasoning I'd still say this PR makes sense.

klihub:
> I kept the PR as draft, as I had a doubt that it will not work anymore in the case where `cri-resmgr` runs as a NRI plugin.

That's a good point. You would need to use the `/var/run/nri.sock` socket then. And that would normally be much more popular with `fuser` among processes... Maybe it is good idea to keep this then.

klihub:
@jukkar Sorry, I had to fix a few remaining test case name vs. test copy/paste typos in the tests and some comma spelling errors...

jukkar (approved):

askervin (approved):
LGTM, works fine. Thanks @klihub!

Not related to this patch, but when I was testing this, the first observation was something that I didn't expect: if I start one cri-resmgr, suspend it in terminal (ctrl-z), and start another one, the latter one prints:
```
...
W0105 13:43:19.196350   19065 server.go:203] removing abandoned socket '/var/run/cri-resmgr/cri-resmgr.sock' in use...
...
```
and starts just fine. After a couple of repetitions this results in:
```
# fuser -v /var/run/cri-resmgr/cri-resmgr.sock
USER        PID ACCESS COMMAND
/run/cri-resmgr/cri-resmgr.sock:
                     root      18963 F.... cri-resmgr
                     root      19065 F.... cri-resmgr
                     root      19400 F.... cri-resmgr
                     root      19631 F.... cri-resmgr
                     root      19733 F.... cri-resmgr
```
      
But obviously this is a corner case. The logic works fine and prints the PID of the running cri-resmgr when it runs normally in the background.

klihub:
> LGTM, works fine. Thanks @klihub!
> 
> Not related to this patch, but when I was testing this, the first observation was something that I didn't expect: if I start one cri-resmgr, suspend it in terminal (ctrl-z), and start another one, the latter one prints:
> 
> ```
> ...
> W0105 13:43:19.196350   19065 server.go:203] removing abandoned socket '/var/run/cri-resmgr/cri-resmgr.sock' in use...
> ...
> ```

Hmm... indeed a corner case. But I bet that behavior could be improved a lot by better socket probing. Now we're really dumb about it, simply try to connect to the socket, and all failures are considered to indicate an abandoned socket.

klihub:
> Not related to this patch, but when I was testing this, the first observation was something that I didn't expect: if I start one cri-resmgr, suspend it in terminal (ctrl-z), and start another one, the latter one prints:
> 
> ```
> ...
> W0105 13:43:19.196350   19065 server.go:203] removing abandoned socket '/var/run/cri-resmgr/cri-resmgr.sock' in use...
> ...
> ```

This corner case should be fixed by [this PR](https://github.com/intel/cri-resource-manager/pull/757).

PR #757, opened 2022-01-05
fixes: better socket probing.

klihub:
Invert active socket test logic by testing if connecting to the socket fails with ECONNREFUSED instead of going through the full gRPC stack and checking if a connection can be established.

This PR should fix the apparent misbehavior for the corner case described [in this comment](https://github.com/intel/cri-resource-manager/pull/756#pullrequestreview-844621132).

### DIFF
--- a/pkg/utils/net.go
+++ b/pkg/utils/net.go
@@ -15,6 +15,7 @@
 package utils
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/utils/net.go
```diff
@@ -94,9 +95,19 @@ func WaitForServer(socket string, timeout time.Duration, opts ...interface{}) er
 	}
 }
 
-// ServerActiveAt checks if a gRPC server is accepting connections at the socket.
-func ServerActiveAt(socket string) bool {
-	return WaitForServer(socket, time.Second) == nil
+// IsNonlisteningSocket returns true if nobody is listening at a socket.
```
jukkar, line 98: This is probably only me, but I have noticed that I always need to do mental twists with these negative tests when reading the code. Would it be possible to change this to `IsListeningSocket()` instead?

jukkar (approved):

klihub:
@askervin PTAL

askervin (approved):
LGTM, works fine!

PR #764, opened 2022-02-01
topology-aware: Remove duplicate branch

jukkar:
Both branches are the same so remove the if-check as it is not
needed here.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
@@ -636,23 +636,28 @@ func (p *policy) applyGrant(grant Grant) {
 			log.Debug("  => not pinning CPUs, allocated cpuset is empty...")
 		}
 		container.SetCpusetCpus(cpus)
-		if exclusive.IsEmpty() {
-			container.SetCPUShares(int64(cache.MilliCPUToShares(cpuPortion)))
-		} else {
-			// Notes:
-			//   Hmm... I think setting CPU shares according to the normal formula
-			//   can be dangerous when we do mixed allocations (both exclusive and
-			//   shared CPUs assigned). If the exclusive cpuset is not isolated and
-			//   there are other processes (unbeknown to us) running on some of the
-			//   same exclusive CPU(s) with CPU shares not set by us, those processes
-			//   can starve our containers with supposedly exclusive CPUs...
-			//   There's not much we can do though... if we don't set the CPU shares
-			//   then any process/thread in the container that might sched_setaffinity
-			//   itself to the shared subset will not get properly weighted wrt. other
-			//   processes sharing the same CPUs.
-			//
-			container.SetCPUShares(int64(cache.MilliCPUToShares(cpuPortion)))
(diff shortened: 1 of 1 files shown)


pkg/cri/resource-manager/policy/builtin/topology-aware/pools.go
```diff
@@ -636,23 +636,7 @@ func (p *policy) applyGrant(grant Grant) {
...
-			//   there are other processes (unbeknown to us) running on some of the
-			//   same exclusive CPU(s) with CPU shares not set by us, those processes
-			//   can starve our containers with supposedly exclusive CPUs...
-			//   There's not much we can do though... if we don't set the CPU shares
-			//   then any process/thread in the container that might sched_setaffinity
-			//   itself to the shared subset will not get properly weighted wrt. other
-			//   processes sharing the same CPUs.
-			//
-			container.SetCPUShares(int64(cache.MilliCPUToShares(cpuPortion)))
-		}
```
klihub, line 655: I would like to leave a word of warning there about the inherent danger in the (single) scheduling weight (per container) for mixed (exclusive + part of shared pool) allocations. It should be a more concise description of the problem/remedies instead of this train of thought á la Matti Nykänen...

klihub: Maybe something along the lines of

```
//
// Notes:
//     It is extremely important to ensure that the exclusive subset of mixed CPU allocations are really
//     exclusive at the level of the whole system and not just the orchestration. This is something we
//     can't really do from here reliably ATM.
//
//     We set the CPU scheduling weight for the whole container (all processes within the container)
//     according to container's partial allocation. This is typically a sub-CPU allocation (< 1000 mCPU)
//     which is meant to be consumed by an 'infra/mgmt' process within the container from the shared
//     subset of CPUs assigned to the container. The container entry point or the processes within the
//     container are supposed to arrange so that the 'infra' process(es) are pinned to the shared CPUs
//     and the 'data/performance critical' critical' process(es) to the exclusive CPU(s).
//
//     With this setup the kernel will slice out the correct amount of CPU from the shared pool for the
//     'infra' process as it competes with other workloads' processes in the same pool. Also the 'data'
//     process should run fine, since it does not need to compete for CPU with any other processes
//     in the system... as long as that allocation is genuinely system-wide exclusive.
```

klihub: Then again, maybe such a verbose explanation should go in the docs in connection with the introduction of the concept of mixed allocations, and here a short reference to the corresponding section in the documentation would be the correct way to address/document this.

@jukkar @askervin @marquiz @kad WDYT, guys ?

jukkar: Putting this at least to source file looks a correct thing to do, it is there next to a setting that affects it.

jukkar:
New version that adds the comment.

jukkar:
Rebased against upstream.

klihub (approved):
LGTM

PR #765, opened 2022-02-02
.githooks: add hook to check go version conistency.

klihub:
Add a hook to check the consistency of declared golang version requirement vs. the one used in github workflows. No need to always wait for the already slow and overloaded github workers to run our tests till failure then let us know we forgot to update the other.

### DIFF
--- a/.githooks/pre-commit.d/20-go-version
+++ b/.githooks/pre-commit.d/20-go-version
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+WORKFLOWS=".github/workflows/verify.yml"
+
+gomod=$(git diff --cached go.mod | grep '^+go ' | sed 's/^.* //g')
+[ -z "$gomod" ] && exit 0
+
+status=0
+for wf in $WORKFLOWS; do
+    workflow=$(grep 'go-version:' .github/workflows/verify.yml | sed 's/^.*: //')
+    if [ "gomod" != "$workflow" ]; then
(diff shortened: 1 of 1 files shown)


.githooks/pre-commit.d/20-go-version
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+WORKFLOWS=".github/workflows/verify.yml"
+
+gomod=$(git diff --cached go.mod | grep '^+go ' | sed 's/^.* //g')
```
marquiz, line 5: Would it be good to catch the problem even if not part of this commit, i.e. just grep `go.mod` directly without doing git diff(?)

marquiz (dismissed):
One question but looks good in this form, too

.githooks/pre-commit.d/20-go-version
klihub: Hmm... maybe. Should we then also always automagically enable our git hooks in the cloned repos (by making `install-git-hooks` a prerequisite of the `all` or `build` Makefile target). Otherwise the whole test is only run, and such mistakes are caught, by those (yours, truly) who enable the git hooks by running `make install-git-hooks`... Then again, if we enable the hooks for everyone, how those versions could go out of sync ever again, other people brutally bypassing all hooks using `git commit -n` ?

jukkar (dismissed):

klihub:
@marquiz @jukkar Rebased on latest master to CI passed, PTAL, and give a thumbs up/merge...

jukkar (approved):

PR #766, opened 2022-02-02
deps: update dependencies to k8s 1.23.3.

klihub:


marquiz (approved):
Let's merge.

One optional suggestion, though. We could update directly to v1.23.3 which is the latest afaiu.

On a sidenote we should really drop circleci...

klihub:
> Let's merge.
> 
> One optional suggestion, though. We could update directly to v1.23.3 which is the latest afaiu.

Ok, let me check if updating there goes nice and easy and if it does we can jump directly there...

> On a sidenote we should really drop circleci...

I'm fine with dropping circleci... @kad WDYT ?

jukkar (approved):

klihub:
@marquiz Updated deps to 1.23.3. If CI tests pass, PTAL...

marquiz (approved):
👍  LGTM

PR #777, opened 2022-02-18
topology-aware: Do not advertise shared CPUs if there are none

jukkar:
Only write shared CPU information to resource.sh if there
are shared CPUs allocated to the container.


klihub (dismissed):
LGTM

klihub (approved):

PR #779, opened 2022-02-23
docs: mention that golang 1.17 is needed for building

okartau:
Currently, "make build" fails with golang 1.16.x but succeeds with 1.17, so we should update the doc.
This PR proposes trivial 1.14 -> 1.17 change.
But also "Don't merge yet", RFC: what about we stop mentioning exact version which tends to be correct for short time
and needs manual update often. And it is usually correct while it states the latest version, isn't it?
So what about avoiding static binding and use something more fuzzy (but less incorrect) like "decent version",
leaving the possible determining of working version to the user (which will happen anyway in case of trouble)?

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -18,7 +18,7 @@ Although not recommended, you can install CRI Resource Manager from sources:
   - get the sources: `git clone https://github.com/intel/cri-resource-manager`
   - build and install: `cd cri-resource-manager; make build && sudo make install`
 
-You will need at least `git`, `golang 1.14` or newer, `GNU make`, `bash`,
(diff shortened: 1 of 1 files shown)


klihub:
Yes, I think we could have something like this...
...
You will need at least `git`, `golang` as indicated by `go.mod` or newer, `GNU make`, `bash`,
...
You can check the least required `golang` version by running `grep '^go ' go.mod` in the root of the cloned repository.

With Sphinx it should/would also be possible to extract that version from go.mod and use it in the generated documentation...

klihub:
> Of course with Sphinx it should/would also be possible to extract that version from go.mod and use it in the generated documentation...

#834 implements this.

PR #780, opened 2022-03-01
topology-aware: Add AdditionalReservedPoolNamespaces option support

jukkar:
User is able to mark containers in specific namespaces to reserved pool.


pkg/cri/resource-manager/policy/builtin/topology-aware/flags.go
```diff
@@ -55,16 +61,22 @@ func (fh *fakehints) merge(hints fakehints) {
...
 }
 
 // Register us for configuration handling.
 func init() {
 	config.Register(PolicyPath, PolicyDescription, opt, defaultOptions)
 	config.Register(AliasPath, PolicyDescription, aliasOpt, defaultOptions)
+
+	for _, str := range opt.AdditionalReservedPoolNamespaces {
+		additionalReservedPoolNamespacesRe = append(additionalReservedPoolNamespacesRe,
+			glob.MustCompile(str))
```
klihub, line 80: I think we need to delay this and trigger it from `configNotify()`.

As an additional nano-nit, since we do correctly test-compile glob-patterns, we might as well collect them into a precompiled `policy.reservedPoolNamespaces []filepath.Glob`. And if we choose do that, we could always make sure that slice has `kube-system` as the first item and change the current special casing of `kube-system` to just loop through that `Glob` slice trying to `Match()` each configured item.

klihub (commented):
Generally LGTM. I had a few small comments.
Additionally, I think we'll need a dedicated e2e test for this.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	contrib.go.opencensus.io/exporter/prometheus v0.1.1-0.20191218042359-6151c48ac7fa
 	github.com/cilium/ebpf v0.6.2
 	github.com/evanphx/json-patch v4.12.0+incompatible
+	github.com/gobwas/glob v0.2.3
```
klihub, line 10: Can't we just use the stock go `filepath.Glob` implementation here ? It's probably less capable than this one, but if we allow multiple patterns and them as logically OR'd I think the stock one should be good enough (IIRC, the only thing it badly lacks is support for `foo{bar1,bar2,bar3}xyzzy`).

jukkar:
Intermediate version, will check the e2e tests next.

@klihub Why is that circleci still present, I thought we got rid of it?

klihub:
> Intermediate version, will check the e2e tests next.
> 
> @klihub Why is that circleci still present, I thought we got rid of it?

Because only @kad has the rights to update the repo settings and nuke the CircleCI. Now CircleCI fails because it can't find a `.circleci/config.yml` configuration file for itself. AFAICT, this will keep showing up until we update the repo settings.

jukkar:
Updated according to comments.

jukkar:
* simplified the reserved namespace check
* e2e tests

klihub (approved):
LGTM.

Will need to update the docs, but we can do it with a separate PR.

PR #781, opened 2022-03-01
pkg/cri: switch internals to CRI v1, add ingress/egress v1/v1alpha2 CRI conversion.

klihub:
Switch the native/internal representation of pods and containers everywhere to CRI v1.
Implement ingress and egress protocol conversion between v1alpha2 and v1 to support
both pre-v1 clients and runtimes.

### DIFF
--- a/pkg/cri/client/client.go
+++ b/pkg/cri/client/client.go
@@ -24,7 +24,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 
-	api "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	api "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	"github.com/intel/cri-resource-manager/pkg/instrumentation"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
--- a/pkg/cri/client/v1/client.go
+++ b/pkg/cri/client/v1/client.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Intel Corporation. All Rights Reserved.
+// Copyright Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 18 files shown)


pkg/cri/server/v1alpha2/server.go
```diff
@@ -0,0 +1,225 @@
...
+package server
+
+import (
+	"context"
+	"unsafe"
+
+	"google.golang.org/grpc"
+
+	v1 "k8s.io/cri-api/pkg/apis/runtime/v1"
+	api "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
```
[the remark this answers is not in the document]
klihub: Updated. Didn't manage to leave off that remaining explicit (and IMO unnecessary) `v1` as `gofmt` seems to insist having it.

kad (approved):
In general, I'm ok with all of this. One separate worry that I have is about "unsafe.Pointer()" cast for practically 100% of data structures: does match in all cases of we have something that might be problematic?

jukkar (dismissed):
LGTM. Right now the option is enabled by default, is the plan to disable it by default or remove it, in the future?

klihub:
> In general, I'm ok with all of this. One separate worry that I have is about "unsafe.Pointer()" cast for practically 100% of data structures: does match in all cases of we have something that might be problematic?

The CRI-O guys landed a few commits in the `v1alpha2` CRI implementation to bring it struct-wise identical to `v1` with the explicit goal of being able to do this type of near-zero-cost protocol bridging. The CRI-O implementation is essentially doing identical thing to what we do here, and that is the only reason why I dared to go with that instead of a convert-by-copying approach.

klihub:
> LGTM. Right now the option is enabled by default, is the plan to disable it by default or remove it, in the future?

Yes, that's exactly what I had in mind: first we have it as opt-out, once we're sure we can expect mostly everybody to talk v1 we make it opt-in, and eventually we remove it altogether.

askervin (commented):
Tested on debian-sid with
- kubelet v1.19.0 (CRI v1alpha2) + containerd v1.6.0-41-gc523102b5
- kubelet v1.23.4 (CRI v1) + containerd v1.6.0-41-gc523102b5
...and it works fine.

Also tried out with containerd v1.4.0, that talks CRI v1alpha2.
As promised in the commit message, this is not supposed to work. cri-resmgr output:
```
F0303 07:32:12.850610    4447 main.go:89] failed to start resource manager: resource-manager: cache synchronization pod query failed: rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService
```

My only question would be related to this. Merging the PR will prevent using CRI-RM master branch with older container runtime versions, which will make it harder to integrate CRI-RM to certain environments. On the other hand, without this PR, CRI-RM works with old and new kubelet and runtimes as our primary backend runtimes implement compatibility layers.

What do you think, should we add support for v1alpha2 towards CRI runtimes, too? (In this PR or in a separate PR?)

klihub:
> Tested on debian-sid with
> 
> * kubelet v1.19.0 (CRI v1alpha2) + containerd v1.6.0-41-gc523102b5
> * kubelet v1.23.4 (CRI v1) + containerd v1.6.0-41-gc523102b5
>   ...and it works fine.
> 
> Also tried out with containerd v1.4.0, that talks CRI v1alpha2. As promised in the commit message, this is not supposed to work. cri-resmgr output:
> 
> ```
> F0303 07:32:12.850610    4447 main.go:89] failed to start resource manager: resource-manager: cache synchronization pod query failed: rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService
> ```
> 
> My only question would be related to this. Merging the PR will prevent using CRI-RM master branch with older container runtime versions, which will make it harder to integrate CRI-RM to certain environments. On the other hand, without this PR, CRI-RM works with old and new kubelet and runtimes as our primary backend runtimes implement compatibility layers.
> 
> What do you think, should we add support for v1alpha2 towards CRI runtimes, too? (In this PR or in a separate PR?)

The initial idea was to support bridging in both directions. Then I quite quickly changed my mind regarding bridging from *new clients to old runtimes*. I think that is neither worth the effort nor necessarily very safe: I'm not sure if clients talking the new protocol version would expect talking to that old a runtime. Also looking at the [official EOL chart](https://endoflife.date/kubernetes), I think dropping supporting for 1.19/older (using the same branch/binary as the latest one) should not be a problem.

If we consider keeping v1alpha2 support around for a longer time, then I think using a dedicated branch/pre-release branches for it could/would probably be the way to go. That branch/those branches would simply have one additional commit swapping v1 for v1alpha2.

WDYT?

askervin (approved):
Thanks @klihub! This sounds reasonable.

LGTM.

klihub:
@askervin @jukkar @kad @marquiz PTAL, I had to rebase this already approved one to resolve a conflict.

jukkar (dismissed):

klihub:
@kad, @marquiz, @askervin, @jukkar I consider merging this for the `0.7.0` release... WDYT ?

klihub:
> @kad, @marquiz, @askervin, @jukkar I consider merging this for the `0.7.0` release... WDYT ?

Okay, after testing that idea against our current e2e test set, it's better to keep this one out for now. Many of our tests pull in old runtimes without support for the v1 protocol.

fmuyassarov (commented):
I'm not well aware of the internals of CRI-resource-manager yet, but these changes look good to me from the code perspective.

askervin:
In our e2e test point of view... I didn't run full tests here, but checked that cri-resmgr can talk to distro's on-the-stock containerd:

`distro=debian-sid`: ok

`distro=ubuntu-22.04`: ok

`distro=ubuntu-20.04`: fail (containerd-1.5.9)
```
root@vm> cri-resmgr -relay-socket /var/run/cri-resmgr/cri-resmgr.sock -runtime-socket /var/run/containerd/containerd.sock -image-socket /var/run/containerd/containerd.sock -force-config cri-resmgr.cfg   >cri-resmgr.output.txt 2>&1 &
cri-resmgr last output line:
F: [   cri-resmgr   ] failed to start resource manager: resource-manager: failed to query runtime version: rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService
```

`distro=centos-8`: ok

`distro=fedora`: ok

ubuntu-20.04 end-of-life is Aug/2025. What do you think?

klihub:
> `distro=ubuntu-20.04`: fail (containerd-1.5.9)
...
> ubuntu-20.04 end-of-life is Aug/2025. What do you think?

I'm trying to become a better human being, so I can't/won't tell you that...

But I can see 3 choices here:
1) We forget about this PR for now.
2) We add dual-version bridging to the egress side (relay/server) as well.  
3) If we think there is a viable and new enough alternative to 20.04, we stop supporting it despite it not being EOL'd.

jukkar:
> 3\. If we think there is a viable more recent alternative to 20.04, we stop supporting it despite it not being EOL'd.

Hmm, please do not remove the support for 20.04 yet, I am using it in the VM with github actions.

klihub:
> > 3. If we think there is a viable more recent alternative to 20.04, we stop supporting it despite it not being EOL'd.
> 
> Hmm, please do not remove the support for 20.04 yet, I am using it in the VM with github actions.

Okay... I wonder if we has this enabled for our repo:

/hold

klihub:
> > > 3. If we think there is a viable more recent alternative to 20.04, we stop supporting it despite it not being EOL'd.
> > 
> > 
> > Hmm, please do not remove the support for 20.04 yet, I am using it in the VM with github actions.
> 
> Okay... I wonder if we has this enabled for our repo:
> 
> /hold

Hmm. The robots I'm looking for are not these...

klihub:
We'll need to pick this up again and get it eventually merged: v1alpha2 is [going the way of the dodo bird](https://github.com/cri-o/cri-o/pull/6347).

pkg/cri/client/v1alpha2/client.go
```diff
@@ -0,0 +1,901 @@
...
+}
+
+func (c *client) ListMetricDescriptors(ctx context.Context, in *v1.ListMetricDescriptorsRequest, opts ...grpc.CallOption) (*v1.ListMetricDescriptorsResponse, error) {
+	return nil, fmt.Errorf("unimplemented by CRI v1alpha2 RuntimeService")
+}
+
+func (c *client) ListPodSandboxMetrics(ctx context.Context, in *v1.ListPodSandboxMetricsRequest, opts ...grpc.CallOption) (*v1.ListPodSandboxMetricsResponse, error) {
+	return nil, fmt.Errorf("unimplemented by CRI v1alpha2 RuntimeService")
+}
+*/
```
jukkar, line 735: Why commented?

pkg/cri/client/client.go
```diff
@@ -237,6 +261,256 @@ func (c *client) dialNotify(socket string) {
...
+}
+
+func (c *client) ListMetricDescriptors(ctx context.Context, in *api.ListMetricDescriptorsRequest, opts ...grpc.CallOption) (*api.ListMetricDescriptorsResponse, error) {
+	return nil, fmt.Errorf("unimplemented by CRI RuntimeService")
+}
+
+func (c *client) ListPodSandboxMetrics(ctx context.Context, in *api.ListPodSandboxMetricsRequest, opts ...grpc.CallOption) (*api.ListPodSandboxMetricsResponse, error) {
+	return nil, fmt.Errorf("unimplemented by CRI RuntimeService")
+}
+*/
```
jukkar, line 472: ditto

jukkar (commented):
Perhaps some explanation could be added to the  commented out code, now it is unclear why it is done like that.

pkg/cri/server/v1alpha2/server.go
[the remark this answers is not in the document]
marquiz: nit: I'd probably use a "k8s style" naming like criv1 and criv1alpha1 , also elsewhere, but I'll leave it up to you

pkg/cri/client/v1alpha2/client.go
```diff
@@ -0,0 +1,901 @@
...
+	}
+
+	resp, err := c.rsc.Version(ctx, &req)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := alphaRespToV1(resp, v1resp); err != nil {
+		return nil, err
+	}
```
marquiz, line 116: nit/note: the "asymmetry" of the `req` and `v1resp` variables just caught my eye, ie. the other is pointer and the other is not. I think dropping the `var` block would improve the readability (and reduce LOC). WDYT?

```suggestion
	req := &api.VersionRequest{}
	if err := v1ReqToAlpha(in, req); err != nil {
		return nil, err
	}

	resp, err := c.rsc.Version(ctx, req)
	if err != nil {
		return nil, err
	}

	v1resp := &v1.VersionResponse{}
	if err := alphaRespToV1(resp, v1resp); err != nil {
		return nil, err
	}
```

marquiz (commented):
Thanks @klihub. Looks basically really good to me. Just a few notes that I leave to your consideration

pkg/cri/server/v1alpha2/server.go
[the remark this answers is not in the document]
klihub: Updating import naming convention like that.

pkg/cri/client/v1alpha2/client.go
klihub: They are being defined but not available yet in the CRI API version we use.

pkg/cri/client/client.go
klihub: Same reason.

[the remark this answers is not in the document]
klihub: Being defined, but not present yet in the currently used CRI API.

pkg/cri/client/v1/client.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func (c *client) ListMetricDescriptors(ctx context.Context, in *v1.ListMetricDescriptorsRequest, opts ...grpc.CallOption) (*v1.ListMetricDescriptorsResponse, error) {
+	return nil, fmt.Errorf("unimplemented by CRI v1alpha2 RuntimeService")
+}
+
+func (c *client) ListPodSandboxMetrics(ctx context.Context, in *v1.ListPodSandboxMetricsRequest, opts ...grpc.CallOption) (*v1.ListPodSandboxMetricsResponse, error) {
+	return nil, fmt.Errorf("unimplemented by CRI v1alpha2 RuntimeService")
+}
+*/
```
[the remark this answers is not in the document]
klihub: Being defined, but not yet present in the version being used.

pkg/cri/client/v1/client.go
```diff
@@ -0,0 +1,347 @@
...
+
+// HasRuntimeService checks if the client is configured with runtime services.
+func (c *client) HasRuntimeService() bool {
+	return c.rcc != nil
+}
+
+// HasImageService checks if the client is configured with image services.
+func (c *client) HasImageService() bool {
+	return c.icc != nil
+}
```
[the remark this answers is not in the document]
klihub: `HasImageService()` is unused ATM, so I removed it. The different logic for HasRuntimeService in `pkg/client/client.go` is necessary (until we get rid of the `null` pseudo-policy), because it is use during startup to certain initialization actions. So it needs to consider configuration/desired state instead of the current state, which might be transiently different.

pkg/cri/client/v1alpha2/client.go
```diff
@@ -0,0 +1,901 @@
...
+
+// HasRuntimeService checks if the client is configured with runtime services.
+func (c *client) HasRuntimeService() bool {
+	return c.rcc != nil
+}
+
+// HasImageService checks if the client is configured with image services.
+func (c *client) HasImageService() bool {
+	return c.icc != nil
+}
```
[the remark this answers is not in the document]
klihub: Yes, these were just copy-paste leftover. Removed.

pkg/cri/client/v1alpha2/client.go
klihub: That asymmetry is a very good point, thanks for spotting it ! Rewrote it, not exactly as you suggested but instead using a var-block with two pointer assignments as my personal preference here.

pkg/cri/resource-manager/policy/builtin/topology-aware/mocks_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/intel/goresctrl/pkg/sst"
 	idset "github.com/intel/goresctrl/pkg/utils"
 	v1 "k8s.io/api/core/v1"
-	cri "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	cri "k8s.io/cri-api/pkg/apis/runtime/v1"
```
marquiz, line 29: Now that you renamed (almost) all the cri imports would we want to use `criv1` everywhere, for consistency? Again, I'll leave up to you. I think we merge PR also as is.

klihub: Definitely. I just haven't realized that I missed some. Bad grep pattern... And ditto for the other remaining inconsistencies in `cache`.

jukkar (approved):

PR #782, opened 2022-03-02
podpools: never count default pool CPUs in Instances percentages

askervin:
- This change enables easy node configuration by using allowing syntax
  "Instances: 100%" in user-defined pod pools. This notation means
  "allocate all the remaining CPUs to this pool".
- The behavior is the same as previously, if the CPUs of the built-in
  "default" pool are not customized by the user.
- The behavior differs from previous when a new set of CPUs is
  allocated to the "default" pool, that is, CPUs of the default pool and
  the reserved pool are disjoint. The previous implementation did not
  subtract these default CPUs from the CPU count that is the basis
  for "n %" notation. As a result, it was impossible to guess the
  percentage that would mean "all the rest CPUs" when core count varies.


klihub (approved):
LGTM.

jukkar (approved):

PR #788, opened 2022-03-10
resmgr: rediscover container resources on startup.

klihub:
This PR improves startup-time resource discovery of running containers.

It does this primarily by sending each discovered container a `ContainerStatus` request with the `Verbose` flag set, digging out the bits corresponding to the original `CRI LinuxContainerResources`, and reconstructing that from the data. This reconstructed synthetic version is then used as the original one to estimate the container's resource requirements.

In addition to this, the CPU shares setup logic has been altered to allow more correct/accurater rediscovery of CPU requests of running containers, in case the stored cache on disk of `cri-resource-manager` has been nuked and it needs to start with zero knowledge about the system state. See the committed comment changes for more details.

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -67,6 +67,9 @@ const (
 
 	// TopologyHintsKey can be used to opt out from automatic topology hint generation.
 	TopologyHintsKey = "topologyhints" + "." + kubernetes.ResmgrKeyNamespace
+
+	// CPURequestKey annotates saved CPU requests eg. for mixed allocations.
+	CPURequestKey = "cpu-request" + "." + kubernetes.ResmgrKeyNamespace
 )
 
 // allControllers is a slice of all controller domains.
@@ -321,6 +324,10 @@ type Container interface {
 	GetCPUQuota() int64
 	// GetCPUShares gets the CFS CPU shares of the container.
 	GetCPUShares() int64
+	// HasCPURequestAnnotations checks for a AnnotateCPURequest() annotation.
+	HasCPURequestAnnotation() bool
+	// ParseCPURequestAnnotation parses an annotation saved by AnnotateCPURequest().
+	ParseCPURequestAnnotation(string) int
 	// GetmemoryLimit gets the memory limit in bytes for the container.
 	GetMemoryLimit() int64
 	// GetOomScoreAdj gets the OOM score adjustment for the container.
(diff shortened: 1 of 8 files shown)


pkg/cri/resource-manager/cache/cache.go
```diff
@@ -901,7 +910,7 @@ func (cch *cache) LookupPod(id string) (Pod, bool) {
 }
 
 // Insert a container into the cache.
-func (cch *cache) InsertContainer(msg interface{}) (Container, error) {
+func (cch *cache) InsertContainer(msg interface{}, status *cri.ContainerStatusResponse) (Container, error) {
```
marquiz, line 913: nit/note: this function prototype just looks really cranky

Could we e.g. split this into two separate functions, one called from outside of the `cache´ package:

```
func (cch *cache) InsertContainer(msg *cri.CreateContainerRequest) (Container, error) {
	c := &container{
		cache: cch,
	}
	if err := c.fromCreateRequest(msg); err != nil {
		return nil, cacheError("failed to insert container %s: %v", c.CacheID, err)
	}
	return cch.insertContainer(c)
}
```

...and one internal function that would contain most of the logic here:

```
func (cch *cache) insertContainer(c *container) (Container, error) {
   ...
}
```

We would then move the logic from `case *cri.Container: ` down into the `RefreshContainers()` method.

WDYT? Could be done in a separate pr, too

pkg/cri/resource-manager/cache/utils.go
```diff
@@ -41,6 +43,77 @@ func IsPodQOSClassName(class string) bool {
...
+		return nil
+	}
+
+	info := struct {
+		RuntimeSpec *rspec.Spec `json:"runtimeSpec"`
+	}{}
+
+	err := json.Unmarshal([]byte(s.Info["info"]), &info)
+	if err != nil {
+		return nil
```
marquiz, line 58: Would we want to log this failure?

pkg/cri/resource-manager/policy/builtin/topology-aware/resources.go
```diff
@@ -949,6 +949,22 @@ func newRequest(container cache.Container) Request {
...
 		container.PrettyName(), cpuType, full, fraction, isolate)
 
+	if full > 0 && fraction > 0 {
+		state := container.GetState()
+		if state != cache.ContainerStateCreating {
+			if !container.HasCPURequestAnnotation() {
+				fraction += 1000 * full
+				full = 0
+				log.Warn("%s: state %d => falling back to shared CPU (%d) allocation",
+					container.PrettyName(), container.GetState(), fraction)
```
marquiz, line 959: I think a short note/reasoning comment here too would make future readers happier :) I.e. why are we doing this "hack" for mixed allocations,

marquiz (commented):
Some small nits/comments but I think this change makes sense and would improve things

klihub:
All valid comments. I'll address those, but I also need to update the actual rediscovery code. IIRC, it was not fully kosher yet...

PR #789, opened 2022-03-14
topology-aware: Add ReservedPoolNamespaces documentation

jukkar:
Document how to use ReservedPoolNamespaces option.

### DIFF
--- a/docs/policy/topology-aware.md
+++ b/docs/policy/topology-aware.md
@@ -369,3 +369,26 @@ the amount of memory for containers in `Burstable` pods or run the
 [resource-annotating webhook](../webhook.md) to provide `cri-resmgr` with
 an exact copy of the resource requirements from the Pod Spec as an extra
 Pod annotation.
+
+## Reserved pool namespaces
+
+User is able to mark certain namespaces to have a reserved CPU allocation.
+Containers belonging to such namespaces will only run on CPUs set aside
+according to the global CPU reservation, as configured by the ReservedResources
+configuration option in the policy section.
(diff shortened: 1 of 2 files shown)


docs/policy/topology-aware.md
```diff
@@ -369,3 +369,22 @@ the amount of memory for containers in `Burstable` pods or run the
...
+The `ReservedPoolNamespaces` option is a list of namespace globs that will be
+allocated to reserved CPU class.
+
+For example:
+
+```yaml
+policy:
+  Active: topology-aware
+  topology-aware:
+    ReservedPoolNamespaces: ["my-pool","reserved-*","kube-system"]
```
klihub, line 385: I think it'd be better to leave out `kube-system` here, because as mentioned later it is implicitly always a reserved pool.

docs/policy/topology-aware.md
```diff
@@ -369,3 +369,22 @@ the amount of memory for containers in `Burstable` pods or run the
 [resource-annotating webhook](../webhook.md) to provide `cri-resmgr` with
 an exact copy of the resource requirements from the Pod Spec as an extra
 Pod annotation.
+
+## Reserved pool namespaces
+
+User is able to mark certain namespaces to have reserved CPU allocation.
```
klihub, line 375: I think we should explain a bit more (a few sentences max.) what this means in practice. Maybe something like "Containers belonging to such namespaces will only run on CPUs set aside according to the global CPU reservation, as configured by the `ReservedResources` configuration option in the `policy` section.

klihub (approved):
LGTM.

askervin (approved):
LGTM

PR #790, opened 2022-03-18
e2e: fix podXcY allowed resources finder

askervin:
- Prevent shell expansion of process pattern in case there happens to
  be a file that matches the pattern.


jukkar (approved):

klihub (approved):
LGTM

PR #793, opened 2022-03-18
Statically linked binaries in container images

marquiz:
Use statically linked binaries in the agent and webhook container
images. We now have scratch-based images than only contain one file
(which is the agent or webhook binary) - clean and simple. With this we
also ditch the cumbersome clearlinux based build which does not even
support go v1.17 (or later).


jukkar (approved):

klihub (approved):
LGTM

PR #795, opened 2022-03-21
sysfs: workaround for bogus NUMA meminfo entry.

klihub:
On some HW we've seen more free that total memory being reported.
This causes an unsigned integer underflow in the calculation of
the amount of memory being used, resulting in exorbitant amounts
being reported. As a workaround, we calculate the amount of used
memory ourselves, capping the free memory to the total available
at max. to avoid similar underflows.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -891,7 +891,6 @@ func (n *node) MemoryInfo() (*MemInfo, error) {
 		map[string]interface{}{
 			"MemTotal:": &buf.MemTotal,
 			"MemFree:":  &buf.MemFree,
-			"MemUsed:":  &buf.MemUsed,
 		},
 		func(line string) (string, string, error) {
 			fields := strings.Fields(strings.TrimSpace(line))
@@ -910,6 +909,23 @@ func (n *node) MemoryInfo() (*MemInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	//
+	// On some HW and kernel combinations we've seen more free than total
+	// memory being reported. This causes exorbitant usage of memory being
+	// reported which later can cause failures in policies which trust and
+	// rely on this information.
+	//
+	// Give here a clear(er) error about that. This should also prevent us
+	// immediately from starting up.
+	//
+	if buf.MemFree > buf.MemTotal {
+		return nil, sysfsError(meminfo, "System reports more free than total memory. "+
+			"This can be caused by a kernel bug. Please update your kernel.")
+	}
+
+	buf.MemUsed = buf.MemTotal - buf.MemFree
+
 	return buf, nil
 }
 


pkg/sysfs/system.go
```diff
@@ -891,7 +891,6 @@ func (n *node) MemoryInfo() (*MemInfo, error) {
 		map[string]interface{}{
 			"MemTotal:": &buf.MemTotal,
 			"MemFree:":  &buf.MemFree,
-			"MemUsed:":  &buf.MemUsed,
```
jukkar, line 894: Why do we remove this? It is not obvious from the commit.

klihub: Because I fsck'd up the commit...

klihub: Updated with a hopefully less b0rken fix.

jukkar (approved):

askervin (dismissed):
LGTM

pkg/sysfs/system.go
```diff
@@ -910,6 +909,20 @@ func (n *node) MemoryInfo() (*MemInfo, error) {
...
+	//
+	// On some HW we've seen more free that total memory being reported.
+	// This causes an unsigned integer underflow in the calculation of
+	// the amount of memory being used, resulting in exorbitant amounts
+	// being reported. As a workaround, we calculate the amount of used
+	// memory ourselves, capping the free memory to the total available
+	// at max. to avoid similar underflows.
+	//
+	if buf.MemFree > buf.MemTotal {
+		buf.MemFree = buf.MemTotal
```
kad, line 922: I'd say `MemFree = MemTotal - MemUsed` in this case.

kad:
The more I'm thinking about that, the more I'm inclined to report that as an "buggy kernel detected. Please upgrade/fix your kernel".

klihub:
> The more I'm thinking about that, the more I'm inclined to report that as an "buggy kernel detected. Please upgrade/fix your kernel".

Should we report it as an unrecoverable error either with `log.Fatal` (doing an `os.Exit()`) or with `log.Panic()` (`panic()`ing) ?

jukkar (dismissed):
LGTM

kad:
I think we agreed that this is a kernel bug, so workarround might be not good to implement. Shall we close this PR?

klihub:
> I think we agreed that this is a kernel bug, so workarround might be not good to implement. Shall we close this PR?

But a `log.Fatal()` or a `panic()` would still probably be the right course of action.

klihub:
@askervin @jukkar @kad @marquiz PTAL, I updated this to bail out with an error if we run into such a buggy kernel.

marquiz (commented):
One typo found.

Completely refusing to start in this scenario feels like a very strong action  Just wondering what if every software piece in the system refused to start if it found out that there's an unfixed kernel bug 😜 But obviously you've thought about it more thoroughly so can go with this

klihub:
> One typo found.
> 
> Completely refusing to start in this scenario feels like a very strong action Just wondering what if every software piece in the system refused to start if it found out that there's an unfixed kernel bug stuck_out_tongue_winking_eye But obviously you've thought about it more thoroughly so can go with this

Well, the original patch tested for the same condition, then capped the free memory to the total available if the former was larger than the latter. I changed it to this bail out (encouraged by the fact that we've never seen this problem again once our downstream customer updated their kernel. At least with the topology-aware policy enabled, even if we didn't bail out here there would be a failure to start later when the policy tries to initialize itself. However, I think it is better to fail contextually as close to detecting the real problem as possible as it usually provides better error context and helps the end user in diagnosing the problem instead of sending him down a wild goose chase.

pkg/sysfs/system.go
```diff
@@ -910,6 +909,23 @@ func (n *node) MemoryInfo() (*MemInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	//
+	// On some HW we've seen more free that total memory being reported.
```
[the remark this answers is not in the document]
klihub: And there was that `whcih` that went unnoticed by either of us... I rewrote the comment and the commit message to be cleared... hopefully without typos this time.

marquiz (approved):
lgtm

PR #796, opened 2022-03-23
crash fix: don't try to rebalance/deliver events/trigger post-update hooks with nil policy.

klihub:
Don't trigger rebalancing without a policy and crash as an
end result (in the generic policy.Rebalance()). One way to
trigger such a crash (sans this commit) is to run with the
agent enabled and some external adjustment CustomResources
in place while no/nil policy is active.

Put in similar protection against trying to deliver events
or running post-update hooks for an updated configuration
with nil policy.


askervin (approved):
LGTM

jukkar (approved):

PR #798, opened 2022-03-30
RFC: Makefile: help linking against custom libc installed on host

askervin:
- Example: make GLIBC_PREFIX=/my/glibc/install-prefix
  links against given glibc version.


askervin:
Abandoning this PR. Linking against a custom glibc will not be needed, as we can build static binaries with PR #800 .

PR #799, opened 2022-03-30
control: add cpu controller

marquiz:
For controlling cpu parameters such as cpufreq attributes. This patch
adds a stub controller that is missing the functionality for actually
configuring the cpus (cpufreq).

It implements dynamic configuration updates as well as internal state
tracking (stored in cri-rm cache) that stores the information about
which classes cpus are assigned to (if any).

The actual controller hook functionality is also currently a stub and it
does not rely on any container lifetime hooks.

An example config:

```
cpu:
  classes:
    high-prio:
        minFreq: 2000000
        maxFreq: 5000000
    low-prio:
        minFreq: 1000000
        maxFreq: 1500000
```

### DIFF
--- a/pkg/cri/resource-manager/cache/cache.go
+++ b/pkg/cri/resource-manager/cache/cache.go
@@ -38,6 +38,8 @@ import (
 )
 
 const (
+	// CPU marks changes that can be applied by the CPU controller.
+	CPU = "cpu"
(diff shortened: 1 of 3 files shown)


marquiz:
I think this is ready for review

ping @klihub @jukkar @askervin

jukkar (approved):
LGTM

PR #802, opened 2022-04-06
Add CPU controller calls to balloons policy

jukkar:
This PR adds calls to the CPU controller to set a CPUs to a given CPU class for the balloons policy.
First patch is from @marquiz and can be found here #799

### DIFF
--- a/pkg/cri/resource-manager/control/cpu/cpu.go
+++ b/pkg/cri/resource-manager/control/cpu/cpu.go
@@ -140,7 +140,13 @@ func (ctl *cpuctl) assign(class string, cpus ...int) error {
 		return fmt.Errorf("non-existent cpu class %q", class)
 	}
 
-	//TODO: configure cpus (sysfs)
+	if err := utils.SetCPUsScalingMinFreq(cpus, (int)(ctl.config.Classes[class].MinFreq)); err != nil {
+		return fmt.Errorf("Cannot set min freq %d: %w", ctl.config.Classes[class].MinFreq, err)
+	}
+
+	if err := utils.SetCPUsScalingMaxFreq(cpus, (int)(ctl.config.Classes[class].MaxFreq)); err != nil {
+		return fmt.Errorf("Cannot set max freq %d: %w", ctl.config.Classes[class].MaxFreq, err)
+	}
 
 	// Store the class assignment. Assign cpus to a class and remove them from
 	// other classes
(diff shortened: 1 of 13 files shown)


pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
```diff
@@ -848,6 +851,7 @@ func (p *balloons) setConfig(bpoptions *BalloonsOptions) error {
 	for blnIdx, bln := range balloons {
 		log.Info("- balloon %d: %s", blnIdx, bln)
 	}
+
```
jukkar, line 854: Stray change, I will remove this one.

jukkar:
Fixing the test compilation.

pkg/cri/resource-manager/control/cpu/cpu.go
```diff
@@ -0,0 +1,253 @@
...
+	return cache.TraversePendingConfig(Assign)
+}
+
+// assign assigns a cpuset to a class
+func (ctl *cpuctl) assign(class string, cpus ...int) error {
+	if _, ok := ctl.config.Classes[class]; !ok {
+		return fmt.Errorf("non-existent cpu class %q", class)
+	}
+
+	if ok := utils.SetCPUsScalingMinFreq(cpus, (int)(ctl.config.Classes[class].MinFreq)); ok != nil {
```
askervin, line 143: Nitpick: usually a return value of the `error` type has been captured to `err` variable. `ok` has captured booleans.

askervin (commented):
One nitpick.

Functionality looks pretty good. I sketched an [e2e test extension](https://github.com/askervin/cri-resource-manager/commit/d3294e635bde171bc9446acd071bdd04f5a992a4) that would verify classes in your e2e cri-resmgr.cfg. It's far from complete (maybe checking CPU ids and/or freq changes after deflating balloons would make sense, too), but I think it could be used as is to have something to start with.

I was not able to verify that the reserved CPUs would get correct frequency.

I did not test frequency changes on deflating or popping balloons.

pkg/cri/resource-manager/control/cpu/cpu.go
jukkar: Sure, that can be fixed of course. I can update the PR later today and add other comments if there are any.

askervin:
Update: extended [the e2e test](https://github.com/askervin/cri-resource-manager/commit/5370327acd10a2ec029a213a2c9e8af18668701c) to cover writing scaling_min_freq of a correct CPU id. The CPU id in the error message must match any of the CPUs in the cpuset of the balloon. With [updated goresctrl](https://github.com/intel/goresctrl/pull/92) the test is passed.

jukkar:
Replaced by #806

PR #804, opened 2022-04-11
sst: Pass around pointer to SST info struct

jukkar:
Earlier code manipulated full SST info in CPU discovery, but this
is not needed as we can just use the pointer to SST info in the code.
This clarifies the usage, as the SstPackageInfo struct contains a slice
which can confuse the user if SstPackageInfo is not passed around as a
pointer value. This also saves some tiny amount of memory.

### DIFF
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -446,6 +446,7 @@ func (c *topologyCache) discoverSstCPUPriority(sys sysfs.System, pkgID idset.ID)
 	// 2. SST-CP dictates over SST-BF
 	// 3. SST-BF is meaningful if neither SST-TF nor SST-CP is enabled
 	switch {
+	case sst == nil:
 	case sst.TFEnabled:
 		log.Debug("package #%d: using SST-TF based CPU prioritization", pkgID)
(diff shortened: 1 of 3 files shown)


pkg/cpuallocator/allocator.go
```diff
@@ -438,54 +438,56 @@ func (c *topologyCache) discoverSstCPUPriority(sys sysfs.System, pkgID idset.ID)
...
-				p = PriorityHigh
+	if sst != nil {
+		// Determine SST-based priority. Based on experimentation there is some
+		// hierarchy between the SST features. Without trying to be too smart
+		// we follow the principles below:
+		// 1. SST-TF has highest preference, mastering over SST-BF and making most
+		//    of SST-CP settings ineffective
+		// 2. SST-CP dictates over SST-BF
+		// 3. SST-BF is meaningful if neither SST-TF nor SST-CP is enabled
+		switch {
```
klihub, line 449: nitpick: since you have an expressionless switch, I'd suggest making `case sst == nil:` the first branch/case.

klihub (commented):
LGTM... with a small nitpick.

marquiz (approved):
lgtm

klihub (approved):
LGTM.

PR #805, opened 2022-04-13
agent: only print warning about missing watch once

marquiz:
Reduce annoying/confusing noise in logs.

### DIFF
--- a/pkg/agent/kubernetes.go
+++ b/pkg/agent/kubernetes.go
@@ -262,9 +262,17 @@ func (w *watch) Start(name string) {
 
 		for {
 			if events == nil {
-				w.parent.Info("creating %s watch", w.Name())
+				if ratelimit == nil {
+					w.parent.Info("creating %s watch", w.Name())
+				} else {
+					w.parent.Debug("creating %s watch", w.Name())
+				}
 				if k8w, err = w.openfn(w.ns, w.name); err != nil {
-					w.parent.Warn("failed to create %s watch: %v", w.Name(), err)
+					if ratelimit == nil {
+						w.parent.Warn("failed to create %s watch: %v", w.Name(), err)
+					} else {
+						w.parent.Debug("failed to create %s watch: %v", w.Name(), err)
+					}
 					ratelimit = time.After(1 * time.Second)
 				} else {
 					events = k8w.ResultChan()


klihub:
Is this a good idea ? Shouldn't we get rid of the busy-looping, instead of silently busy-looping ?

marquiz:
> Is this a good idea ? Shouldn't we get rid of busy-looping, instead of just busy-looping silently ?

How do we detect when/if the CRD is created without busy-looping?

askervin (approved):
LGTM.

marquiz:
Hmm, ok, it was actually because the CRD (definition) was missing from my new cluster.
```
k apply -f ./pkg/apis/resmgr/v1alpha1/adjustment-schema.yaml 
```
fixed the problem. I was mislead thinking that an actual adjustment object is missing. Do any of our deployments actually create the CRD? Because that log thrashing is very annoying.

Maybe the solution would be for the agent to create the CRD if it doesn't exist?

klihub:
> Hmm, ok, it was actually because the CRD (definition) was missing from my new cluster.
> 
> ```
> k apply -f ./pkg/apis/resmgr/v1alpha1/adjustment-schema.yaml 
> ```
> 
> fixed the problem. I was mislead thinking that an actual adjustment object is missing. Do any of our deployments actually create the CRD? Because that log thrashing is very annoying.

Was the low-pass filter/rate-limiter effective in your case then ? IOW, did you get only 1 or 2 log trashing messages per second or more ?

> Maybe the solution would be for the agent to create the CRD if it doesn't exist?

Yes, probably that's the right course of action if we can't find a way to make this work saner in any other way.

klihub:
> Do any of our deployments actually create the CRD? Because that log thrashing is very annoying.

Not at the moment. It should be added to the agent deployment file.

I actually tried that a few weeks ago in a test cluster when I ran into this problem. I added the CRD creation to the agent deployment file. But then at my first try I got a weird error which discouraged me and made me suspect that `kubectl apply -f` processes objects from a single stream in parallel... But then I could never reproduce the problem. Every time I later deleted everything and recreated from a single deployment file, things worked like a charm. We need to check, and if it really works reliably then that would be clearly the easiest and most convenient fix.

marquiz:
> Was the low-pass filter/rate-limiter effective in your case then ? IOW, did you get only 1 or 2 log trashing messages per second or more ?

Yea, the rate limiting was effective, but it's still four figure number messages per hour.

marquiz:
@klihub was right, I don't think this is not the right thing to do. We should make sure that the CRD is always available but that's another story... Closing the PR

PR #806, opened 2022-04-13
Add cpu controller and integrate to balloons policy

marquiz:
Implement a new controller for handling cpu parameters such as cpufreq
attributes. It supports dynamic configuration updates and stores its
state in the cache so that configuration is restored in startup. This
patch adds a stub controller in the sense that it is still missing the
integration for actually enforcing the cpu configuration in sysfs
(cpufreq). The actual controller event hooks are stubs, too, as the
functionality is not directly tied to any container lifetime events.

An example configuration:

```
cpu:
  classes:
    high-prio:
        minFreq: 2000000
        maxFreq: 5000000
    low-prio:
        minFreq: 1000000
        maxFreq: 1500000
```

**NOTE/TODO**: the internal state is stored in policy-specific data store.
Implementing generic functionality (and API) for storing
controller-specific data and a proper mechanism for events/hooks not
tied to container lifetime events is left as a future exercise.

This PR also integrates the cpu class assignment to the balloons policy.

Supersedes: #799 and #802


marquiz:
Ping @klihub @jukkar @askervin 

What do you think about this PR? I still maintain that it would be simpler/cleaner to have this "POC cpu controller only" PR. And, address the cache API changes and new event/hook mechanism properly in a separate PR, giving thought to it.

jukkar (approved):

PR #810, opened 2022-05-06
topology-aware: Add support for reserved CPU annotations

jukkar:
Allow user to add pod annotations that cause the CPUs to be
allocated from reserved pool.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
+++ b/pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
@@ -41,6 +41,8 @@ const (
 	keyMemoryTypePreference = "memory-type"
 	// annotation key for type "cold start" of workloads
 	keyColdStartPreference = "cold-start"
+	// annotation key for reserved pools
+	keyReservedCPUsPreference = "prefer-reserved-cpus"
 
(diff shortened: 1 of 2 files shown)


kad (approved):
/lgtm

pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
```diff
@@ -339,6 +343,20 @@ func checkReservedPoolNamespaces(namespace string) bool {
 	return false
 }
 
+func checkReservedCPUsAnnotations(c cache.Container) bool {
+	hintSetting, ok := c.GetEffectiveAnnotation(preferReservedCPUsKey)
+	if !ok {
+		return false
+	}
+
+	preference, err := strconv.ParseBool(hintSetting)
```
klihub, line 352: I think this would warrant logging an error.

klihub (changes requested):
LGTM.

One small comment about silently swallowing an annotation error, which I think we should fix. Apart from that an end-to-end test would be nice. If we want to get this merged before the 11:30 chat, we can do that if we address the swallowed-error issue. The e2e test we can also add with another PR.

klihub (approved):
LGTM

kad (approved):

PR #813, opened 2022-05-09
balloons: reassign all containers on start and config change

### LINKED ISSUE #809: Balloons policy prometheus metrics issues
`curl http://localhost:8891/metrics` output:
- `balloontype[0]` appears twice when initialized MinBalloons: 2
- not updated always (new balloons not always showing up)
- sometimes getting CONN_FAIL

askervin:
Fixes resource accounting on `cri-resmgr` (re)launch and config change. Fixes #809 .


jukkar (commented):
I am sure this works, but it would be nice to get some more description in commit message what this actually does, and perhaps have "Fixes: ...." link to the issue. It is not obvious from the commit, why we need to do this.

pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
```diff
@@ -713,7 +714,7 @@ func (p *balloons) configNotify(event pkgcfg.Event, source pkgcfg.Source) error
 		return err
 	}
 	log.Info("config updated successfully")
-	p.Sync(p.cch.GetContainers(), nil)
+	p.Sync(p.cch.GetContainers(), p.cch.GetContainers())
```
marquiz, line 717: Hmm, is there a big penalty, or possible issues, in doing a full resource re-allocation (dealloc followed by alloc). Does an unchanged configuration dterministically resolve into the same resource allocation every time? 🧐

Just started to think about the re-config events possibly arriving in short intervals (dynamic cpu class configuration can is done through the config).  In any case this is probably ok for a POC phase. But we might start thinking about detecting whether the effective configuration actually changed. WDYT?

askervin: I agree, we should split this later on to make it more efficient. For instance:
- Do full re-allocation only if any of balloon types has changed on those parts that affect
  1. CPUs allocated in balloon instances (MinCPUs, MaxCPUs, AllocatorPriority, MinBalloons), or
  2. how containers should be placed in balloons (Prefer* parameters, Namespaces, balloon type Name).
- Do only reconfiguration of CPU frequencies if CPUClass of a balloon type has changed, or if CPU class parameters have changed.
- Do only re-pinning of container CPUs/mems, if PinCPU/PinMemory of a those parameters have changed.

marquiz: A simple method would be a hash (or comparison) over the raw configuration data. It would be good to add this functionality in the upper config management layer, though, I think. So that other modules could also take advantage of this.

Just a sidenote that we wouldn't need to be concnerned about the CPU class parameters in the balloons policy. Those should be handled by the cpu controller

marquiz (dismissed):
Looks good to me.

One suggestion: what about adding a TODO note (code comment) that we might want to improve this in the future, by only re-allocating if the config has actually changed. WDYT(?)

askervin:
Thanks @jukkar and @marquiz for your feedback! TODO added and commit message improved.

jukkar (approved):

marquiz (approved):
LGTM

pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
klihub: > Hmm, is there a big penalty, or possible issues, in doing a full resource re-allocation (dealloc followed by alloc). Does an unchanged configuration dterministically resolve into the same resource allocation every time? monocle_face
> 
> Just started to think about the re-config events possibly arriving in short intervals (dynamic cpu class configuration can is done through the config).

IMO, rate-limiting configuration changes should be done at the configuration ingress. I don't recall now without checking if we still have it (I suspect we do), but we used to have a low-pass filter in the agent that anticipated configuration bursts and only delivered the new data once (frequent enough) changes stopped coming in.

PR #815, opened 2022-05-10
control/cpu: add uncore frequency controls

marquiz:
Support uncore frequency limits in the cpu controller. This patch adds
new 'uncoreMinFreq' and 'uncoreMaxFreq' fields to the cpu class
configuration.

Uncore frequency limits are only controllable on the cpu die level in
the HW. When resolving the per-die effective uncore frequency the
controller prefers performance i.e. selects the highest requested value.
For every die it selects the highest value (for both min and max) of all
cpu classes that are effective on that die.

An example configuration with uncore limits:

```
    cpu:
      classes:
        high-prio:
            minFreq: 2000000
            maxFreq: 5000000
            uncoreMinFreq: 2000000
            uncoreMaxFreq: 2400000
        low-prio:
            minFreq: 1000000
            maxFreq: 1500000
            uncoreMinFreq: 1200000
            uncoreMaxFreq: 1800000
```

### DIFF
--- a/pkg/cri/resource-manager/control/cpu/cpu.go
+++ b/pkg/cri/resource-manager/control/cpu/cpu.go
@@ -17,11 +17,14 @@ package cpu
 import (
 	"fmt"
 
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+
 	pkgcfg "github.com/intel/cri-resource-manager/pkg/config"
 	"github.com/intel/cri-resource-manager/pkg/cri/client"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/cache"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/control"
 	logger "github.com/intel/cri-resource-manager/pkg/log"
+	"github.com/intel/cri-resource-manager/pkg/sysfs"
 	"github.com/intel/goresctrl/pkg/utils"
 )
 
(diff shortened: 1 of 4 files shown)


marquiz:
Depends on https://github.com/intel/goresctrl/pull/94

pkg/cri/resource-manager/control/cpu/cpu.go
```diff
@@ -113,36 +127,117 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
+			// Check if this die is affected by the specified cpuset
+			if cpus.Size() == 0 || dieCPUs.Intersection(cpus).Size() > 0 {
+				min, max, minCls, maxCls := effectiveUncoreFreqs(utils.NewIDSet(dieCPUs.ToSlice()...), ctl.config.Classes, assignments)
+
+				if min == 0 && max == 0 {
+					log.Debug("no uncore frequency limits for cpu package/die %d/%d", cpuPkgID, cpuDieID)
+					continue
+				} else if !utils.UncoreFreqAvailable() {
+					return fmt.Errorf("unable to set uncore frequency limits, driver not available, make sure that the intel_uncore_frequency driver is loaded")
+				}
```
jukkar, line 169: Why not check this in the beginning of the function in order to avoid all the processing we have already done at this point?

pkg/cri/resource-manager/control/cpu/cpu.go
```diff
@@ -113,36 +127,117 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
+
+				log.Debug("enforcing uncore min freq to %d (class %q), max freq to %d (class %q) on cpu package/die %d/%d", min, minCls, max, maxCls, cpuPkgID, cpuDieID)
+				if min > 0 {
```
jukkar, line 172: Is it possible that user is able to set min > max

pkg/cri/resource-manager/control/cpu/cpu.go
marquiz: Good question, and I was pondering the failure paths myself. The idea behind this implementation was "fail late" - i.e. the function does not fail if no uncore limits are specified (min and max are both set to zero).

pkg/cri/resource-manager/control/cpu/cpu.go
marquiz: Yes it is. I added a warning log message if this is detected.

marquiz:
Updated, PTAL

askervin (approved):
LGTM.

jukkar (approved):

PR #816, opened 2022-05-27
Add memtier-benchmark-02, a multi-memtier-containers variant.

okartau:
This is variant of memtier benchmark, cloned from existing one, with ability to start multiple memtier containers.


okartau:
I updated the PR with 2 changes as requested, $outfile contains CONTCOUNT and printf format tolerates decimals.
Setting parameter values: yes I agree, lets improve this part, but take it off-line from this PR (new issue for tracking?).
If using such a script for testing, I usually end up changing parameters a lot and hardly can avoid modifying the code files,
although it is cleaner to provide env.vars.
Fixed number of operations: looking to memtier-benchmark options, there is option --requests=NUMBER: total number of requests per client, but I haven't tried it and I don't know does this mean what we want. Again, let's continue with that in another thread.

askervin:
LGTM. Let's continue with the improvements in issues and other PRs. I really like the idea of separating configuration options to a separate file.

askervin (dismissed):

test/e2e/benchmarks.test-suite/memtier_benchmark/n4c16/test02-multi-memtier/code.var.sh
```diff
@@ -0,0 +1,63 @@
...
+NAME=redis-service wait="" create redis-service
+
+# Reset counters in order to keep creating pod0...
+reset counters
+
+benchmark_output_dir="$OUTPUT_DIR/benchmark/multi-memtier"
+mkdir -p "$benchmark_output_dir"
+
+export NAME=memtier-benchmark
+set +x
```
klihub, line 31: nitpick: is this deliberately here, or is it a forgotten leftover from a debugging session ?

klihub (commented):
LGTM, I just have a minor nitpick/question...

test/e2e/benchmarks.test-suite/memtier_benchmark/n4c16/test02-multi-memtier/code.var.sh
klihub: @okartau Do we leave that `set +x` in place ?

okartau:
indeed, set -x was for debug only, removed now, thanks!

askervin (approved):
LGTM

PR #818, opened 2022-05-31
e2e: enable running tests on multi-node clusters

askervin:
If k8smaster environment variable is set, test/e2e/run.sh will join
new vm to an existing cluster instead of creating its own single-node
cluster.


klihub (approved):
LGTM.

Issue #817, opened 2022-05-31
RDT examples configs are outdated


marquiz:
**Describe the bug**

Looks like the RDT example configurations (in `docs/` and `sample-configs/`) are outdated. They follow the pre-goresctrl spec which is slightly different from what is expected.

**Expected behavior**

Example configs should work

**To Reproduce**

**Environment**

**Additional context**

marquiz:
Fixed in #819

PR #819, opened 2022-05-31
docs: update rdt config examples

marquiz:
The examples were pre-goresctrl config format and thus outdated.

Addresses #817


askervin (commented):
s/k3Allocation/l3Allocation/g

marquiz:
> s/k3Allocation/l3Allocation/g

🤦 Thanks, fixed....

docs/policy/rdt.md
```diff
@@ -104,124 +104,52 @@ Besteffort, Besteffort getting only 50% of this. Guaranteed class gets full
...
-        classes:
-          # Burstable gets 100% of what was allocated for the partition
-          Burstable:
-          BestEffort:
-            l3Schema: "50%"
-            # Besteffort gets 50% of the 50% (i.e. 25% of total) or 1000MBps
-            mbSchema: ["50%", "1000MBps"]
-```
+configure cache allocations exactly as required. For detailed description of the RDT configuration format with examples see the
+[goresctrl library documentation](https://github.com/intel/goresctrl/blob/v0.2.0/doc/rdt.md)
```
jukkar, line 152: Why not point the link to latest version `https://github.com/intel/goresctrl/blob/main/doc/rdt.md`, that way this documentation is always "correct" one and we do not need to adjust it when new goresctrl version are created.

marquiz: Good question. I was thinking about this but decided that we better refer to the version that we actually depend on. Future versions of goresctrl may have new features that we don't support (or even have incompatible changes, who knows) which would confuse users. I now added a patch that specifies the goresctrl version in sphinx config which hopefully makes the maintenance pain little smaller.

askervin (approved):
LGTM. Nice goresctrl version replacement. :)

PR #821, opened 2022-06-02
resmgr: disallow talking to untested runtimes by default.

klihub:
Disallow talking to/proxying for unknown runtimes, unless explicitly told to do so on the command line. This could save us quite a bit of time from chasing ghosts.

Inspired by an idea from @jukkar over a lunch discussion. Thanks !

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -341,36 +341,11 @@ Starting from scratch:
    * [Cgroup and Kata containers](https://github.com/kata-containers/kata-containers/blob/stable-2.0.0/docs/design/host-cgroups.md)
 
 
-## Using Docker\* as the runtime
+## Running with Untested Runtimes
 
-If you must use `docker` as the runtime then the proxying setup is slightly
-more complex. Docker does not natively support the CRI API. Normally kubelet
-runs an internal protocol translator, `dockershim` to translate between CRI
-and the native docker API. To let CRI Resource Manager effectively proxy
-between kubelet and `docker` it needs to actually proxy between kubelet and
-`dockershim`. For this to be possible, you need to run two instances of
-kubelet:
-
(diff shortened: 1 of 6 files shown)


marquiz (dismissed):
Heh 👌 
/lgtm

askervin:
A bit mixed feelings about this... a warning would have been easier for me, but at the same time I know that nobody would notice the warning, except for us when we are after a ghost already.

And on the other hand the scenario where we should to take back this change and allow docker by default, seems to be highly unlikely, says the crystal ball.

As a downside, we will need to add a couple of sentences about docker in our documentation, yet getting rid of it completely would be even nicer.

So after all, I suppose this ok.

klihub:
> A bit mixed feelings about this... a warning would have been easier for me, but at the same time I know that nobody would notice the warning, except for us when we are after a ghost already.

There are mainly two reasons I chose to go with this approach:
1. What you also state above: nobody seems to read logs/warnings and if they do, they don't seem to pay attention.
2. We really want docker-usage to be opt-in. It should not work out of the box.
 
> And on the other hand the scenario where we should to take back this change and allow docker by default, seems to be highly unlikely, says the crystal ball.


 
> As a downside, we will need to add a couple of sentences about docker in our documentation, yet getting rid of it completely would be even nicer.

I think we should remove all documentation reference to docker. And add a single paragraph stating that:
1. Using docker is highly discouraged.
2. People should not expect feature parity with containerd/crio when using docker.
3. --allow-docker must be passed on the command line to cri-resmgr, otherwise it will refuse to talk to docker

> So after all, I suppose this is ok.

I added a mention of the necessary command line option to the rejection error message.

jukkar (approved):

PR #822, opened 2022-06-06
e2e: add support for distro=ubuntu-22.04

askervin:


askervin:
This works out-of-the-box with `k8scri=containerd`. However, cilium DaemonSet ends up in crash loop with `k8scri=crio crio_src=/path/to/current-main` (tested with cri-o main `afab4b78`, cilium 1.10.11/cilium image 1.11.5, k8s v1.24.1 built with go 1.18.2).

PR #824, opened 2022-06-07
Add NRI support to e2e scripts

jukkar:
Make sure that e2e tests can be run with NRI support enabled in cri-rm and containerd. As of this writing, support with cri-o is not tested. This PR is still WIP.
Some tests report error randomly, it seems that if cri-resmgr exists (for whatever reason), it is not restarted, because it is not run under systemd in e2e tests, which then leads to test failures.


jukkar:
Replaced by #901

PR #825, opened 2022-06-07
topology-aware: implement pod/namespace co-location using implicit affinities.

klihub:
Implement implicit affinities for pod and namespace-based container co-location. If enabled, the topology-aware policy will inject affinity expressions which should cause containers in the same pod or in the same namespace to have affinity for each other. If a container has explicit affinities annotated by the user these affinities are not injected. Both of these annotations are opt-in. They are behind boolean TA policy configuration flags which default to false.

In preparation for these functional additions, the cache-level implicit affinity infra has been slightly reworked to allow dynamic injection of implicit affinities tailored per container.

The PR also updates the documentation, describing the recently added TA policy options.


docs/policy/topology-aware.md
```diff
@@ -300,6 +306,21 @@ Topology hint generation is globally enabled by default. Therefore, using the
 Pod annotation as opt in only has an effect when the whole pod is annotated to
 opt out from hint-aware pool selection.
 
+### Topological Co-location for Pods and Namespaces
+
+The `ColocatePods` or `ColocateNamespaces` configuration options control whether
+the policy will try to co-locate, IOW allocate topologically close to each other
+containers within the same Pod or K8s namespace. Both of these options are opt-in,
+they are off by default. The actual co-location/allocation strategy is implemented
```
askervin, line 314: Suggesting simplifying, switching from implementation to user point of view, and offering direct configuration words "true"/"false" instead of "opt-in"/"off":

Original:
_Both of these options are opt-in, they are off by default. The actual co-location/allocation strategy is implemented by simply injecting implicit affinities to containers which cause them to have affinity for all other containers in the same Pod or namespace._

_Any container that has explicitly annotated affinities by the user will be excluded from annotation injection by these options. Note that since annotations injected to other containers can still be evaluated against these containers, extra care must be taken if explicit annotations are also used while either or both of these options are enabled. Both of these implicitly injected affinities use a weight of 10._

A suggestion:
Both of these options are `false` by default. Setting to `true` is a shorthand for adding affinities of weight 10 to containers in same pods or namespaces.

Containers with user-defined affinities are not extended with colocation affinities. However, colocation affinities of other containers affect the placement of these containers, too. Therefore mixing user-defined and colocation affinities requires carefully thought on affinity weights.

klihub:
@kad @marquiz @jukkar @askervin While testing these I ran a few times into surprising/unexpected behavior.

As earlier, these were caused by the intuitively best fitting pool not having enough resources, and an alternative pool getting picked based on its distance-diluted affinity... and that choice was not immediately obvious.

I have the feeling that we will need to do something about this eventually... we just need to figure out what first. I think that
- at least dilution by distance should be controllable per affinity (IOW possible to override in the affinity annotation)
- perhaps the dilution model should be configurable (which is a generalization of the above idea)
- calculating a single effective affinity as a linear combination of all affinities might not be the best idea

askervin (dismissed):
LGTM. Great stuff!

The only thing that I'm not quite sure abut is: do we gain anything by leaving out colocation affinities in containers with user-defined affinities. Or does exception only add complexity without solving any problem? Adding colocation affinities in the same way to all containers in same pods/namespaces might make estimation of total affection weights easier (yet this may be an illusion).

Added a documentation suggestion, but I'm not quite sure if it is reasonably accurate to call colocation flags as a shorthand for affinities.

docs/policy/topology-aware.md
klihub: > _ ... Note that since annotations injected to other containers can still be evaluated against these containers, extra care must be taken if explicit annotations are also used while either or both of these options are enabled. Both of these implicitly injected affinities use a weight of 10._
> 
> A suggestion: ... However, colocation affinities of other containers affect the placement of these containers, too. Therefore mixing user-defined and colocation affinities requires carefully thought on affinity weights.

Point taken, I'll try to rephrase. But can't do it like that, because that statement is very different from the original one, and also false.

klihub: Rephrased and updated.

docs/policy/topology-aware.md
```diff
@@ -300,6 +306,21 @@ Topology hint generation is globally enabled by default. Therefore, using the
 Pod annotation as opt in only has an effect when the whole pod is annotated to
 opt out from hint-aware pool selection.
 
+### Topological Co-location for Pods and Namespaces
+
+The `ColocatePods` or `ColocateNamespaces` configuration options control whether
+the policy will try to co-locate, IOW allocate topologically close to each other
```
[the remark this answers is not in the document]
klihub: Updated.

jukkar (approved):
LGTM

askervin (approved):
Thanks! LGTM.

PR #826, opened 2022-06-08
e2e: update ./run.sh debug to work with recent golang and dlv

askervin:
- Use "go install" instead of "go get".
- Update dlv configuration file syntax to work with latest dlv.
- Now crirm_src=/path/to/project vm=vm-name ./run.sh debug
  works again: copies sources to vm and sets up dlv for debugging.


klihub (approved):
LGTM

PR #834, opened 2022-06-15
fixes: substitute golang version from go.mod.

klihub:


jukkar (dismissed):
LGTM

marquiz (approved):
I think we can merge this

Just curious why not use
```
--build-arg ARG GOLANG_VERSION=$(GOLANG_VERSION)
```
instead of generating dockerfiles (with yet another make target)?

PR #831, opened 2022-06-10
e2e: add test for dynamic balloons configuration update

askervin:
- Add new test that changes balloons and CPU configurations using a
  config map.
- Refactor verify-metrics functions to a common source file.

PR #837, opened 2022-06-16
test/e2e: clean up after namespace colocation tests.

klihub:
Properly clean up pods and restore default configuration after having run the namespace-colocation tests.


jukkar (approved):

askervin (approved):
LGTM.

In similar situations I've also used a [cleanup function](https://github.com/intel/cri-resource-manager/pull/831/files#diff-51d000da56949c62b43aeacf79fcb109967130c212dfe8862bcda402fef0b05fR6) that is executed both in the beginning and at the end of the test. The benefit is that it has been easier to rerun a failing test when the next run can recover from the mess that a failed test left behind. The downside is that there may be (ignored) error messages in the test output when there's nothing to clean.

marquiz (approved):
lgtm

PR #840, opened 2022-06-16
test/e2e: don't use invalid '0' CPU reservation.

klihub:
Don't use invalid 'cpu: 0' CPU reservation. Don't even use a test-specific configuration file.
Instead, parameterize the top-level one correctly for the test case.


askervin (approved):
LGTM.

marquiz (approved):
LGTM

PR #841, opened 2022-06-17
e2e: ambigious allowed resources for each name is an error

askervin:
Previous behavior produced an union of allowed resources if there were
several containers running with the same name. This
feature/consequence was not used, but it produced unnecessary trouble
in debugging tests that fail due to unclean test environment.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -227,6 +227,11 @@ get-py-allowed() {
         command-error "error fetching res_allowed from $VM_NAME"
     }
     echo -e "$COMMAND_OUTPUT" > "$res_allowed_file"
+    # Validate res_allowed_file. Error out if there is same container
+    # name with two different sets of allowed CPUs or memories.
+    awk -F "[ /]" '{if (pod[$1]!=0 && pod[$1]!=""$3""$4){print "error: ambiguous allowed resources for name "$1; exit(1)};pod[$1]=""$3""$4}' "$res_allowed_file" || {
+        error "container/process name collision: test environment needs cleanup."
+    }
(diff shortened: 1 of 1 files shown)


jukkar (approved):

klihub:
Maybe we could add an alternative collection method to `topology.py`, for instance `kubectl_res_allowed` which would essentially collect from the output of this:
```
for nspod in $(kubectl get pods --all-namespaces | tr -s '\t ' | grep -v ^NAMESPACE |
    grep ' pod[0-9]*' | cut -d ' ' -f1,2 | tr -s ' ' '/'); do
    ns=${nspod%/*}
    pod=${nspod#*/}
    for ctr in $(kubectl get pod -n $ns $pod -o json | jq '.spec.containers[].name'| tr -d '"'); do
       kubectl exec -n $ns -c $ctr $pod -- cat /proc/self/status | grep _allowed: |
          tr -s '\t ' ' ' | tr -s '\n' ' ' | sed "s#^#$ns/$pod:$ctr #g;s/Cpus_allowed: /c:/g;s/Mems_allowed: /m:/g"
    done
done
```

If the caller knows the prereqs are fullfilled then it could use this to collect the resources, which then would be provided with the almost exactly same output, but collected per namespace+pod+container (prefixed with `$namespace/$pod:$container` instead of `$container/$pid`). This requires `/bin/sh` in the containers which we practically do have 100% of the time in our tests, and `jq` on the host which is behind a single `dnf|apt-get|zypper install jq`.

marquiz (approved):
lgtm

klihub (approved):
LGTM

PR #830, opened 2022-06-10
balloons: improve dynamic configuration update

askervin:
- Recreate balloons and reassign workloads only if necessary.
- CPU class of balloon types can be switched without reassigning
  workloads.

PR #842, opened 2022-06-17
e2e: fix balloons test03-reserved

askervin:
Add explicit CPU and memory pinning.

### DIFF
--- a/test/e2e/policies.test-suite/balloons/n4c16/test03-reserved/balloons-reserved.cfg
+++ b/test/e2e/policies.test-suite/balloons/n4c16/test03-reserved/balloons-reserved.cfg
@@ -3,6 +3,8 @@ policy:
   ReservedResources:
     CPU: cpuset:0-2
   balloons:
+    PinCPU: true
+    PinMemory: true
     IdleCPUClass: idle-class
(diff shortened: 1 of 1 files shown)


test/e2e/policies.test-suite/balloons/n4c16/test03-reserved/balloons-reserved.cfg
```diff
@@ -3,6 +3,8 @@ policy:
   ReservedResources:
     CPU: cpuset:0-2
   balloons:
+    PinCPU: true
+    PinMemory: true
```
klihub, line 7: Can I ask why these are not the defaults ? I mean, isn't this exactly what the user us expected to use in 99% of the cases ? If it is, I think these should be opt-out, not opt-in.

marquiz: Hmm, yes, good question @klihub, that same question came to my mind as well 😎

askervin:
PR #830 counts undefined PinCPU and PinMemory as true. I'll close this PR because without this change test03 can be used for testing the undefined case.

PR #845, opened 2022-06-20
webhook: update example deployment node selector and toleration

askervin:
Node role "master" has been changed to "control-plane" in Kubernetes.

### DIFF
--- a/cmd/cri-resmgr-webhook/webhook-deployment.yaml
+++ b/cmd/cri-resmgr-webhook/webhook-deployment.yaml
@@ -57,9 +57,9 @@ spec:
           periodSeconds: 30
 
       nodeSelector:
-        node-role.kubernetes.io/master: ""
+        node-role.kubernetes.io/control-plane: ""
       tolerations:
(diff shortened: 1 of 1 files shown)


marquiz (commented):
Hmm, I there are still a lot of pre-1.24 clusters with the old label and taint name. Maybe we could/should support both with smth like

```yaml
      nodeAffinity:
        preferredDuringSchedulingIgnoredDuringExecution:
        - weight: 1
          preference:
            matchExpressions:
            - key: "node-role.kubernetes.io/master"
              operator: In
              values: [""]
        - weight: 1
          preference:
            matchExpressions:
            - key: node-role.kubernetes.io/control-plane
              operator: In
              values: [""]

      tolerations:
        - key: "node-role.kubernetes.io/master"
          operator: "Equal"
          value: ""
          effect: "NoSchedule"
        - key: "node-role.kubernetes.io/control-plane"
          operator: "Equal"
          value: ""
          effect: "NoSchedule"
```

klihub (approved):

PR #851, opened 2022-06-21
cri-resmgr: fix cgroup v2 paths in available resources

askervin:


askervin:
This issue was found when running e2e tests with Ubuntu 22.04.

pkg/cri/resource-manager/policy/flags.go
```diff
@@ -163,11 +165,28 @@ func (cs *ConstraintSet) parseCPUQuantity(value string) error {
...
+		// dir is a relative path starting from the cgroup mount point
+		path = ifExists(filepath.Join(cgroups.GetMountDir(), dir, "cpuset.cpus"))
+	}
+	if path == "" {
+		// dir is a relative path starting from the cpuset controller (cgroup v1)
+		path = ifExists(filepath.Join(cgroups.Cpuset.Path(), dir, "cpuset.cpus"))
+	}
+	if path == "" {
+		return policyError("failed to find cgroup with cpuset.cpus for CPU cpuset constraint %q", dir)
 	}
```
klihub, line 189: Looks basically fine to me.
nitpick: Why not an expressionless switch with equivalent cases than an effective if - elif - elif - elif - else -construct.

pkg/cri/resource-manager/policy/flags.go
```diff
@@ -163,11 +165,28 @@ func (cs *ConstraintSet) parseCPUQuantity(value string) error {
...
-	if !strings.HasPrefix(dir, cpusetDir+"/") {
-		dir = filepath.Join(cpusetDir, dir)
+	ifExists := func(path string) string {
+		if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
+			return ""
+		}
+		return path
+	}
+	path := ""
+	if dir[0] == '/' {
```
klihub, line 175: BTW, I think crashes if called with `dir := ""`...

askervin: That crash happens already on `parseCPU()`. I think it makes to fix that in this same commit. Working on it.

askervin: Fixed in both places. (Here it is dead code, but good to have if later on called with an empty string.)

pkg/cri/resource-manager/policy/flags.go
askervin: Switched to `switch`.

jukkar: Sorry for my ignorance but is the switch working the same way now? I mean the original code tries to figure out the path multiple ways, but the switch is now doing things only once.

klihub (approved):
LGTM

PR #854, opened 2022-06-21
build: build binary tarball from static binaries.

klihub:
Add targets for building static binaries and cleaning up built binaries. Change `binary-dist` to build tarball from static binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,13 +241,16 @@ all: build update-workflows
 
 build: $(BUILD_BINS)
 
+build-static:
+	$(MAKE) STATIC=1 build
+
 install: $(BUILD_BINS) $(foreach dir,$(BUILD_DIRS),install-bin-$(dir)) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -241,13 +241,16 @@ all: build update-workflows
 
 build: $(BUILD_BINS)
 
+build-static: clean-bin
```
marquiz, line 244: Should we add the `clean-bin` dep to `build:` instead? Just thought about altering static and non-static builds...

klihub: Hmm, yes. Maybe that's a better idea.

klihub: Okay, I updated it so, that binaries are rebuilt if
1. they are out of date, or
2. existing binaries were built with a different STATIC setting than the current one

askervin (approved):
LGTM

marquiz (approved):
lgtm

PR #867, opened 2022-08-29
Makefile: don't change dir when running go build

marquiz:
This fixes build of agent and webhook images. Otherwise they will fail
because go-build is not able to find the vcs information. Could be
handled with '-buildvcs=false' build flag but just removing chdir seems
simpler. Plus, we get the vcs information in the go binary.


klihub (approved):
If this works for all our binaries, then LGTM.

jukkar (approved):

PR #869, opened 2022-08-29
[release 0.7] backport image building fixes

marquiz:
- images: build statically linked binaries
  We use scratch as the base image now so now dynamically loaded libs are
  available.
- Makefile: don't change dir when running go build
  This fixes build of agent and webhook images. Otherwise they will fail
  because go-build is not able to find the vcs information. Could be
  handled with '-buildvcs=false' build flag but just removing chdir seems
  simpler. Plus, we get the vcs information in the go binary.


klihub (approved):
LGTM.

PR #871, opened 2022-08-30
Makefile: use statically linked binaries in e2e-tests

marquiz:
Otherwise tests might fail because of library version incompatibilities
wrt build host vs the container image that we're running the binaries.


askervin (approved):
LGTM

klihub (approved):
LGTM.

PR #875, opened 2022-08-31
Misc fixes to e2e tests

jukkar:
Small and minor tweaks to e2e tests. Noticed these when running e2e tests inside a VM.

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -397,8 +397,8 @@ vm-monitor() { # script API
 vm-wait-process() { # script API
     # Usage: vm-wait-process [--timeout TIMEOUT] [--pidfile PIDFILE] PROCESS
     #
-    # Wait for a PROCESS (string) to appear in process list (ps -A output).
-    # If pidfile parameter is given, we also check that the process has the file open.
+    # Wait for a PROCESS (string) to appear in process list (pidof output).
+    # If pidfile parameter is given, we also check that the process has that file open.
     # The default TIMEOUT is 30 seconds.
     local process timeout pidfile invalid
     timeout=30
@@ -423,7 +423,7 @@ vm-wait-process() { # script API
         return 1
     fi
     process="$1"
-    vm-run-until --timeout "$timeout" "ps -A | grep -q \"$process\"" || error "timeout while waiting $process"
+    vm-run-until --timeout "$timeout" "pidof -q \"$process\"" || error "timeout while waiting $process"
 
     # As we first wait for the process, and then wait for the pidfile (if enabled)
     # we might wait longer than expected. Accept that anomaly atm.
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -997,7 +997,7 @@ vm-create-singlenode-cluster() {
     vm-command "kubectl taint nodes --all node-role.kubernetes.io/control-plane-"
     vm-command "kubectl taint nodes --all node-role.kubernetes.io/master-"
     vm-install-cni-"$(distro-k8s-cni)"
-    if ! vm-command "kubectl wait --for=condition=Ready node/\$(hostname) --timeout=120s"; then
+    if ! vm-command "kubectl wait --for=condition=Ready node/\$(hostname) --timeout=240s"; then
         command-error "kubectl waiting for node readiness timed out"
     fi
 }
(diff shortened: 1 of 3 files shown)


askervin (dismissed):
LGTM

demo/lib/vm.bash
```diff
@@ -174,6 +174,49 @@ vm-check-running-binary() {
...
+    done
+    if [ -n "$invalid" ]; then
+        error "invalid options: $invalid"
+        return 1
+    fi
+    process="$1"
+
+    # As we first wait for the process, and then wait for the pidfile (if enabled)
+    # we might wait longer than expected. Accept that anomaly atm.
+    vm-wait-process --timeout $timeout "$process" || command-error "timeout while waiting $process"
```
askervin, line 210: Use `error` rather than `command-error` in all occurrences in this function.

`command-error` works only for reporting an error from previous command execution that is visible to the user, that is, command execution that is wrapped with `command-start` and `command-end`. As a rule of thumb, it works immediately after `vm-command`, but elsewhere almost certainly `error` is a better choice.

demo/lib/vm.bash
```diff
@@ -174,6 +174,49 @@ vm-check-running-binary() {
...
+    # The default TIMEOUT is 30 seconds.
+
+    local process timeout pidfile invalid
+
+    timeout=30
+    while [ "${1#-}" != "$1" ] && [ -n "$1" ]; do
+        case "$1" in
+            --timeout)
+                timeout="$2"
+                shift; shift
```
klihub, line 190: annoying besserwisser nitpick: `shift 2` (we do require and only use bash) ?

demo/lib/vm.bash
```diff
@@ -174,6 +174,49 @@ vm-check-running-binary() {
...
+                ;;
+            --pidfile)
+                pidfile="$2"
+                shift; shift
```
klihub, line 194: ditto here...

klihub (approved):
LGTM.

demo/lib/vm.bash
jukkar: Good point, will change.

demo/lib/vm.bash
```diff
@@ -174,6 +174,49 @@ vm-check-running-binary() {
     return 0
 }
 
+vm-wait-process-startup() {
```
askervin, line 177: `vm-wait-process-startup [--timeout X]` seems to do the same as `vm-wait-process [--timeout X]`. What do you think, would it make sense to add `--pidfile` directly to `vm-wait-process` instead of adding a wrapper with the parameter?

But if there are reasons to have two separate functions, this is ok, too.

test/e2e/policies.test-suite/topology-aware/n4c16/test09-container-exit/code.var.sh
```diff
@@ -9,7 +9,8 @@ pyexec 'assert "pod0c0" in allocations'
 out '### Crash and restart pod0c0'
 vm-command "kubectl get pods pod0"
 vm-command "kill -KILL \$(pgrep -f pod0c0)"
-sleep 2
+vm-command 'kubectl wait --for=condition=Ready pods/pod0'
```
askervin, line 12: In my VM, `kubectl wait --for=condition=Ready pods/pod0` seems to return with `pod/pod0 condition met` right after killing the process, because Kubernetes has not yet noticed that there is something wrong with the pod. Maybe a sleep would be useful after all to make this wait do something. Or just drop the wait.

demo/lib/vm.bash
jukkar: Indeed, those two could be combined as the only difference is the pidfile option. I will change this.

test/e2e/policies.test-suite/topology-aware/n4c16/test09-container-exit/code.var.sh
jukkar: Hmm, ok. I did not notice any difference whether the sleep was there or not, but could be because of difference in host/VM combination. I can add the sleep back just in case. The wait-for is nice to have there too, if the startup takes more than 2sec.

askervin (approved):
LGTM. Thanks Jukka, let's get this in.

PR #876, opened 2022-09-02
jenkinsfile: build *:devel images from master branch

marquiz:
So far we've only built container images from tags. This patch adds
special handling for the master branch so that container images built
from it are tagged with *:devel tag (instead of *:master). This naming
matches what we have in documentation, for example.


klihub (approved):
LGTM... with a warning about my Jenkins knowledge being zilch.

marquiz:
> with a warning about my Jenkins knowledge being zilch.

Don't worry, I'm not much more proficient in this field. Or maybe you should worry...

PR #879, opened 2022-09-02
e2e: add host-mount-vm helper

askervin:
`host-mount-vm` uses sshfs to mount test vm file system with root access under `~/vms/data/VMNAME/sshfs`. This enables using Your Favourite Editor as normal user on host for accessing files on the test vm, making testing and development more convenient.

The reason why I considered a separate helper function to be useful, instead of just using sshfs without helpers, is that the root access part is slightly tricky.


klihub (approved):
Cool ! LGTM.

PR #880, opened 2022-09-02
dockerfiles: rename suse 15.2 to 15.4

marquiz:
Rename suse 15.2 (openSUSE Leap 15.2) dockerfile to 15.4 and
specifically set builder image to use opensuse/leap:15.4 tag. Previously
we didn't specify any version of the builder image so it was always
simply using the latest release, not actually matching the name of the
dockerfile or the package that was build.

Also, drop suse 15.1 dockerfile as it was virtually identical to the
15.2 dockerfile, also building on the latest openSUSE Leap release (and
not 15.1 as it suggested).


klihub (approved):
LGTM.

askervin (approved):
LGTM.

PR #881, opened 2022-09-02
resmgr: allow disabling the agent ingress and egress interfaces.

klihub:
This patch set allows one to disable the incoming and outgoing config/agent interfaces by explicitly setting the corresponding command line options to `"disabled"` or `""`.

### DIFF
--- a/pkg/cri/resource-manager/agent/agent.go
+++ b/pkg/cri/resource-manager/agent/agent.go
@@ -31,8 +31,14 @@ import (
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/config"
 )
 
+const (
+	SocketDisabled = "disabled"
+)
+
 // Interface describe interfaces of cri-resource-manager agent
 type Interface interface {
+	IsDisabled() bool
+
 	GetNode(time.Duration) (core_v1.Node, error)
 	PatchNode([]*agent_v1.JsonPatch, time.Duration) error
 	UpdateNodeCapacity(map[string]string, time.Duration) error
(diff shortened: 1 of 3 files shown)


askervin:
`cri-resmgr -agent-socket disabled` produces less noise. Great!

LGTM.

askervin (approved):

jukkar (dismissed):

pkg/cri/resource-manager/agent/agent.go
```diff
@@ -55,13 +61,20 @@ type Interface interface {
...
 // and return a new Interface
 func NewAgentInterface(socket string) (Interface, error) {
-	a := &agentInterface{}
+	a := &agentInterface{
+		socket: socket,
+	}
+
+	if a.IsDisabled() {
+		return a, nil
+	}
```
marquiz, line 77: Just thinking aloud: my first thought was that should we even call `NewAgentInterface` it it's disabled 🤔 That would of course mean doing `if agent != nil` checks elsewhere in the code base (instead of `IsDisabled()` checks in the agentinterface methods). Matybe it's safer this way.

Another point in this pondering is that I don't think the static pools policy works without agent(?) And it prolly shouldn't start at all 🧐

klihub: I wanted to make this as little intrusive as possible, hence the IsDisabled() checks.

klihub: Updated `static-pools` to bail out (also) if agent is disabled.

jukkar (approved):

PR #865, opened 2022-08-19
balloons: add MaxBalloons configuration option

askervin:

PR #882, opened 2022-09-02
balloons: require MaxCPUs in all balloon types, allow setting it to "unlimited"

askervin:
Ignore before #865 is merged.

This draft build on top of PR #865. Add explicit "unlimited" to both MaxCPUs and MaxBalloons.

Needs testing "unlimited" CPU balloon inflation and MaxBalloons behavior.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
@@ -122,8 +122,11 @@ func (bln Balloon) AvailMilliCpus() int {
 	return bln.Cpus.Size() * 1000
 }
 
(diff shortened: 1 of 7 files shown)


marquiz (commented):
I'm a bit unsure, do we really want this? Somehow I draw a parallel to k8s resource limits: if nothing is specified it means unlimited.

If we really do this we must support "string-or-int" style unmarshalling so that you can specify an integer in the configuration (not just a string). So that the user does NOT need to specify integer limits like this (mincpus and maxcpus formatted differently):

```
minCpus: 2
maxCpus: "4"
```

askervin:
Definitely, defining nothing as a limit must equal to unlimited. Abandoning this PR.

PR #884, opened 2022-09-06
test/e2e/run.sh: add qemu_cpumem='...' option.

klihub:
Add $qemu_cpumem option which can be used to directly specify
qemu CPU and memory configuration options for the VM to being
created. It is mutually exclusive with $topology.

Among other things this new option can be used to bootstrap
a KVM-enabled test VM (if nested virtualization is enabled on
the host), for intstance using this commnand line:

```
vm=kvm-test topology='' qemu_cpumem='-cpu Haswell-noTSX-IBRS,vme=on,ss=on,vmx=on,pdcm=on,f16c=on,rdrand=on,hypervisor=on,arat=on,tsc-adjust=on,umip=on,md-clear=on,stibp=on,arch-capabilities=on,ssbd=on,xsaveopt=on,pdpe1gb=on,abm=on,ibpb=on,ibrs=on,amd-stibp=on,amd-ssbd=on,skip-l1dfl-vmentry=on,pschange-mc-no=on -smp cpus=8 -m 8G' distro=fedora k8scri='containerd' ./test/e2e/run.sh interactive
```

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -98,6 +98,9 @@ usage() {
     echo "  VM configuration VARs: (effective when vm is not already configured)"
     echo "    topology: JSON to override NUMA node list used in tests."
     echo "             See: python3 ${DEMO_LIB_DIR}/topology2qemuopts.py --help"
+    echo "    qemu_cpumem: direct CPU and memory parameters to pass to qemu."
+    echo "             Should include at least -m MEMORY and -smp CPUCORES."
+    echo "             This parameter is mutually exclusive with topology."
     echo "    distro:  Linux distribution to be / already installed on vm."
     echo "             Supported values: centos-7, centos-8, debian-10, debian-sid"
     echo "                 fedora, fedora-33, opensuse-tumbleweed,"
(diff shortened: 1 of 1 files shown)


jukkar:
Can't we use the `VM_QEMU_EXTRA` env for this?

jukkar (approved):

askervin:
Every now and then I have needed to add `-cpu` parameter to Qemu, but still have been using the topology configuration. Example:
```
VM_QEMU_EXTRA='-cpu host,topoext=on' \
vm=mydebian \
distro=debian-sid \
on_vm_online='host-mount-vm;exit' \
topology='[{"cores":4,"mem":"8G"}]' \
cri-resource-manager/test/e2e/run.sh interactive
```

Now that this patch makes Qemu CPU (and memory) options a documented parameter (which is great!), I'm wondering if the same parameter could solve the use case "configure both `topology` and `-cpu`"  Or at minimum allow solving this use case later on without being too confusing to the user.

There would be at least two options that I can think of:
- If `qemu_cpumem` has only `-cpu` and not `-m` or `-smp`, then `topology` would be effective. If `qemu_cpumem` has either `-m` or `-smp`, it must have both of them and `topology` must be empty, otherwise this is an error.
- Add another parameter, `qemu_cpu`, that does not affect `topology` handling, and that must not include `-m` or `-smp`.

What do you think would be least confusing to users?

klihub:
> Every now and then I have needed to add `-cpu` parameter to Qemu, but still have been using the topology configuration. Example:
> 
> ```
> VM_QEMU_EXTRA='-cpu host,topoext=on' \
> vm=mydebian \
> distro=debian-sid \
> on_vm_online='host-mount-vm;exit' \
> topology='[{"cores":4,"mem":"8G"}]' \
> cri-resource-manager/test/e2e/run.sh interactive
> ```
> 
> Now that this patch makes Qemu CPU (and memory) options a documented parameter (which is great!), I'm wondering if the same parameter could solve the use case "configure both `topology` and `-cpu`" Or at minimum allow solving this use case later on without being too confusing to the user.
> 
> There would be at least two options that I can think of:
> 
> * If `qemu_cpumem` has only `-cpu` and not `-m` or `-smp`, then `topology` would be effective. If `qemu_cpumem` has either `-m` or `-smp`, it must have both of them and `topology` must be empty, otherwise this is an error.
> * Add another parameter, `qemu_cpu`, that does not affect `topology` handling, and that must not include `-m` or `-smp`.
> 
> What do you think would be least confusing to users?

Well, I just needed a quick way to set up a VM with exact CPU parameters so that I can test KVM/kata containers using nested virtualization. I only looked at VM_QEMU_CPUMEM to override the defaults, but that gets unconditionally overridden 
by topology, even the implicit/default one.

I simply opted for adding a really-big-hammer option for cases where the user know exactly what the desired qemu options are and made the script politely get out of the way when that option is used. TBH, I never tried passing the same options using VM_QEMU_EXTRA. If that works then this PR is not really necessary as the same can be achieved using that parameter.

jukkar:
Hmm, just pondering here. If you use VM_QEMU_EXTRA and not set VM_QEMU_CPUMEM, wouldn't there be conflicting values somewhere as there are default values set to VM_QEMU_CPUMEM in vm.bash?

PR #886, opened 2022-09-07
release-0.6: cache: don't crash if pod has no status data.

klihub:


jukkar (approved):

askervin (approved):
LGTM.

PR #887, opened 2022-09-07
master: cache: don't crash if pod has no status data.

klihub:

This PR is #886 cherry-picker to master.


jukkar (approved):

askervin (approved):
LGTM.

marquiz:
Shouldn't this be backported to `release-0.7` branch, too?

PR #889, opened 2022-09-08
CODEOWNERS: update maintainers.

klihub:
Remove those previous owners who are nowadays mostly guilty by association.


kad (dismissed):
fine by me, but I think we can keep @mythi  as backup for some emergency cases.

klihub:
> fine by me, but I think we can keep @mythi as backup for some emergency cases.

Good point. I added him back to the list.

jukkar (approved):

PR #892, opened 2022-09-08
release-0.7: cache: don't crash if pod has no status data.

klihub:

This PR is #886 cherry-picked to release-0.7.


askervin (approved):
LGTM

jukkar (approved):

PR #896, opened 2022-09-29
e2e: add .internal and 192.168.0.0/12 to no_proxy

askervin:
Extending no_proxy fixes a problem of running minikube in a Ubuntu VM behind a proxy.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -923,7 +923,7 @@ default-setup-proxies() {
         return 0
     fi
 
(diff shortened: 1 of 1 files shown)


jukkar (approved):

askervin:
> btw, correct "martians" that can be used for local networks are `10.0.0.0/8`, `172.16.0.0/12` and `192.168.0.0/16`. /12 netmask is incorrect.

Luckily the typo was only in the commit message topic. Looks very stupid in the history but does not really break anything...

PR #897, opened 2022-09-29
e2e: add vm-install-minikube and vm-create-minikube-cluster

askervin:
- Tested on debian, fedora and ubuntu.
- Example: distro=fedora on_vm_online=' vm-install-minikube; vm-create-minikube-cluster; vm-command "minikube kubectl -- get pods -A"; exit' test/e2e/run.sh interactive

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -1,7 +1,9 @@
 # shellcheck disable=SC2120
 GO_URLDIR=https://golang.org/dl
-GO_VERSION=1.16.8
+GO_VERSION=1.19.1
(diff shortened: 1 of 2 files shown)


askervin:
This needs PR #896 if used behind a http_proxy without the default no_proxy including `.internal` and other no_proxy extensions in the PR.

askervin:
> Could some reasoning mentioned in the commit message that why we are adding minikube support?

Good point. Reasoning added.

klihub (approved):
If we really need this then LGTM...

PR #888, opened 2022-09-07
draft: experimental minimal NRI plugin conversion.

klihub:
This draft PR adds a new command line option `--use-nri-plugin`. When passed to `cri-resmgr` this option disables CRI request proxying and enables experimental code for resource management as an NRI plugin. Note that for testing this in practice, you'll need a version of either `containerd` or `cri-o` patched with NRI support for testing.

PR #898, opened 2022-10-04
e2e tests using Github self hosted runner

jukkar:


klihub:
Closing, same reason as #888. SHR primary focus is [NRI plugins collection](https://github.com/jukkar/nri-resmgr).

PR #899, opened 2022-10-07
Add -check-config option to validate config.

ppalucki:
Exists with proper exit code to (0-ok) and prints error message.

Can be used as a check before applying new configuration.

Just an idea - probably in our extension we will programatically API of config module - more about it here: https://github.com/intel/gardener-extension-cri-resmgr/issues/56

BTW: I found undocumented config-help/help arguments to are usefull - are they undocumented on purpose or just missed? 

Output looks like:
```
# go run ./cmd/cri-resmgr/ -check-config sample-configs/balloons-policy.cfg
...
config ok
# echo $?
0
```

or
```
# go run ./cmd/cri-resmgr/ -check-config broken.cfg
validation error:  config error: module logger: given unknown configuration data Debuug
# echo $?
1
```
with broken.cfg:
```
logger:
    Debuug: "foo"
```

Marking it as draft because I don't how usefull it would be - probably it just checks syntax (proper keys and types) but not sematics e.g. ? (cpuset is parsed for example, but I can set **policy.Active to non-existents and still returns ok!!!**)

e.g. non-existing-policy.cfg
```
#cat non-existing-policy.cfg
policy:
    Active: "ThereIsNoSuchPolicy"

# go run ./cmd/cri-resmgr/ -check-config non-existing-policy.cfg
config ok
```

### DIFF
--- a/cmd/cri-resmgr/main.go
+++ b/cmd/cri-resmgr/main.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/intel/goresctrl/pkg/rdt"
 
-	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager"
+	resmgr "github.com/intel/cri-resource-manager/pkg/cri/resource-manager"
 	"github.com/intel/cri-resource-manager/pkg/cri/resource-manager/policy"
 	"github.com/intel/cri-resource-manager/pkg/instrumentation"
 
@@ -43,9 +43,21 @@ func main() {
 
 	printConfig := flag.Bool("print-config", false, "Print configuration and exit.")
 	listPolicies := flag.Bool("list-policies", false, "List available policies.")
+	checkConfig := flag.String("check-config", "", "Check configuration and exit.")
 	flag.Parse()
 
 	switch {
+	case *checkConfig != "":
+		if err := config.SetConfigFromFile(*checkConfig); err != nil {
+			fmt.Printf("validation error: %s\n", err)
+			os.Exit(1)
+		}
+		if *printConfig {
+			config.Print(nil)
+		}
+		fmt.Println("config ok")
+		os.Exit(0)
+
 	case *printConfig:
 		config.Print(nil)
 		os.Exit(0)


klihub (approved):

klihub:
LGTM

cmd/cri-resmgr/main.go
```diff
@@ -43,9 +43,21 @@ func main() {
 
 	printConfig := flag.Bool("print-config", false, "Print configuration and exit.")
 	listPolicies := flag.Bool("list-policies", false, "List available policies.")
+	checkConfig := flag.String("check-config", "", "Check configuration and exit.")
```
klihub, line 46: nitpick: maybe a `check-config-file-syntax` would be a more apt name for this.

klihub:
> Exists with proper exit code to (0-ok) and prints error message.
> 
> Can be used as a check before applying new configuration.
> 
> Just an idea - probably in our extension we will programatically API of config module - more about it here: [intel/gardener-extension-cri-resmgr#56](https://github.com/intel/gardener-extension-cri-resmgr/issues/56)
> 
> BTW: I found undocumented config-help/help arguments to are usefull - are they undocumented on purpose or just missed?

We have not properly added documentation for all conifguration bits, so it's not advertised.

>
> ...
> 
> Marking it as draft because I don't how usefull it would be - probably it just checks syntax (proper keys and types) but not sematics e.g. ? (cpuset is parsed for example, but I can set **policy.Active to non-existents and still returns ok!!!**)

Basically this effectively checks that the configuration file can be loaded and unmarshalled into the internal go data structure representations. That said, there is also a varying level of other validation done in the components that register themselves for configuration, since 1) with go one can override/wrap the default JSON/YAML marshaller and do an extra round of limited semantic validation in the custom one, and 2) our internal config infra allows one to associate post-update notifier callbacks to the data they registered for configuration.

> 
> e.g. non-existing-policy.cfg
> 
> ```
> #cat non-existing-policy.cfg
> policy:
>     Active: "ThereIsNoSuchPolicy"
> 
> # go run ./cmd/cri-resmgr/ -check-config non-existing-policy.cfg
> config ok
> ```

Yeah... the generic policy layer does not do checking of the active policy after (re)configuration... mostly because ATM you can't change the active policy.

This PR and the feedback from you is actually pretty good timing. We're reworking the configuration bits in preparation for adding cleaner support for runtime-switchable policies. While, most of the basic plumbing is done, we're playing around with what extra functionality we want/need to add and how (exactly things like, support for semantic configuration checking, self-documenting configuration, etc.). For instance, if we want, I can easily add enforced checking that any data (type) which is registered for (receiving) configuration (data) needs to at least look like self-documenting..., for instance by requiring that they implement the interface

```go
type [config.]Fragment interface {
    // Provide user-readable documentation about the configuration fragment.
    Describe() string
    // Provide a reasonable sample configuration. With comments included.
    // Unmarshallability of this data is verified during registration.
    SampleYAMLConfig() string
}
```

etc... I can share the devel branch with you for a sneak peak, pre-review.

klihub:
> etc... I can share the devel branch with you for a sneak peak, pre-review.

#903

ppalucki:
Lets close it, and focuse on #903 - which looks as much more comprehesive solution -  - I'm not longer sure I want to follow a path of "static" config validation - for ordinary user it makes no sense - it will just run cri-rm and check output - for our case I want to focus on policy passed by agent - if switching policy will be possible with agent (and maybe other options e.g. metric-interval) - then I don't need this low-level mechanic of validation.

PR #900, opened 2022-10-11
resmgr: make sysfs/other hostfs pseudoroot directory configurable on the command line.

klihub:
Allow overriding the default `sysfs` mount point using the newly introduced `--sysfs-root $dir` command line option. When used, the real host sysfs should be available/mounted as `$dir/sys`.

Target usage for this is to allow an NRI-enabled `cri-resmgr` to run as  a `DaemonSet` inside a container, yet be able to do proper HW-discovery using the full `sysfs` of the host.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -30,9 +30,12 @@ import (
 	idset "github.com/intel/goresctrl/pkg/utils"
 )
 
+var (
+	// Parent directory under which host sysfs, etc. is mounted (if non-standard location).
+	sysRoot = ""
+)
+
 const (
-	// SysfsRootPath is the mount path of sysfs.
-	SysfsRootPath = "/sys"
 	// sysfs devices/cpu subdirectory path
 	sysfsCPUPath = "devices/system/cpu"
 	// sysfs device/node subdirectory path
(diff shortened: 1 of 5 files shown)


askervin:
When making this configurable, wouldn't it be even more future-proof, if one could give sysfs mountpoint, defaulting to `/sys`, rather than giving a root directory and expecting the mountpoint to be called `sys` in it?

klihub:
> When making this configurable, wouldn't it be even more future-proof, if one could give sysfs mountpoint, defaulting to `/sys`, rather than giving a root directory and expecting the mountpoint to be called `sys` in it?

When I looked into `pkg/topology` I realized that it has some correct but hard-coded assumptions about how an LFS-like sysfs hierarchy looks like, with support for an optional prefix (initially only for mocking sysfs for tests). Since it was the fastest/most trivial for me to modify configurability in `pkg/sysfs` to do the same, I opted to go that route. If you know of a setup where this would not work, please let me know and I can take another look.

marquiz:
I think the current approach is actually better wrt. what @askervin suggested, as this is more general. It could be used for other system directories, too, if needed in the future. I'd just rename the option something else to reflect this. E.g. `--sys-root` or `--host-root` or smth. WDYT?

askervin (approved):
LGTM

pkg/sysfs/system.go
```diff
@@ -30,9 +30,12 @@ import (
 	idset "github.com/intel/goresctrl/pkg/utils"
 )
 
+var (
+	// Sysfs root directory.
+	sysfsRoot = ""
```
marquiz, line 35: nit: I think this is misleading here as it kinda represents the "host" root mount point. Rename to `sysRoot` or `hostRoot`, WDYT?

klihub: Okay, true. I'll update.

klihub: Updated taking that comment into account, and therefore also changing the function names to `SetSysRoot()` as opposed to the original `SetSysfsRoot()`.

marquiz (approved):
Thanks @klihub 
lgtm

PR #901, opened 2022-10-12
Experimenting cri-rm with nri and DaemonSet

jukkar:
Tested with
```
reinstall_cri_resmgr=1 k8scri="containerd&cri-resmgr" distro=fedora \
     containerd_src=../containerd vm=fedora-nri-containerd ./run.sh interactive
```

Works well. By default e2e run.sh creates DaemonSet if NRI support is enabled. If NRI needs to be run with systemd or manually, one needs to set `cri_resmgr_plain=1` when executing `run.sh`


klihub:
Closing, for same reason as #888.

PR #902, opened 2022-10-13
e2e: add debian support for distro-install-kernel-dev

askervin:
distro-install-kernel-dev prepares VM for building the latest kernel using distribution's current kernel configuration as a starting point.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -237,6 +237,14 @@ debian-install-golang() {
     debian-install-pkg golang git-core
 }
 
+debian-install-kernel-dev() {
+    distro-refresh-pkg-db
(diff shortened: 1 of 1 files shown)


jukkar (approved):
LGTM, although it is not clear to me why do we want to compile the kernel in the VM.

askervin:
> it is not clear to me why do we want to compile the kernel in the VM.

Waiting for the day when I learn to add reasoning to commit messages... :)

I needed this for studying new sysfs interface in Linux 6.0+ (not-yet-released 6.1) from those parts that are needed by memory activity tracking. And as other parts of CRI-RM use kernel interfaces, too, I see potential value in making testing different versions and configurations of kernel easy.

klihub:
LGTM.

Though, it would be nice to eventually get a subsequent PR for implementing this for fedora-based distros...

klihub (approved):

PR #907, opened 2022-10-24
balloons: fix empty cpuset of a balloon that has containers

askervin:
Keep at least 1 CPU in every balloon that has containers, even if the sum of requested CPUs of the containers would be 0. Otherwise we would allow cpuset of a balloon to become empty, and thus the CPUs of those containers would not be pinned anymore.


marquiz (approved):
lgtm

marquiz:
ping @klihub @jukkar

jukkar (approved):

PR #914, opened 2022-11-15
Don't track static binaries in Git

fmuyassarov:
Intentionally ignore CRI-RM static binaries from git.
Those binaries could appear on the local repository
for example after `make build-static`.


askervin (approved):
LGTM

marquiz (approved):
lgtm

PR #916, opened 2022-11-15
control/memory: check top tier memory control on startup.

klihub:
Check and return an error if the kernels lack the top tier memory control knob. This should cause the control setup code to disable the top tier memory controller and prevent reoccuring warning/error messages whenever we put policy decision in practice.


klihub:
@ppalucki Here is a PR to get rid of the repeated top tier memory controller error messages.

marquiz (approved):
lgtm

PR #921, opened 2022-11-16
Remove the qemu-version verifier  container when it exits

fmuyassarov:
Remove the qemu-version verifier container when it exits.
Otherwise, we end up with some garbage collected when
running e2e tests locally.


marquiz (approved):
lgtm

PR #925, opened 2022-11-21
fixes: update(s) to build with more recent protoc & friends

klihub:
- Makefile: update rules to generate protoc/protobuf golang/gRPC binding
- Makefile, scripts: add scripts and Makefile targets for install protoc and related rules
- agent, config: update for code generated with more recent protoc toolchain


marquiz (commented):
Just wonder what's the reason for buiding protobuf ourselves? Why not use the released version (from https://github.com/protocolbuffers/protobuf/releases)?

Also, started to think that should we do all of this inside a container instead of "polluting" the host system. Somewhat comparable to how we build the documentation for example. Different projects require different versions of the tools/libs and management of all of this is a mess. Also, building in a container would allow to eliminate most of the new variables and rules in the Makefile (just hardcode everything in the Dockerfile) keeping it cleaner and simpler. WDYT?

klihub:
> Just wonder what's the reason for buiding protobuf ourselves? Why not use the released version (from https://github.com/protocolbuffers/protobuf/releases)?

AFAICT the script, which was verbatim taken from containerd repository, does use the exactly those releases:

```
...
wget -O "$PROTOBUF_DIR/protobuf" "https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOBUF_VERSION/protoc-$PROTOBUF_VERSION-linux-x86_64.zip"
...
```

Looks like the same URL prefix... For unrecognized architecures (the `*)` bash-case) it tries to compile from sources...
 
> Also, started to think that should we do all of this inside a container instead of "polluting" the host system. Somewhat comparable to how we build the documentation for example. Different projects require different versions of the tools/libs and management of all of this is a mess. Also, building in a container would allow to eliminate most of the new variables and rules in the Makefile (just hardcode everything in the Dockerfile) keeping it cleaner and simpler. WDYT?

This is not something that normally would be needed. It is only necessary for hacking/changing the proto definitions and then updating the related/generated code. I understand and agree in general with the 'keep all the crap out of my distro in a container' sentiment. Unfortunately based on my practical experience, if you plan/need to do anything with protobuf-defined gRPC/ttRPC protocols, then it's going to be an involved process. And you'll typically want the toolchain(s) natively on your system, not just in a container...

marquiz:
> Looks like the same URL prefix... For unrecognized architecures (the `*)` bash-case) it tries to compile from sources...

Ach sorry, I just glanced at it too quickly. Nevertheless the multiarch build environment for protobuf looks a bit like an overkill... 😊

> This is not something that normally would be needed. It is only necessary for hacking/changing the proto definitions and then updating the related/generated code. I understand and agree in general with the 'keep all the crap out of my distro in a container' sentiment. Unfortunately based on my practical experience, if you plan/need to do anything with protobuf-defined gRPC/ttRPC protocols, then it's going to be an involved process. And you'll typically want the toolchain(s) natively on your system, not just in a container...

There shouldn't be anything supernatural in that. In NFD we do auto-generation (a lot more than just grpc protos) in a container and works fine:
https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/Dockerfile_generator

Anyway, I'm not against this PR. Just had thoughts to keep things a bit tidier and more controlled.

klihub:
> Ach sorry, I just glanced at it too quickly. Nevertheless the multiarch build environment for protobuf looks a bit like an overkill... blush

Totally agree.

> There shouldn't be anything supernatural in that. In NFD we do auto-generation (a lot more than just grpc protos) in a container and works fine: https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/Dockerfile_generator

Okay, sorry, I was too vague in my statement. Definitely there is nothing supernatural in using containers to generate gRPC/ttRPC protoc-bindings from protobufs. I didn't mean that by 'being an involved process'. I meant that in my experience hacking on (anything else than totally trivial) protobuf-based APIs is usually in itself an involved process, during which protoc and friends pretty much become an integral part of your compilation toolchain, basically an extension to the golang compiler itself.

> Anyway, I'm not against this PR. Just had thoughts to keep things a bit tidier and more controlled.

So for the 'controlled' part, I think we'd need to make sure that there is no need to regenerate the gRPC bindings after any commit, at least not inherent in the repo content itself. Any commit to the contrary should be considered incorrect and rejected: the author updated some protobuf's without regenerating or remembering to commit the related golang code. That could be done simply by replacing the current warning-else branch with one that errors out.

About the 'tidy' part... maybe this is not the right way and we need something else.

jukkar (dismissed):
LGTM

askervin (approved):
LGTM

PR #778, opened 2022-02-18
scripts: pairwise maximizes covered triplets

askervin:
- Taking triplet coverage into account improves data coverage without
  increasing the number of parameter sets.
- As a downside, this implementation is naive and may cause
  performance problems with large data sets.

PR #929, opened 2022-11-23
[rebased #778] scripts: pairwise maximizes covered triplets

klihub:
This PR is #778 rebased on the latest master.

- Taking triplet coverage into account improves data coverage without increasing the number of parameter sets.
- As a downside, this implementation is naive and may cause performance problems with large data sets.


marquiz (approved):
lgtm

Issue #927, opened 2022-11-22
Export information shared/dedicated cpus in prometheus metrics (only for new ShareIdleCpus feautre)


ppalucki:
**Describe the solution you'd like**
That only make sense for new ShareIdleCpus feautre.

What we have now is:
```
balloons{balloon="foo[0]",balloon_type="foo",containers="...",cpu_class="", cpus="1-4",cpus_max="32",cpus_min="0",mems="1",tot_req_millicpu="3000"} 3
```
and I would like to see:
```
balloons{balloon="foo[0]",balloon_type="foo",containers="...",cpu_class="", 
cpus="1-4", cpus_shared="4", cpus_dedicated="1-3",
cpus_max="32",cpus_min="0",mems="1",tot_req_millicpu="3000"} 4
```

optionally this info may be just available somewhere in:
/var/lib/cri-resmgr/cache 

It would be even better if you can also count number of cpus:
```
balloons{balloon="foo[0]",balloon_type="foo",containers="...",cpu_class="", 
cpus="1-4", 
cpus_shared="4", cpus_shared_count="1", 
cpus_dedicated="1-3", , cpus_shared_count="3", 
cpus_max="32",cpus_min="0",mems="1",tot_req_millicpu="3000"} 4
```

**Why this is needed**
help to debug which cpus are shared/dedicated in case of share idle cpus feature

@askervin

askervin:
Thanks @ppalucki , this is very nice idea.

You can try out this feature in my 5LT_prototype_marquiz_allocator branch, commit d4a80212d7e34b4af273e2a401ad95da552f0240.

The simplest way to think about the ShareIdleCpus is probably this: this feature does not change the size of the balloons. They can be created, destroyed, inflated and deflated exactly as without this feature. The only difference is that after every creation/destruction/inflation/deflation the policy checks which CPUs are idle and (re)shares those CPUs to workloads in all balloons that are interested in extra CPUs.

Therefore I implemented this metrics feature with slightly different naming, in order to keep the meaning of "cpus" the same as before: they are the (dedicated) CPUs of a balloon. Now the metrics of every balloon include:
- `cpus`: dedicated CPUs of the balloon, cpuset format
- `cpus_count`: the number of dedicated CPUs of the balloon
- `sharedidlecpus`: extra CPUs for workloads in the balloon from the ShareIdleCpus feature, cpuset format
- `shareidlecpus_count`: the number of the extra CPUs
[...]

Issue #908, opened 2022-10-24
Balloons: option for inflating balloons as large as possible without crossing NUMA boundaries


askervin:
Add new balloon type and/or policy option that prefers such CPU allocation for balloons so that balloons can be inflated to their `MaxCPUs` size without crossing the NUMA boundary.

**Why this is needed**

Currently two small balloons can be created on the same NUMA node. When both of these balloons keep inflating, their CPUs do not fit to the same NUMA node anymore, both balloons will get CPUs from both NUMA nodes.

The new option forces balloons to be created on different NUMA nodes in the beginning, and when the balloons are inflated, making sure that they both use only the CPUs from the same NUMA node as long as possible

fmuyassarov:
I'm interested to work on this
/assign

askervin:
Fixed in PR #945 .

PR #931, opened 2022-11-25
balloons: Topology aware create/inflate/deflate and idle CPU sharing

askervin:
Fixes issue #927 .
Fixes issue #908 .

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/balloons/cputree_test.go
+++ b/pkg/cri/resource-manager/policy/builtin/balloons/cputree_test.go
@@ -1,0 +1,606 @@
+// Copyright 2022 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package balloons
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+)
+
+type cpuInTopology struct {
+	packageID, dieID, numaID, coreID, threadID, cpuID             int
+	packageName, dieName, numaName, coreName, threadName, cpuName string
+}
+
+type cpusInTopology map[int]cpuInTopology
+
+func (cit cpuInTopology) TopoName(topoLevel string) string {
+	switch topoLevel {
+	case "thread":
+		return cit.threadName
+	case "core":
+		return cit.coreName
+	case "numa":
+		return cit.numaName
+	case "die":
+		return cit.dieName
+	case "package":
+		return cit.packageName
+	}
+	panic("invalid topoLevel")
+}
+
+func (csit cpusInTopology) dumps(nameCpus map[string]cpuset.CPUSet) string {
+	lines := []string{}
+	names := make([]string, 0, len(nameCpus))
+	for name := range nameCpus {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	for cpuID := 0; cpuID < len(csit); cpuID++ {
+		line := fmt.Sprintf("cpu%02d %s", cpuID, csit[cpuID].threadName)
+		for _, name := range names {
+			if nameCpus[name].Contains(cpuID) {
+				line = fmt.Sprintf("%s %s", line, name)
+			}
+		}
+		lines = append(lines, line)
+	}
+	return strings.Join(lines, "\n")
+}
+
+func newCpuTreeFromInt5(pdnct [5]int) (*cpuTreeNode, cpusInTopology) {
+	pkgs := pdnct[0]
+	dies := pdnct[1]
+	numas := pdnct[2]
+	cores := pdnct[3]
+	threads := pdnct[4]
+	cpuID := 0
+	sysTree := NewCpuTree("system")
+	sysTree.level = CPUTopologyLevelSystem
+	csit := cpusInTopology{}
+	for packageID := 0; packageID < pkgs; packageID++ {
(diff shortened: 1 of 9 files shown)


pkg/cri/resource-manager/policy/builtin/balloons/cputree_test.go
```diff
@@ -0,0 +1,606 @@
...
+	24, // cpu on p1/d1/n0/c0/t0
+	25, // cpu on p1/d1/n0/c0/t1
+	26, // cpu on p1/d1/n0/c1/t0
+	27, // cpu on p1/d1/n0/c1/t1
+	28, // cpu on p1/d1/n1/c0/t0
+	29, // cpu on p1/d1/n1/c0/t1
+	30, // cpu on p1/d1/n1/c1/t0
+	31, // cpu on p1/d1/n1/c1/t1
+},
+*/
```
marquiz, line 203: Is there a reason to have this commented-out block?

marquiz (commented):
Doesn't build atm

klihub:
@askervin Quick 1st question (which I rather ask than try to figure out from the patchset)...

IIUC, with the right set of configuration options, you can now create a balloon which is confined to some of the nodes of the underlying HW topology-tree. When you are running with such a setup and you allocate resources for a container, does the allocator take into account any possible memory allocation promises already given to containers in other balloons which are associated with nodes on the path from the chosen balloon and the tree root ? Or is it so that the configuration is runtime-restricted so that you reject setups where balloons could be assigned to multiple levels of the topology tree ?

pkg/cri/resource-manager/policy/builtin/balloons/cputree_test.go
askervin: Pondered between removing this section or commenting why it is there.

Now ended up adding a comment to give a reason why it exists. If you prefer, I can also delete it. However, it may increase the pain of the poor developer who has to debug why a test starts failing, or write a test. In the first glance it is not obvious where in the topology CPU _i_ belongs to. And I thought that an explicit list might be easier for a developer who doesn't notice that one can get 2-2-2-2-2-topology location for CPU _i_ with

```
(echo obase=2; echo i) | bc
```
or
```
python3 -c 'print(bin(i))'
```
etc.

klihub:
@askervin Could you please rebase this on latest master/HEAD ? Should fix the build problems...

klihub (dismissed):
Please rebase on latest master/HEAD (with commit b88a9a6291a77ecde3737966d6b4d3d1faa84129 included). That should fix the build problems and then we could eventually kick this forward, too.

test/e2e/policies.test-suite/balloons/cri-resmgr.cfg
```diff
@@ -25,6 +25,7 @@ policy:
         Namespaces:
           - "three"
         MinCPUs: 3
+        MaxCPUs: unlimited
```
marquiz, line 28: I thought we ditched this change(?)

marquiz (commented):
I'd suggest to drop the "limit" stuff and rebase on latest master

test/e2e/policies.test-suite/balloons/cri-resmgr.cfg
askervin: I fixed the problems in the ditched version: parsing quotation and unlimited not being the default if `MaxCPUs` or `MaxBalloons` is missing. There were two reasons for advocating the `Limit` instead of the `int` type in the configuration.

First, value 0 needs to be handled as a special case in all comparisons and arithmetics where these two options are involved. Own type forces to pay attention on these operations, thus protecting programmers from stupid errors.

Second, enabling also other than numbers being entered to these fields was supposed to pave the road for small extensions later on, like allowing the number of CPUs/balloon and the number of balloons being limited to the number of CPUs, NUMA nodes, dies or packages on a system, or a combination of these (like `CPUsPerNUMA` or `CPUsPerPackage`).

If you think enabling a user to give a separate `unlimited` value is unwanted (I totally understand this because it adds quite a lot of code only for a single value for now), I could alternatively define simply `typedef Limit int`, and allow only integers in the user interface?

marquiz: > If you think enabling a user to give a separate unlimited value is unwanted (I totally understand this because it adds quite a lot of code only for a single value for now), I could alternatively define simply typedef Limit int, and allow only integers in the user interface?

I think that would preferred as that's also kind of unrelated to the "topic" of this PR.

I'd suggest to add it in a separate PR when we really want to have it. It might be a bit more subtle than it looks. I.e. I think we should probably use the same type for MinCpus and MinBalloon parameters, too. And then, does the available string values depend on the field type....

marquiz:
> Given the reasoning above, would you @marquiz
> 
> 1. accept the fixed "limit" stuff in this patch as is
> 2. accept having only typedef limit in the code, and user giving always numeric values (drop user interface of limit)
> 3. accept only having `int` directly in the configuration (drop user interface and type safety of limit)

If you ask me I think we could keep `const NoLimit = 0` and drop everything else related to the limits 😊 But I don't oppose if others want to merge this as is. One point is also have the sort of orthogonal changes in separate PRs -> makes reviews a lot easier, faster (and probably better quality)

askervin:
@marquiz, I'll remove the Limit type and go with `const NoLimit = 0`.

test/e2e/policies.test-suite/balloons/cri-resmgr.cfg
askervin: Limits removed.

askervin:
@klihub , unit test errors fixed, rebased on top of master. All e2e tests pass. Allocator/releaser sorting made more predictable.

marquiz:
Thanks @askervin for the update.

I just took a quick look and was overwhelmed by the big size 😅 I started to wonder if it would would be possible (or make sense) to split this in separate PRs, e.g. the "topology aware alloc/resize" and "idle cpu sharing" separately. WDYT? It would make it easier for concentrate and faster to review for me, at least. If it doesn't make sense then let's keep this as one mega-PR

askervin:
@marquiz , @klihub , I marked this PR to be a draft, and I'll try to split this into easier-to-review PRs.

The first PR is #941, a set of utility functions that are needed in order to implement numa-aware create/inflate/deflate and idle CPU sharing into the balloons policy.

Do you think it would be feasible to review this set of functions as one PR, and then build the rest of the PRs on top of this?

askervin:
Closing this PR, all pieces of it are created and (almost) merged already.

PR #932, opened 2022-11-25
Balloons: add e2e tests for topology aware allocation and idle CPU sharing

fmuyassarov:
This PR extends E2E to test idle CPU sharing (`ShareIdleCpusInSame` field)
and topology aware creation/inflation/deflation that is implemented in https://github.com/intel/cri-resource-manager/pull/931.

### DIFF
--- a/test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
+++ b/test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
@@ -1,0 +1,76 @@
+terminate cri-resmgr
+cri_resmgr_cfg=${TEST_DIR}/balloons-dynamic.cfg cri_resmgr_extra_args="-metrics-interval 4s" launch cri-resmgr
+
+# pod0-pod7: create 8 balloons, where each lands on a different NUMA node.
+# Each balloon (except one that lands on the NUMA node with reserved CPUs)
+# has 1 shared CPU at the most since a NUMA node has 4 CPUs and a pod is
+# requesting 1 CPU. Only one of the balloon that using NUMA node with
+#reserved CPU has 0 shared CPUs.
+CPUREQLIM="3"
+INITCPUREQLIM="100m-100m 100m-100m 100m-100m"
+POD_ANNOTATION="balloon.balloons.cri-resource-manager.intel.com: dynamic"
+n=8 create multicontainerpod
+verify-metrics-has-line 'balloon="dynamic\[0\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[1\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[2\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[3\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[4\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[5\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[6\]".*cpus_count="3"*'
+verify-metrics-has-line 'balloon="dynamic\[7\]".*cpus_count="3"*'
+verify-metrics-has-no-line 'cpus_count="4"'
+verify-metrics-has-line 'sharedidlecpus_count="1"'
+verify-metrics-has-line 'cpus_allowed_count="4"'
+verify-metrics-has-line 'sharedidlecpus_count="0"'
+verify-metrics-has-line 'cpus_allowed_count="3"'
+verify-metrics-has-no-line 'sharedidlecpus_count="2"'
+verify-metrics-has-no-line 'cpus_allowed_count="5"'
+verify 'disjoint_sets(nodes["pod0c0"], nodes["pod1c0"], nodes["pod2c0"], nodes["pod3c0"], nodes["pod4c0"], nodes["pod5c0"], nodes["pod6c0"], nodes["pod7c0"])' \
+       'len(nodes["pod0c0"]) == len(nodes["pod1c0"]) == len(nodes["pod2c0"]) == \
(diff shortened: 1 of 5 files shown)


test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
```diff
@@ -0,0 +1,83 @@
...
+# pod0-pod7: create 8 balloons, where each lands on a different NUMA node.
+# Each balloon (except one that lands on the NUMA node with reserved CPUs)
+# has 1 shared CPU at the most since a NUMA node has 4 CPUs and a pod is
+# requesting 1 CPU. Only one of the balloon that using NUMA node with
+#reserved CPU has 0 shared CPUs.
+CPUREQLIM="3"
+INITCPUREQLIM="100m-100m 100m-100m 100m-100m"
+POD_ANNOTATION="balloon.balloons.cri-resource-manager.intel.com: dynamic"
+n=8 create multicontainerpod
+sleep 5
```
jukkar, line 13: I have reservation about the forced sleeps, they just delay the progress of the test without much benefit. I think `verify-metrics-has-line` will try the fetch the stuff multiple times anyway. If not possible to remove the sleep, it should be made smaller.

test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
```diff
@@ -0,0 +1,83 @@
...
+        len(nodes["pod3c0"]) == len(nodes["pod4c0"]) == len(nodes["pod5c0"]) == \
+        len(nodes["pod6c0"]) == len(nodes["pod7c0"]) == 1'
+
+# pod8: Add one more pod with 2 CPUs to inflate over NUMAs nodes, which should cross
+# the NUMA node boundaries but not the die boundaries. Because two NUMA nodes can offer
+# 2 CPUs in total. 
+CPUREQLIM="2"
+POD_ANNOTATION="balloon.balloons.cri-resource-manager.intel.com: dynamic"
+create multicontainerpod
+sleep 5
```
jukkar, line 40: same here

test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
```diff
@@ -0,0 +1,83 @@
...
+verify-metrics-has-line 'cpus_count="5"'
+verify-metrics-has-line 'sharedidlecpus="",sharedidlecpus_count="0"'
+verify-metrics-has-line 'sharedidlecpus_count="1"'
+verify 'len(nodes["pod8c0"])==2' \
+       'len(dies["pod8c0"])==1' \
+       'len(packages["pod8c0"])==1'
+kubectl delete pod pod8 --now
+sleep 8
```
jukkar, line 48: and same comment here too, and also the other sleeps below

test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
fmuyassarov: During local runs of these tests, I noticed that some metrics (like deletion of a pod and CPUs of it is becoming free) didn't get updated fast enough (default -metrics-interval 4 seconds) that caused some verification to fail. And similar issue was discovered by @askervin during his trials. And I just realized after your comment, that we don't need sleep for those checks that doesn't use verify-metrics-has-no-line and only verify-metrics-has-line. But for checks with verify-metrics-has-no-line we can either sleep here as it is now (should be more than default interval 4 secs) or I can add a timeout to verify-metrics-has-no-line as we have for [verify-metrics-has-no-line](https://github.com/intel/cri-resource-manager/blob/8a8ebb63b3e199ded81bd0e355b0d2ac56036d86/test/e2e/policies.test-suite/balloons/verify.source.sh#L7). WDYT?

askervin: @fmuyassarov , +1 from me to adding the timeout to verify-metrics-has-no-line, and getting rid of sleeps.

jukkar: Adding timeout to verify-metrics func sounds good, +1 to that.

klihub (changes requested):
And when you address those review comments, please rebase this on latest master/HEAD. That should fix the build errors.

test/e2e/policies.test-suite/balloons/n4c32/test01-dynamic-baloons/code.var.sh
fmuyassarov: done

fmuyassarov:
> And when you address those review comments, please rebase this on latest master/HEAD. That should fix the build errors.

Thanks, done

jukkar (approved):
LGTM

askervin (dismissed):
Thanks @fmuyassarov , LGTM.

jukkar (approved):

klihub (approved):

fmuyassarov:
I don't have rights to merge it. If looks good, can someone please merge this patch?

klihub:
@fmuyassarov Merged... but I already regret it a bit. I mean... ideally shouldn't we have waited for #932 getting merged before merging this ?

fmuyassarov:
> @fmuyassarov Merged... but I already regret it a bit. I mean... ideally shouldn't we have waited for #931 getting merged before merging this ?

Yeah you are right, my bad :(
However, GitHub action doesn't run E2E test (I think) so even if we waited we can't run those tests on top of master but only locally as we do now. But, now we merged tests for improvements that don't exists yet on the master. Perhaps it is okay for now. If there are some functional changes that may happen on 931, then I will make sure to update the E2E tests accordingly.

Thanks for merging btw 😅 and I'm guilty if something comes up due to this

PR #934, opened 2022-11-25
fixes: add missing bits to formally comply with CRI v1 v0.25.4

klihub:


pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
```diff
@@ -343,20 +343,20 @@ func checkReservedPoolNamespaces(namespace string) bool {
 	return false
 }
 
-func checkReservedCPUsAnnotations(c cache.Container) bool {
+func checkReservedCPUsAnnotations(c cache.Container) (bool, bool) {
```
marquiz, line 346: How are the policy changes related??

pkg/cri/server/services.go
```diff
@@ -481,3 +482,20 @@ func (s *server) Status(ctx context.Context,
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return rsp.(*criv1.CheckpointContainerResponse), err
+}
+
+func (s *server) GetContainerEvents(req *criv1.GetEventsRequest, src criv1.RuntimeService_GetContainerEventsServer) error {
+	return nil
```
marquiz, line 500: Hmm, I was wondering would the right thing to do be returning a "not implemented" error?

marquiz (commented):
Thanks @klihub for making the project buildable again 😎 I think it's ok (and good) to state GetContainerEvents as unimplemented for the time being.

pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go
klihub: Argh... not in any way. That was supposed to be a separate PR... I just hacked it on the same branch and failed to drop the patch afterwards. Thanks for noticing.

pkg/cri/server/services.go
klihub: It depends whether we want to mimick `containerd` or `cri-o` ;-) But seriously... it might be better to just return an error. I'll update it like that.

marquiz (approved):
lgtm

askervin (approved):
LGTM

jukkar (approved):

PR #936, opened 2022-11-25
topology-aware: don't restore config on startup

klihub:
**Notes:** This PR depends/is stacked on #934. Please review / merge that one first.

Do not save and especially don't restore policy configuration as policy-specific data. It silently overwrites the effective configuration during policy startup. This cases really tricky to debug failures in some tests that use forced configuration to alter behavior but are not run first after startup.


marquiz (approved):
Makes totally sense. Wonder why we even had this in the policy 🧐

LGTM but let's wait for #934 to be merged first

askervin (approved):
LGTM

PR #937, opened 2022-11-29
fixes: ignore topology pseudo-hints

klihub:
Ignore topology pseudo-hints generated based on the original CPU allocation/pinning by the CPUManager in kubelet. Such hints are only present if we run with the CPUManager enabled which is not a recommended setup. Even if we run in such an environment, normally the original allocation attempt should have no effect on our decisions and just get ignored.


marquiz (approved):
LGTM

PR #938, opened 2022-11-29
resource-manager: remove 'null' pseudo-policy.

klihub:
Remove the 'null' pseudo-policy and the related bypass code-paths which leave part of the normal CRI relaying and intercepting infra uninitialized. Use the no-op `none` policy as the default. This setup should be a much more reasonable default. It should also allow us to change the handling of configuration updates to enable/handle the switch from the default `none` policy to any other one.


marquiz (approved):
Thanks @klihub for this change. Makes the code cleaner and more importantly the software easier to understand
LGTM

askervin (approved):
So long, Bypassed().

LGTM.

jukkar (approved):

PR #942, opened 2022-12-02
balloons: fix handling of undefined (unlimited) MaxCPUs

askervin:
- When MaxCpus=NoLimit=0, maximum available CPUs of a balloon depends on the number of free CPUs in the system.
- Validate min/max values in balloons policy configurations.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
@@ -50,6 +50,8 @@ const (
 	reservedBalloonDefName = "reserved"
 	// defaultBalloonDefName is the name in the default balloon definition.
 	defaultBalloonDefName = "default"
+	// NoLimit value denotes no limit being set.
+	NoLimit = 0
 )
 
 // balloons contains configuration and runtime attributes of the balloons policy
@@ -122,7 +124,10 @@ func (bln Balloon) AvailMilliCpus() int {
 	return bln.Cpus.Size() * 1000
 }
 
-func (bln Balloon) MaxAvailMilliCpus() int {
+func (bln Balloon) MaxAvailMilliCpus(freeCpus cpuset.CPUSet) int {
+	if bln.Def.MaxCpus == NoLimit {
+		return (bln.Cpus.Size() + freeCpus.Size()) * 1000
+	}
 	return bln.Def.MaxCpus * 1000
 }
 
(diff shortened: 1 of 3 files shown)


askervin:
This patch is a part of the changes in #931 mega-PR.

pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
```diff
@@ -922,11 +927,31 @@ func (p *balloons) applyBalloonDef(balloons *[]*Balloon, blnDef *BalloonDef, fre
...
 
+func (p *balloons) validateConfig(bpoptions *BalloonsOptions) error {
+	for _, blnDef := range bpoptions.BalloonDefs {
+		if blnDef.MaxCpus != NoLimit && blnDef.MinCpus > blnDef.MaxCpus {
+			return balloonsError("MinCpus (%d) > MaxCpus (%d) in balloon type %q",
+				blnDef.MinCpus, blnDef.MaxCpus, blnDef.Name)
+		}
+		if blnDef.Name == reservedBalloonDefName || blnDef.Name == defaultBalloonDefName {
+			continue
+		}
```
fmuyassarov, line 938: just for my sake of knowledge, what it this for?

fmuyassarov (dismissed):
Looks good. Just one question in line.

pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go
askervin: It's a very valid question, @fmuyassarov .

There are two special balloon types which both have only one instance: the reserved balloon and the default balloon. The default balloon is sometimes almost an alias to the reserved balloon as it uses the same CPUs. But sometimes  it is not: it might have been configured explicitly, which then makes it a clearly separate balloon.

In this patch version, validations before this check are done to all balloon types (including the two built-in balloon types), while validations after this check only to user-configured balloon types.

However, your comment raises two questions:
1. If this kind of a separation is really wanted/needed, then it should be clearly commented in the code.
2. If this separation is not absolutely necessary (why built-in types shouldn't comply with all checks that are done to user-defined types), then why to have it.

I suspect that at some point I had some code that didn't really work with the built-in balloon types, and now that code does not exist anymore. Therefore I think I'll remove this check.

Thanks for pointing this out, @fmuyassarov !

askervin: Fixed (removed).

fmuyassarov: Thanks for detailed info @askervin. LGTM now.

fmuyassarov (approved):

klihub (approved):
@marquiz @fmuyassarov @jukkar LGTM. If you guys have BW to take a better/closer look, please do so.

jukkar (approved):

PR #949, opened 2022-12-09
Makefile: specify go version in a make variable

marquiz:
Automatic detection of Golang version from go.mod does not work properly as with that we don't get the latest patch release, but the dot-zero version of it. It worked correctly for webhook and agent builds as the golang:maj.min docker images (that we use for building those) always point to the latest patch release but the rpm and deb builds were flawed.

The upside of this change is that we get more reproducible builds as the golang version is fixed. The downside is that we need to manually keep updating the version.


klihub (approved):
LGTM

marquiz:
Verified that this actually works 😊

PR #951, opened 2022-12-09
github: update release instructions

marquiz:
Instruct to use 'cross-tar' make target for building the "distroless" binary tarball. This builds inside a container (like cross targets) which means that we have more reproducible builds, e.g. using the golang version specified in the Makefile instead of using whatever happens to be installed on the host system (if any).


klihub (approved):
LGTM.

As a small nit, while we're at it, I think we should make `binary-dist` an alias for `cross-tar`.

marquiz:
> As a small nit, while we're at it, I think we should make `binary-dist` an alias for `cross-tar`

`binary-dist` is a "sub-target` of  `cross-tar`, i.e. `cross-tar` does `make binary-dist` inside the container -> I didn't want to change that in the Makefile, just to go with minimal changes there

jukkar (approved):

fmuyassarov (approved):

marquiz:
ping @klihub @jukkar @askervin anyone. Let's get this one merged and I'll update #951 accordingly

PR #953, opened 2022-12-12
Makefile: add cross-tar to cross-packages

marquiz:
Also build the "distroless" binary tarball with the corss-packages
target similar to the deb and rpm packages.

Also, change packaging-tests target to depend on cross-packages so that
it also tests the binary tarball. Even if it currently only tests the
buildability of the tarball as there is no actual e2e test to install
and verify it. Adding a packaging e2e-test for the binary tarball is
left as a future exercise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,7 +451,7 @@ e2e-tests: build-static
 	    exit 1; \
 	fi
 
-packaging-tests: cross-deb cross-rpm
+packaging-tests: cross-packages
 	$(Q)for dir in test/e2e/packages.test-suite/*; do \
 	    cleanup=1 omit_agent=1 $(E2E_RUN) $$dir; \
 	    echo "--------------------------------------"; \
@@ -480,7 +480,7 @@ else
     packages: cross-$(DISTRO_PACKAGE).$(DISTRO_ID)
 endif
 
-cross-packages: cross-rpm cross-deb
+cross-packages: cross-rpm cross-deb cross-tar
 
 cross-rpm: $(foreach d,$(SUPPORTED_RPM_DISTROS),cross-rpm.$(d))
 


marquiz:
ping @klihub WDYT?

The binary tarball hasn't traditionally been considered as "packaging" in cri-rm but it'd be nice to have "one make target to rule them all"

klihub (dismissed):

klihub (commented):
This is fine, but is it enough ? AFAICT, with these changes we only build the tarball as a 'packaging test'. IIRC, the other packaging tests check that cri-resmgr can be installed from the built packages, it can start up, and that a container can be created once it is running.

Shouldn't we do the same for the tar package ? It's a bit more involved than a few one-liners, because we'd need to teach `test/e2e/run.sh`, `demo/lib/vm.bash` et al. to handle something like `binsrc=tarball`. So I'm also fine with doing that in another PR... and maybe not for this release.

marquiz:
> packaging tests check that cri-resmgr can be installed from the built packages, it can start up, and that a container can be created once it is running.
> 
> Shouldn't we do the same for the tar package ? It's a bit more involved than a few one-liners, because we'd need to teach `test/e2e/run.sh`, `demo/lib/vm.bash` et al. to handle something like `binsrc=tarball`. So I'm also fine with doing that in another PR... and maybe not for this release.

Yeah, that's what I commented in the commit message, too. There's no real packaging e2e-test for it yet. But I think we can address that gap later. I think testing even buildability is better than nothing...

marquiz:
I changed to scope of the PR. Title and message adjusted accordingly

klihub:
> > packaging tests check that cri-resmgr can be installed from the built packages, it can start up, and that a container can be created once it is running.
> > Shouldn't we do the same for the tar package ? It's a bit more involved than a few one-liners, because we'd need to teach `test/e2e/run.sh`, `demo/lib/vm.bash` et al. to handle something like `binsrc=tarball`. So I'm also fine with doing that in another PR... and maybe not for this release.
> 
> Yeah, that's what I commented in the commit message, too. There's no real packaging e2e-test for it yet. But I think we can address that gap later. I think testing even buildability is better than nothing...

Definitely. We can go with this and add binsrc=tarball support to enable install+start+create-pod test in a further PR.

marquiz:
> There is a typo (extra '§' I assume) in the commit message: 'Also build §the "distroless"', which you might want to fix.

Woops, late night fat-fingering. Thanks for spotting this. Fixed now

marquiz:
ping @klihub @askervin

askervin (approved):
LGTM

jukkar (approved):

PR #952, opened 2022-12-12
test/e2e: update packaging tests

marquiz:

PR #955, opened 2022-12-13
e2e: add opensuse-15.4, drop 15.3, DISABLE_VGA depends on distro

askervin:
- Update govm opensuse to 15.4.
- Drop support from 15.3 that will be EOL'd in Dec 2022.
- Allow switching govm/Qemu VGA disabling depending on distro. OpenSUSE 15.4 fails to boot when Qemu VGA is disabled, but Ubuntu 22.04 guest VM has issues in qemu -vga QXL.

Enables passing opensuse-15.4 packaging tests in PR #952.


jukkar (approved):

marquiz (approved):
Thanks @askervin! 
LGTM

PR #958, opened 2022-12-14
Makefile: dynamically linked binary-dist

marquiz:
Build a dynamically linked cri-resmgr binary with the binary-dist target. Change the builder image from fedora to centos-7 in order to link against as-old-as possible glibc version to mitigate risk of users hitting glibc symbol version skew.

Also clean up fedora builder dockerfile to remove support for static linking.


klihub (approved):
LGTM.

PR #959, opened 2022-12-15
packaging: store licenses

marquiz:
Use the go-licenses tool to save all used licenses (plus MPL-licensed
code) inside all distributed binary "archives". That is deb and rpm
packages, the "distroless" binary tarball and the container images of
cri-resmgr-agent and cri-resmgr-webhook.

For agent and webhook dockerfiles this patch also adds the -x flag to go
mod download to print progress information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ GO_CYCLO    := gocyclo
 GO_LINT     := golint
 GO_CILINT   := golangci-lint
 GO_VERSION  ?= 1.18.9
+GOLICENSES_VERSION  ?= v1.5.0
 
 # TEST_TAGS is the set of extra build tags passed for tests.
 # We disable AVX collector for tests by default.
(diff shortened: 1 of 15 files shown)


jukkar (commented):
Just wondering, why can't we add suitable rules to Makefile and generate the license info file in one place. Then just include needed files when generating the docker image, tar-ball, rpm, deb etc.

marquiz:
Thanks for the review. Updated accordingly and now this also covers all packaging: container images, rpm and deb and the binary tarball.

klihub (approved):
LGTM

askervin:
Great, thanks @marquiz!

I'm fine merging this as is. The only suggestion/question that I have is on handling go-licenses tool installation in Makefile like we do for protoc-tools. I mean adding something like:

```
install-go-licenses:
        $(Q)$(GO_INSTALL) github.com/google/go-licenses@$(GOLICENSES_VERSION)
```

This would enable dropping `GOLICENSES_VERSION` parameter from Dockerfiles, and have the install command in only one place. WDYT?

askervin (approved):
LGTM, with the suggestion/question above. We can address it later, too, if considered necessary.

jukkar (approved):
LGTM

marquiz:
> This would enable dropping `GOLICENSES_VERSION` parameter from Dockerfiles, and have the install command in only one place. WDYT?

The problem with this I see is that if you have `RUN make install-go-licenses` in the Dockerfile  then changing the value of `GOLICENSES_VERSION` doesn't trigger a rebuild of the container (as the Dockerfile) or it's parameters haven't changed.

For the package (rpm, deb, bintar) we could use it I think by doing `make install-go-licenses` inside `DOCKER_RPM_BUILD` et al . However, I suggest to leave that as a future exercise if you're interested in improving this 😉

fmuyassarov (approved):
LGTM

PR #961, opened 2022-12-15
cgroups: fix controller filesystem path in cgroup v2

askervin:
- Fixes page demotion in cgroup v2. This was failing because GetProcesses() and GetCgroupDir() did not work.

### DIFF
--- a/pkg/cgroups/cgroupcontrol.go
+++ b/pkg/cgroups/cgroupcontrol.go
@@ -30,7 +30,7 @@ type Controller int
 // Group represents a control group.
 type Group string
 
-//nolint
+// nolint
 const (
 	// UnkownController represents a controller of unknown type.
 	UnknownController Controller = iota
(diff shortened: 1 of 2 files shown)


pkg/cgroups/cgrouppath.go
```diff
@@ -83,3 +88,14 @@ func init() {
 	flag.StringVar(&KubeletRoot, "kubelet-cgroup-root", KubeletRoot,
 		"--cgroup-root options the kubelet is running with")
 }
+
+func DetectSystemCgroupVersion() int {
```
marquiz, line 92: I'd suggest to do smth like this instead, possibly import/copy it directly
https://github.com/containerd/cgroups/blob/e8802a182774b8e25cc65b5c374389338cc8e6c3/utils.go#L53

pkg/cgroups/cgroupcontrol.go
```diff
@@ -30,7 +30,7 @@ type Controller int
 // Group represents a control group.
 type Group string
 
-//nolint
+// nolint
```
marquiz, line 33: Why is the space removed. Machine-parsed comments should not have a space in-between, IIRC

pkg/cgroups/cgrouppath.go
```diff
@@ -16,11 +16,12 @@ package cgroups
 
 import (
 	"flag"
+	"os"
 	"path"
 	"path/filepath"
 )
 
-//nolint
+// nolint
```
marquiz, line 24: ditto about machine-parsed comments

pkg/cgroups/cgroupcontrol.go
askervin: This change was by `gofmt`.

marquiz: Hmm, ok  nevermind then🧐

marquiz (approved):
We can go with this, too

jukkar (approved):

PR #964, opened 2022-12-19
e2e: relax idle CPU load max from 20 to 30 %

askervin:
After debugging with "perf sched" it turned out that kube-apiserver and etcd where constantly scheduled on certain CPU core in e2e test vms. They generated high enough load to fail the test that checks reporting CPU load of podpools where containers were expected to do nothing but sleep.

### DIFF
--- a/test/e2e/policies.test-suite/podpools/n4c16/test06-prometheus-metrics/code.var.sh
+++ b/test/e2e/policies.test-suite/podpools/n4c16/test06-prometheus-metrics/code.var.sh
@@ -57,7 +57,7 @@ out "### Idle single-container pod"
 CPUREQ="400m" MEMREQ="" CPULIM="400m" MEMLIM=""
 POD_ANNOTATION="pool.podpools.cri-resource-manager.intel.com: 400mCPU" CONTCOUNT=1 create podpools-busybox
 report allowed
-verify-log-vs-metrics pod0:pod0c0 0 20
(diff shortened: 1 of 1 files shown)


klihub (commented):
Looks like an extremely harmless change.

So, out of curiosity, was the failure related to testing in a VM ? If not, aren't the control plane pods/containers isolated to the reserved cpuset in podpools, so they should not get scheduler to cpus of other pools ?

jukkar (approved):

marquiz (approved):
lgtm

marquiz:
> aren't the control plane pods/containers isolated to the reserved cpuset in podpools, so they should not get scheduler to cpus of other pools ?

Yeah, good point, shouldn't we be able to isolate those? 🧐

askervin:
@marquiz @klihub , this needs further investigation. Let's not merge it.

askervin:
Relaxing the CPU usage limits was needed due to a real problem in CPU pinning of already running kube-system containers (kube-apiserver, etcd). Let's fix the problem by pinning these containrs properly to reserved CPUs. A policy-specific solution is in PR #965, but a framework-level solution that provides all relevant containers in Start(add, del) parameters would be even better. If we end up changing the framework, then changes in commit 7a83bcce4526391f7873cc0ff77aaa9ad7517cff should be revisited.

PR #965, opened 2022-12-20
podpools: fix synchronizing all running containers at start

askervin:
This change makes podpools start similar to the balloons policy start: find all running containers in the system.

### DIFF
--- a/pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
+++ b/pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
@@ -178,7 +178,7 @@ func (p *podpools) Description() string {
 // Start prepares this policy for accepting allocation/release requests.
 func (p *podpools) Start(add []cache.Container, del []cache.Container) error {
 	log.Info("%s policy started", PolicyName)
-	return p.Sync(add, del)
+	return p.Sync(p.cch.GetContainers(), del)
 }
 
 // Sync synchronizes the active policy state.
--- a/test/e2e/policies.test-suite/podpools/n4c16/test06-prometheus-metrics/code.var.sh
+++ b/test/e2e/policies.test-suite/podpools/n4c16/test06-prometheus-metrics/code.var.sh
@@ -97,6 +97,8 @@ CPUREQ="" CPULIM=""
 namespace=kube-system CONTCOUNT=3 create podpools-busybox
 report allowed
 vm-command "curl --silent $metrics_url | grep -v ^cgroup_"
-verify-log-vs-metrics pod5:pod5c1 0 20
+# There should be kube-apiserver, etcd etc. running on reserved CPUs as well,
+# therefore allow a lot of CPU usage yet pod5 is not doing anything.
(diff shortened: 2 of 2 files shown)


pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go
```diff
@@ -178,7 +178,7 @@ func (p *podpools) Description() string {
 // Start prepares this policy for accepting allocation/release requests.
 func (p *podpools) Start(add []cache.Container, del []cache.Container) error {
 	log.Info("%s policy started", PolicyName)
-	return p.Sync(add, del)
+	return p.Sync(p.cch.GetContainers(), nil)
```
klihub, line 181: Is that `nil` as deleted containers really correct ? Isn't there another corner case (besides shutdown, change config, start up again, which this attempts to fix IIUC) where some containers are gone while cri-resmgr/the policy is down/away ? And if there is such a case, won't this hack cause those container deletions to go unnoticed and have lingering resource allocations after this change ? It would feel more correct to try freeing everything in del and then allocate everything for all known containers. And that should be the result of calling `p.Sync(p.cch.GetContainers(), del)`, because that's how Sync should work: first free removed containers, then allocate added ones.

askervin: Thanks @klihub , fixed this issue. Added PR #966 to fix the same issue in the balloons policy.

jukkar (approved):

test/e2e/policies.test-suite/podpools/n4c16/test06-prometheus-metrics/code.var.sh
```diff
@@ -97,6 +97,8 @@ CPUREQ="" CPULIM=""
 namespace=kube-system CONTCOUNT=3 create podpools-busybox
 report allowed
 vm-command "curl --silent $metrics_url | grep -v ^cgroup_"
-verify-log-vs-metrics pod5:pod5c1 0 20
+# There should be kube-apiserver, etcd etc. running on reserved CPUs as well,
+# therefore allow a lot of CPU usage yet pod5 is not doing anything.
+verify-log-vs-metrics pod5:pod5c1 0 100
```
marquiz, line 102: Is this somehow related?

askervin: Yes, strongly. Now that the fix in policy Start() makes sure that kube-apiserver and etcd (among others) will be pinned to the reserved CPU - the same CPU where pod5 is scheduled too - the CPU load of that CPU can be much higher than at the time when this test was written. I got one failed test run because of that: the two processes may easily cause > 20% load on the only reserved CPU in our e2e VMs.

marquiz: Thanks 👍

marquiz (approved):
lgtm

PR #967, opened 2022-12-21
e2e: fix isolcpus test on Fedora

askervin:
- Fixes topology-aware/n4c16/test08-isolcpus test on Fedora.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -529,6 +529,10 @@ fedora-install-containerd-pre() {
     fedora-install-containernetworking-plugins
 }
 
+fedora-install-containerd-post() {
+    vm-command "systemctl enable containerd"
+}
+
 fedora-install-k8s() {
     local repo="/etc/yum.repos.d/kubernetes.repo"
     local base="https://packages.cloud.google.com/yum/repos/kubernetes-el7-\$basearch"
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -605,6 +605,17 @@ fi
 EOF
 }
 
+fedora-set-kernel-cmdline() {
+    local e2e_defaults="$*"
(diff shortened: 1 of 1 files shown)


demo/lib/distro.bash
```diff
@@ -529,6 +529,10 @@ fedora-install-containerd-pre() {
     fedora-install-containernetworking-plugins
 }
 
+fedora-install-containerd-post() {
+    vm-command "systemctl enable containerd"
```
marquiz, line 533: not related to kernel cmdline 🧐 A separate update the commit message?

klihub: Yeah, I wonder about same. IMO, there is no need for two PRs, but it would make sense to have those as two separate commits, both with proprer commit messages, if for the related hunk the current message is used. And maybe the PR title could be something a bit more appropriate, like "e2e: fix for isolcpus tests on fedora"...

jukkar:
Argh, I totally forgot to send my fix for this same issue, it has been sitting in the #901 for a long time.
Please check also alternative proposal in https://github.com/intel/cri-resource-manager/pull/901/commits/c4c9f0ccc252bd3a87e9397b0ed9f3e93b43b3a9 if it makes sense at all.

askervin:
Thanks guys!

This is now split in two commits, perhaps slightly unorthodoxically using the same PR for both... but at least both commits fall under the topic of fixing test08-isolcpus on Fedora.

marquiz (approved):
Thanks @askervin 

I changed the PR title to
**e2e: fix isolcpus test on Fedora**

lgtm

marquiz:
> [c4c9f0c](https://github.com/intel/cri-resource-manager/commit/c4c9f0ccc252bd3a87e9397b0ed9f3e93b43b3a9) if it makes sense at all.
> Antti's proposal looks much smaller so I am ok if we use that.

I think let's take that separately and go with this smaller fix for now(?)

PR #970, opened 2023-01-04
sysfs: add SysRoot() getter for SetSysRoot()

askervin:
This change enables policies to read and use cri-resmgr --host-root value, that will help validation and running in custom environments.


jukkar (approved):

marquiz (approved):
Thx @askervin, improves the symmetry of the api of this pkg
LGTM

PR #971, opened 2023-01-18
cpuallocator: fix cpu priority discovery

marquiz:
The CPU priorities were discovered correctly but not stored in the cpuallocator so they were effectively discarded.

### DIFF
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -428,6 +428,7 @@ func (c *topologyCache) discoverCPUPriorities(sys sysfs.System) {
 			prio[p] = prio[p].Union(cset)
 		}
 	}
+	c.cpuPriorities = prio
 }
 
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
LGTM

askervin (approved):
LGTM.

klihub (approved):
LGTM

Should we have a unit test with a mocked sysfs exercising and verifying this feature, using exclusively the publicly visible interfaces ?

jukkar (approved):

marquiz:
> Should we have a unit test with a mocked sysfs exercising and verifying this feature, using exclusively the publicly visible interfaces ?

I think we should. But not just for this feature but for the cpuallocator overall. I'd suggest to leave this as a future exercise as we don't currently have any unit test infra in the cpuallocator to quickly add a new testcase like this

PR #981, opened 2023-02-21
Bump builds to use Golang v1.19.6

marquiz:
Bump from v1.18 to the latest patch release of v1.19. v1.18 has been archived so it's better not to use it anymore.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -10,7 +10,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v1
       with:
-        go-version: 1.18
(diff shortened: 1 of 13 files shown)


klihub:
Is there any dep preventing us from going straight to 1.20 ?

marquiz:
> Is there any dep preventing us from going straight to 1.20 ?

No, I'm not aware of this.

marquiz:
My thinking was just that v1.19 is several releases old and true and tested but I guess v1.20 works as well. Created #982 so you can choose

klihub (approved):

PR #985, opened 2023-02-27
pkg/avx/collector: don't crash on no regexp match.

klihub:
Don't assume blindly that the regexp match in the  avx collector will produce a non-empty slice of results. Otherwise it might crash if we try to index an empty slice with `0`.


jukkar (approved):
LGTM

mythi (approved):

marquiz (approved):
lgtm

PR #988, opened 2023-03-13
Migrating from CRI-RM to NRI

luukasmakila:
Documentation for moving from CRI-RM to NRI.

Summary of changes:
- Added the docs/ directory
- Added migration-to-NRI.md with the steps

### DIFF
--- a/docs/migration-to-NRI.md
+++ b/docs/migration-to-NRI.md
@@ -1,0 +1,216 @@
+# Migrating from CRI-RM to NRI
+
+## Prerequisities
+
+- Up and running CRI Resource Manager
+- One of the two supported policies in use: balloons or topology-aware.
+- For other policies a little bit more work is required and the policies need to be 'ported'. This can be done by just following the example of how balloons or topology-aware were converted.
+
+## Steps for an initial/basic migration test
+
+### Containerd
+
+Replace the containerd version in the system with 1.7 or newer version (NRI server not supported in older versions).
+
+Replace kubelet's --container-runtime-endpoint=/var/run/cri-resmgr/cri-resmgr.sock with --container-runtime-endpoint=/var/run/containerd/containerd.sock
+
+Replacing the runtime endpoint on a node that was setup using Kubeadm:
+```console
+# Get the Kubelet args
+systemctl cat kubelet <- Look for: EnvironmentFile=/.../kubeadm-flags.env
+
+vim /.../kubeadm-flags.env
+  KUBELET_KUBEADM_ARGS="--container-runtime-endpoint=unix:///var/run/containerd/containerd.sock --pod-infra-container-image=registry.k8s.io/pause:3.9"
+
+vim /etc/sysconfig/kubelet
+  KUBELET_EXTRA_ARGS= --container-runtime-endpoint=/var/run/containerd/containerd.sock <- Remember this aswell
+
+systemctl restart kubelet
+```
+
+Edit the containerd config file and look for the section [plugins."io.containerd.nri.v1.nri"] and replace "disable = true" with "disable = false":
+```console
+vim /etc/containerd/config.toml
+```
+```toml
+[plugins."io.containerd.nri.v1.nri"]
+  disable = false
+  disable_connections = false
+  plugin_config_path = "/etc/nri/conf.d"
+  plugin_path = "/opt/nri/plugins"
+  plugin_registration_timeout = "5s"
+  plugin_request_timeout = "2s"
+  socket_path = "/var/run/nri/nri.sock"
+```
+```console
+systemctl restart containerd
+```
+
+### CRI-O
+
+Ensure that crio version 1.26.2 or newer is used.
+
+Replace kubelet's --container-runtime-endpoint=/var/run/cri-resmgr/cri-resmgr.sock with --container-runtime-endpoint=/var/run/crio/crio.sock
+
+Replacing the runtime endpoint on a node that was setup using Kubeadm:
+```console
+# Get the Kubelet args
+systemctl cat kubelet <- Look for: EnvironmentFile=/.../kubeadm-flags.env
+
+vim /.../kubeadm-flags.env
+  KUBELET_KUBEADM_ARGS="--container-runtime-endpoint=unix:///var/run/crio/crio.sock --pod-infra-container-image=registry.k8s.io/pause:3.9"
+
+vim /etc/sysconfig/kubelet
+  KUBELET_EXTRA_ARGS= --container-runtime-endpoint=/var/run/crio/crio.sock <- Remember this aswell
+
+systemctl restart kubelet
+```
+
+Enable NRI:
+```console
+CRIO_CONF=/etc/crio/crio.conf
+cp $CRIO_CONF $CRIO_CONF.orig
+crio --enable-nri config > $CRIO_CONF
+systemctl restart crio
+```
+
+### Build the NRI policies
+
+```console
+git clone https://github.com/containers/nri-plugins.git
+cd nri-plugins
+make
+# Build the images, specify your image repo to easily push the image later.
+make images IMAGE_REPO=my-repo IMAGE_VERSION=my-tag
+```
+
+### Create required CRDs
+
+```console
+kubectl apply -f deployment/base/crds/noderesourcetopology_crd.yaml
+```
+
+### Import the image of the NRI plugin you want to run
+
+<b>Containerd</b>
+
+```console
+ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-*.tar
+```
+
+<b>CRI-O</b>
+
+See the section [below](#steps-for-a-more-real-life-migration-using-self-hosted-image-repository) for instructions on how to push the images to a registry, then pull from there.
+
+### Deploy the plugin
+
+  ```console
+  kubectl apply -f build/images/nri-resmgr-topology-aware-deployment.yaml
+  ```
+
+### Deploy a test pod
+
+```console
+kubectl run mypod --image busybox -- sleep inf
+kubectl exec mypod  -- grep allowed_list: /proc/self/status
+```
+
+### See the resources the pod got assigned with
+
+```console
+kubectl exec $pod -c $container  -- grep allowed_list: /proc/self/status
+# Output should look similar to the output of CRI-RM
+```
+
+## Steps for a more real-life migration using self-hosted image repository
+
+- Same steps as above for enabling NRI with Containerd/CRI-O and building the images.
+- Push the images built to your repository:
+  ```console
+  # Replace my-repo and my-tag with the IMAGE_REPO and IMAGE_VERSION you specified when building the images with make images
+  docker push my-repo:my-tag
+  ```
+
+- Remember to change the image name & pull policy in the plugins .yaml file to match your registyr and image, ex:
+  ```console
+  vim build/images/nri-resmgr-topology-aware-deployment.yaml
+  ```
+
+- Then deploy the plugin simlarly to the earlier step.
+
+## Migrating existing configuration
+
+- The ConfigMap used by the ported policies/infra has a different name/naming scheme than the original one used in CRI-RM, ex:
+  - configMapName:
+    ```diff
+    - configmap-name: cri-resmgr-config
+    + configmap-name: nri-resource-policy-config
+    ```
+  - The details of grouping nodes by labeling to share configuration:
+    ```diff
+    - cri-resource-manager.intel.com/group: $GROUP_NAME
+    + resource-policy.nri.io/group: $GROUP_NAME
+    ```
+
+## Migrating existing workloads
+
+- The annotations one can use to customize how a policy treats a workload use slightly different keys than the original ones in CRI-RM. The collective 'key namespace' for policy- and resource-manager-specific annotation has been changed from cri-resource-manager.intel.com to resource-policy.nri.io.
+
+- For instance, an explicit type annotation for the balloons policy, which used to be:
+  ```yaml
+  ...
+  metadata:
+    annotations:
+      balloon.balloons.cri-resource-manager.intel.com/container.$CONTAINER_NAME: $BALLOON_TYPE`
+  ...
+  ```
+
+- Should now be:
+  ```yaml
+  ...
+  metadata:
+    annotations:
+      balloon.balloons.resource-policy.nri.io/container.$CONTAINER_NAME: $BALLOON_TYPE`
+  ...
+  ```
+
+- Similarly a workload opt-out annotation from exclusive CPU allocation for the topology-aware policy, which used to be:
+  ```yaml
+  ...
+  metadata:
+    annotations:
+      prefer-shared-cpus.cri-resource-manager.intel.com/container.$CONTAINER_NAME: "true"
+  ...
+  ```
+
+- Should now be:
+  ```yaml
+  ...
+  metadata:
+    annotations:
+      prefer-shared-cpus.resource-policy.nri.io/container.$CONTAINER_NAME: "true"
+  ...
+  ```
+
+- Similar changes are needed for any cri-resmgr-specific annotation that uses the same semantic scoping for key syntax.
+
+
+All of the annotations:
+| Was                                                 | Is now                                      |
+| --------------------------------------------------- | ------------------------------------------- |
+| cri-resource-manager.intel.com/afffinity            | resource-policy.nri.io/affinity             |
+| cri-resource-manager.intel.com/anti-afffinity       | resource-policy.nri.io/anti-affinity        |
+| cri-resource-manager.intel.com/prefer-isolated-cpus | resource-policy.nri.io/prefer-isolated-cpus |
+| cri-resource-manager.intel.com/prefer-shared-cpus   | resource-policy.nri.io/prefer-shared-cpus   |
+| cri-resource-manager.intel.com/cold-start           | resource-policy.nri.io/cold-start           |
+| cri-resource-manager.intel.com/memory-type          | resource-policy.nri.io/ memory-type         |
+| prefer-isolated-cpus.cri-resource-manager.intel.com | prefer-isolated-cpus.resource-policy.nri.io |
+| prefer-shared-cpus.cri-resource-manager.intel.com   | prefer-shared-cpus.resource-policy.nri.io   |
+| memory-type.cri-resource-manager.intel.com          | memory-type.resource-policy.nri.io          |
+| cold-start.cri-resource-manager.intel.com           | cold-start.resource-policy.nri.io           |
+| prefer-reserved-cpus.cri-resource-manager.intel.com | prefer-reserved-cpus.resource-policy.nri.io |
+| rdtclass.cri-resource-manager.intel.com             | rdtclass.resource-policy.nri.io             |
+| blockioclass.cri-resource-manager.intel.com         | blockioclass.resource-policy.nri.io         |
+| toptierlimit.cri-resource-manager.intel.com         | toptierlimit.resource-policy.nri.io         |
+| topologyhints.cri-resource-manager.intel.com        | topologyhints.resource-policy.nri.io        |
+| balloon.balloons.cri-resource-manager.intel.com     | balloon.balloons.resource-policy.nri.io     |
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -224,6 +224,9 @@ def normalizePath(docPath,uriPath):
 def fixLocalMDAnchors(app, doctree, docname):
     for node in doctree.traverse(nodes.reference):
         uri = node.get('refuri')
+        if uri is None:
+             print("fixLocalMDAnchor: skipping anchor with no URI at node: ", node)
+             continue
 
         if isHTTPLink(uri):
             continue
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,6 +17,8 @@ Welcome to CRI Resource Manager's documentation!
 
    developers-guide/index.rst
 
+   migration-to-NRI.md
+
    demos/index.rst
 
    reference/index.md


docs/migration-to-NRI.md
```diff
@@ -0,0 +1,90 @@
...
+## 8. Deploy your pod
+
+## 9. See what resources the container got assigned with:
+ - kubectl exec -c $container $pod -- cat /proc/self/status | grep _allowed_list
+
+Output should be something along the lines of:
+
+Cpus_allowed_list:	10-15 </br>
+Mems_allowed_list:	1
+
```
klihub, line 48: Generally this looks good, but I think we need to adjust and reorganize the content a bit and fill in a few still missing details. See my more detailed train of thoughts below... 

Considering the audience, I wonder if the first section ('Enabling CRI-RM') is really necessary. I think the target audience for this documentation is folks who are already up and running with CRI Resource Manager and would like/need to switch to using one the NRI-retrofitted version of their policy.

For such folks, most of the necessary information to make the switch is already present in the subsequent chapters, but I think it would be good to perhaps structure it slightly differently and add a few missing details/bits. So maybe we could restructure the outline of the description like this:

1. Prerequisites
  -  up and running with CRI Resource Manager
  - one of two supported policies in use: balloons or topology-aware
  - anything else needs to be 'ported', a mechanical task following the example of either balloons or topology-aware
2. Steps for an initial/basic migration test
  - this is basically what you have already described, basically the steps outlined below
  - clone repo, build policies and images
  - enable NRI in runtime (containerd, and I also would describe how to enable it for CRI-O because of the following chapter)
  - import policy image
  - give it a test run, check that things work fine, policy gets applied to sample workloads
3. Steps for a more real-life migration, assuming folks run a self-hosted repository where they can publish images
  - same as the steps for above, but describe what needs to be done to avoid having to manually import images
  - build images
  - push images to self-hosted repo
  - change deployment file to reference image in self-hosted repo
4. Migrating existing configuration
  - the ConfigMap used by the ported policies/infra has a different name/naming scheme than the original one used in CRI-RM, so let's point/spell out the differences
5. Migrating existing workloads
  - the annotation one can use to customize how a policy treats a workload use slightly different keys than the original ones in CRI-RM, so let's point/spell out the differences

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,90 @@
...
+## 6. See the resources the pod got assigned with:
+ - kubectl exec $pod -c $container  -- grep allowed_list: /proc/self/status
+ - Output should look similar to the output of CRI-RM
+
+## Common error:
+NRI socket errors, ex: "failed to start resource manager: failed to start NRI plugin: failed to connect to NRI service: dial unix /var/run/nri.sock: connect: connection refused"
+ - sudo vim /etc/containerd/config.toml
+ - Go to the "io.containerd.nri.v1.nri" and look for "socket_path"
+ - Make sure this matches the socket path of your NRI socket.
+ - systemctl restart containerd
```
klihub, line 90: This error should not occur if the NRI versions used to compile the runtime and the policies are in sync (or at least both are past the version where the default socket path change happened, past commit d4f68f05d880). Considering that, I think this would probably come through nicer, if we point them to use
- at least the freshly released containerd 1.7 (which uses NRI v0.3.0).
- at least the freshly released CRI-O 1.26.2 (which uses NRI v0.3.0).
- to use our 'stable' branch (@jukkar maybe the 'preview' one which then needs updating) which uses at least NRI v0.3.0

klihub:
Another nit: let's get rid of that merge commit and only have the one adding the new docs snippet.

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,113 @@
...
+- ### Remember to change the image name & pull policy in the plugins .yaml file to match your registyr and image, ex:
+  - ```console
+    vim build/images/nri-resmgr-topology-aware-deployment.yaml
+    ```
+
+- ### Then deploy the plugin simlarly to the earlier step.
+
+## 4. Migrating existing configuration
+- The ConfigMap used by the ported policies/infra has a different name/naming scheme than the original one used in CRI-RM, ex:
+  - configMapName: "configmap-name: cri-resmgr-config" is now "configmap-name: nri-resmgr-config"
```
klihub, line 109: I think we should mention here that the details of grouping nodes by labeling to share configuration is also slightly different. What used to be
     `cri-resource-manager.intel.com/group: $GROUP_NAME`
is now
    `nri-resmgr.intel.com/group: $GROUP_NAME`

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,113 @@
...
+
+## 5. Migrating existing workloads
+- The annotations one can use to customize how a policy treats a workload use slightly different keys than the original ones in CRI-RM. ex:
+  - ResmgrKeyNamespace: ResmgrKeyNamespace = "cri-resource-manager.intel.com" is now ResmgrKeyNamespace = "nri-resmgr.intel.com"
```
klihub, line 113: Maybe we could state this slightly differently. I suspect we can't expect folks to make the right mental association based on `ResmgKeyNamespace` unless they are familiar with the cri-resmgr source code. So maybe something along the lines of

The annotations one can use to customize how a policy treats a workload use slightly different keys than the original ones in CRI-RM. The collective 'key namespace' for policy- and resource-manager-specific annotation has been changed from ` cri-resource-manager.intel.com` to `nri-resmgr.intel.com`.

For instance, an explicit type annotation for the `balloons` policy, which used to be
```
...
metadata:
  annotations:
    balloon.balloons.cri-resource-manager.intel.com/container.$CONTAINER_NAME: $BALLOON_TYPE`
...
```

should now be  

```
...
metadata:
  annotations:
    balloon.balloons.nri-resmgr.intel.com/container.$CONTAINER_NAME: $BALLOON_TYPE`
...
```

Similarly a workload opt-out annotation from exclusive CPU allocation for the `topology-aware` policy, which used to be

```
...
metadata:
  annotations:
    prefer-shared-cpus.cri-resource-manager.intel.com/container.$CONTAINER_NAME: "true"
...
```

should now be 

```
...
metadata:
  annotations:
    prefer-shared-cpus.nri-resmgr.intel.com/container.$CONTAINER_NAME: "true"
...
```

Similar changes are needed for any cri-resmgr-specific annotation that uses the same semantic scoping for key syntax.

klihub (approved):
LGTM.

@marquiz @askervin @jukkar @kad PTAL.

fmuyassarov:
It looks good but I there are some cosmetic changes that we can make to make it easier for readiers to follow. Because it was a bit hard to follow for me, but that's because of the organization and formatting, which can easily be fixed I think.
For example, there are places where configuration differs depending on the runtime used. Instead of saying for containerd/cri-o... for multiple steps, there could be two separate sections one for containerd and another for cri-o. 

Another comment, those console steps that come without a comment/explanation in between can be merged into a single console block. I think that's much easier to follow and no need to repeat the console (command) for every command line.

Also, in those sections where we want to describe how cri-o or containerd config file should look like (for example for enbaling nri feature), it would be better to have only that part of the config file available instead of having `sudo vim /etc/containerd/config.toml`.

Here is an example of how I was thinking that we could organize, this is just small part of the document

Screenshot from 2023-03-14 16-41-24

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+
+## See the resources the pod got assigned with
+  ```console
+  $ kubectl exec $pod -c $container  -- grep allowed_list: /proc/self/status
+  # Output should look similar to the output of CRI-RM
+  ```
+
+# Steps for a more real-life migration using self-hosted image repository
+- ### Same steps as above for enabling NRI with Containerd/CRI-O and building the images.
+- ### Load the .tar file into your image repository:
```
klihub, line 108: When markdown-rendered to HTML these look weird, probably because they are both list items and <H3>-headers. I suspect you meant them to be either but not both.

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+  docker load -i build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar
+  ```
+
+- ### Tag the image with your registry:
```
klihub, line 113: Same here...

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+  docker tag myimage:<version> myregistry.com/myimage:<version>
+  ```
+
+- ### Push the image to your registry:
+  ```console
+  docker push myregistry.com/myimage:<version>
+  ```
+
+- ### Remember to change the image name & pull policy in the plugins .yaml file to match your registyr and image, ex:
```
klihub, line 123: Ditto...

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+  vim build/images/nri-resmgr-topology-aware-deployment.yaml
+  ```
+
+- ### Then deploy the plugin simlarly to the earlier step.
```
klihub, line 128: Ditto...

klihub (changes requested):
@luukasmakila There is a bit of a formatting weirdness in one section where list items are also marked as H3-headers. Can you take a look at it and see if it's really intentional ?

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
+# Prerequisities
+- Up and running CRI Resource Manager
+- One of the two supported policies in use: balloons or topology-aware.
+- For other policies a little bit more work is required and the policies need to be 'ported'. This can be done by just following the example of how balloons or topology-aware was converted.
```
marquiz, line 4:
```suggestion
- For other policies a little bit more work is required and the policies need to be 'ported'. This can be done by just following the example of how balloons or topology-aware were converted.
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+    disable_connections = false
+    plugin_config_path = "/etc/nri/conf.d"
+    plugin_path = "/opt/nri/plugins"
+    plugin_registration_timeout = "5s"
+    plugin_request_timeout = "2s"
+    socket_path = "/var/run/nri/nri.sock"
+  ```
+
+## CRI-O
+- Replace the crio version in the system with 1.26.2 or newer.
```
marquiz, line 43: nit: I'd suggest a slight rewording, WDYT?

```suggestion
- Ensure that crio version 1.26.2 or newer is used.
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+
+- Same thing with the Kubelet runtime endpoint, but change it to the crio socket and then restart the crio service.
```
marquiz, line 45: What same thing? I think it's best just to spell out the detailed steps here

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  nri_config_file = "/etc/nri/nri.conf"
+
+  # NRI socket to listen on.
+  nri_listen = "/var/run/nri/nri.sock"
+
+  # NRI plugin directory to use.
+  nri_plugin_dir = "/opt/nri/plugins"
+  ```
+
+## Build the policies
```
marquiz, line 69:
```suggestion
## Build the NRI policies
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+    $ git clone https://github.com/jukkar/nri-resmgr.git
+    $ cd nri-resmgr
+    # Switch to a stable branch
+    $ git checkout preview
+    $ make
+    $ make images
```
marquiz, line 76: Suggest make it slightly shorter with `-b`. Also remove `$` in case it's just commands (that could be copy pasted directly to the shell) without any output printed here. Remove leading whitespace

```suggestion
git clone https://github.com/jukkar/nri-resmgr.git -b preview
cd nri-resmgr
make
make images
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Apply required crds
+```console
+kubectl apply -f deployment/base/crds/noderesourcetopology_crd.yaml
+```
+
+## Import the image of the NRI plugin you want to run
+### Containerd
+  ```console
+  $ ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar
```
marquiz, line 87:
```suggestion
ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar
```

Still that^ might not work as the hash probably changes. Should we change it to `*`-

Also, could/should we omit this step and just advice using a self-hosted registry? Would decrease the number of steps that need to be done on each host manually. What do people think?

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```
+
+### CRI-O
+- See the section [below](#steps-for-a-more-real-life-migration-using-self-hosted-image-repository) for instructions on how to push the images to a registry, then pull from there.
+
+## Then deploy the plugin
+  ```console
+  $ kubectl apply -f build/images/nri-resmgr-topology-aware-deployment.yaml
+  ```
```
marquiz, line 96: Remove `$` and leading whitespace

```suggestion

```console
kubectl apply -f build/images/nri-resmgr-topology-aware-deployment.yaml
```
```

File: docs/migration-to-NRI.md [hunk 16]
```diff
@@ -0,0 +1,182 @@
...
+## Deploy your test pod
+
+## See the resources the pod got assigned with
+  ```console
+  $ kubectl exec $pod -c $container  -- grep allowed_list: /proc/self/status
+  # Output should look similar to the output of CRI-RM
+  ```
+
+# Steps for a more real-life migration using self-hosted image repository
+- Same steps as above for enabling NRI with Containerd/CRI-O and building the images.
```
marquiz, line 107: nit: add an empty line after the heading (ditto many other places too)

```suggestion

- Same steps as above for enabling NRI with Containerd/CRI-O and building the images.
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```
+
+## CRI-O
+- Replace the crio version in the system with 1.26.2 or newer.
+
+- Same thing with the Kubelet runtime endpoint, but change it to the crio socket and then restart the crio service.
+
+- Edit the crio config and look for the "[crio.nri]" section, replace "enable_nri = false" with "enable_nri = true" and uncomment the NRI config lines:
+  ```console
+  $ vim /etc/crio/crio.conf
```
marquiz, line 49:
```suggestion
  vim /etc/crio/crio.conf
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
+# Prerequisities
+- Up and running CRI Resource Manager
+- One of the two supported policies in use: balloons or topology-aware.
+- For other policies a little bit more work is required and the policies need to be 'ported'. This can be done by just following the example of how balloons or topology-aware was converted.
+
+# Steps for an initial/basic migration test
+## Containerd 
+- Replace the containerd version in the system with 1.7 or newer version (NRI server not supported in older versions).
```
marquiz, line 8: nit: I think the bullet list looks a bit funny and doesn't really bring any value (as the section does not have any other content than bullets). I suggest to not use bullets

```suggestion

Replace the containerd version in the system with 1.7 or newer version (NRI server not supported in older versions).
```

Also prolly applies to some other sections here

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+    $ git clone https://github.com/jukkar/nri-resmgr.git
+    $ cd nri-resmgr
+    # Switch to a stable branch
+    $ git checkout preview
+    $ make
+    $ make images
+  ```
+
+## Apply required crds
```
marquiz, line 79:
```suggestion
## Create required CRDs
```

marquiz (commented):
Thanks @luukasmakila for the conribution. Some notes based on a quick read-through.

Also one general ask: could you wrap the lines (of basetext, list items) to something reasonable, e.g. 80 characters. Makes reviewing, reading and commenting a lot easier.

docs/migration-to-NRI.md
askervin, line 87: Remove hash from:
`ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar`
because there will be no such file:
`ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-*.tar`

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+## Import the image of the NRI plugin you want to run
+### Containerd
+  ```console
+  $ ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar
+  ```
+
+### CRI-O
+- See the section [below](#steps-for-a-more-real-life-migration-using-self-hosted-image-repository) for instructions on how to push the images to a registry, then pull from there.
+
+## Then deploy the plugin
```
askervin, line 93: For consistency, replace "Then deploy the plugin" with "Deploy the plugin".

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+  $ kubectl exec $pod -c $container  -- grep allowed_list: /proc/self/status
+  # Output should look similar to the output of CRI-RM
+  ```
+
+# Steps for a more real-life migration using self-hosted image repository
+- Same steps as above for enabling NRI with Containerd/CRI-O and building the images.
+- Load the .tar file into your image repository:
+  ```console
+  docker load -i build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar
```
askervin, line 110: Replace the hash with `*`. Consider loading all plugin images (if `docker load -i` supports multiple image files).

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```console
+  $ ctr -n k8s.io images import build/images/nri-resmgr-topology-aware-image-9797e8de7107.tar
+  ```
+
+### CRI-O
+- See the section [below](#steps-for-a-more-real-life-migration-using-self-hosted-image-repository) for instructions on how to push the images to a registry, then pull from there.
+
+## Then deploy the plugin
+  ```console
+  $ kubectl apply -f build/images/nri-resmgr-topology-aware-deployment.yaml
```
askervin, line 95: This deployment fails:
```
  Warning  FailedMount  5s (x6 over 20s)  kubelet            MountVolume.SetUp failed for volume "resmgrnrisock" : hostPath type check failed: /var/run/nri.sock is not a socket file
```

The problem seems to be that there is `/var/run/nri/nri.sock` on the host instead of `/var/run/nri.sock`. I guess this should be fixed in the deployment.yaml `hostPath`, and probably mount path inside the pod, too. @jukkar, @klihub, WDYT?

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  ```
+
+## Deploy your test pod
+
```
askervin, line 99: Maybe we could use a busybox pod as an example in order to make everything below easy to execute. For instance:
```
kubectl run mypod --image busybox -- sleep inf
```

jonka jälkeen seuraavan komennon voisi yksinkertaistaa:

```
kubectl exec mypod  -- grep allowed_list: /proc/self/status
```

askervin (commented):
Added comments and suggestions.

docs/migration-to-NRI.md
jukkar: If there are multiple tar files in the directory, then the *.tar would probably either fail or import wrong tar file. Best option here is to take the latest .tar file and import that.

docs/migration-to-NRI.md
jukkar: The `/var/run/nri/nri.sock` is the correct one, it requires quite recent containerd to work.

docs/migration-to-NRI.md
fmuyassarov: +1

docs/migration-to-NRI.md
klihub: Yeah, exactly as @jukkar said.

The carelessly chosen original socket default was updated in commit d4f68f05d880/merge 6cd7468afcea. Just make sure you are running with everything compiled against NRI v0.3.0. In runtime terms this simply means containerd v1.7.0 or cri-o v1.26.2.

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,182 @@
...
+  - The details of grouping nodes by labeling to share configuration:
+    - Used to be: </br>
+      cri-resource-manager.intel.com/group: $GROUP_NAME
+    - Is now: </br>
+      nri-resmgr.intel.com/group: $GROUP_NAME
+
+# Migrating existing workloads
+- The annotations one can use to customize how a policy treats a workload use slightly different keys than the original ones in CRI-RM. The collective 'key namespace' for policy- and resource-manager-specific annotation has been changed from cri-resource-manager.intel.com to nri-resmgr.intel.com.
+
+- For instance, an explicit type annotation for the balloons policy, which used to be:
```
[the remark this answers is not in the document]
klihub: > Sure, but im not sure where to find them laughing

Your best bet is to `git grep` for `ResmgrKey` in the source tree, then go through the hits one by one to figure out if it is an annotations and if it is what is the full key. There are not prohibitively many:

```
klitkey1-mobl cri-resource-manager $ git grep ResmgrKey '*.go' | grep -v '_test.go' | grep -v kubernetes/
pkg/agent/flags.go:     flag.StringVar(&opts.labelName, "label-name", kubernetes.ResmgrKey("group"), "Name of the label used to assign a node to a configuration group.")
pkg/cri/resource-manager/cache/cache.go:        RDTClassKey = "rdtclass" + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/cache/cache.go:        BlockIOClassKey = "blockioclass" + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/cache/cache.go:        ToptierLimitKey = "toptierlimit" + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/cache/cache.go:        TopologyHintsKey = "topologyhints" + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/cache/container.go:    return keysInNamespace(c.Labels, kubernetes.ResmgrKeyNamespace)
pkg/cri/resource-manager/cache/container.go:    value, ok := c.Labels[kubernetes.ResmgrKey(key)]
pkg/cri/resource-manager/cache/container.go:    return keysInNamespace(c.Annotations, kubernetes.ResmgrKeyNamespace)
pkg/cri/resource-manager/cache/container.go:    return c.GetAnnotation(kubernetes.ResmgrKey(key), objPtr)
pkg/cri/resource-manager/cache/pod.go:  return keysInNamespace(p.Labels, kubernetes.ResmgrKeyNamespace)
pkg/cri/resource-manager/cache/pod.go:  value, ok := p.Labels[kubernetes.ResmgrKey(key)]
pkg/cri/resource-manager/cache/pod.go:  return keysInNamespace(p.Annotations, kubernetes.ResmgrKeyNamespace)
pkg/cri/resource-manager/cache/pod.go:  return p.GetAnnotation(kubernetes.ResmgrKey(key))
pkg/cri/resource-manager/cache/pod.go:  return p.GetAnnotationObject(kubernetes.ResmgrKey(key), objPtr, decode)
pkg/cri/resource-manager/policy/builtin/balloons/balloons-policy.go:    balloonKey = "balloon." + PolicyName + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/dynamic-pools/dyp.go:   dynamicPoolKey = "dynamic-pool." + PolicyName + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/podpools/podpools-policy.go:    podpoolKey = "pool." + PolicyName + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/static/static-policy.go:               kubernetes.ResmgrKey(keyPreferIsolated))
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:     preferIsolatedCPUsKey = keyIsolationPreference + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:     preferSharedCPUsKey = keySharedCPUPreference + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:     preferMemoryTypeKey = keyMemoryTypePreference + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:     preferColdStartKey = keyColdStartPreference + "." + kubernetes.ResmgrKeyNamespace
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:     preferReservedCPUsKey = keyReservedCPUsPreference + "." + kubernetes.ResmgrKeyNamespace

```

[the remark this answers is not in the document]
klihub: Actually, that `git grep` above is unfortunately not enough. You'll also need to at least check these ones too:
```
klitkey1-mobl cri-resource-manager $ git grep -n '\.GetResmgrAnnotation' '*.go' | grep -v _test.go
pkg/cri/resource-manager/cache/pod.go:388:      value, ok := p.GetResmgrAnnotation(keyAffinity)
pkg/cri/resource-manager/cache/pod.go:398:      value, ok = p.GetResmgrAnnotation(keyAntiAffinity)
pkg/cri/resource-manager/policy/builtin/static-plus/static-plus-policy.go:356: if value, ok := pod.GetResmgrAnnotation(keyPreferIsolated); ok {
pkg/cri/resource-manager/policy/builtin/static/static-policy.go:311:           if value, ok := p.GetResmgrAnnotation(keyPreferIsolated); ok {
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:213: value, ok := pod.GetResmgrAnnotation(key)
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:250: value, ok := pod.GetResmgrAnnotation(key)
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:296: value, ok := pod.GetResmgrAnnotation(key)
pkg/cri/resource-manager/policy/builtin/topology-aware/pod-preferences.go:490: value, ok := pod.GetResmgrAnnotation(key)
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,228 @@
...
+nri_listen = "/var/run/nri/nri.sock"
+
+# NRI plugin directory to use.
+nri_plugin_dir = "/opt/nri/plugins"
+```
+
+## Build the NRI policies
+
+```console
+git clone https://github.com/jukkar/nri-resmgr.git -b preview
```
jukkar, line 91: The code was moved to
https://github.com/containers/nri-plugins
so this link can be changed to
```
git clone https://github.com/containers/nri-plugins
```

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,227 @@
...
+enable_nri = true
+
+# NRI configuration file to use.
+nri_config_file = "/etc/nri/nri.conf"
+
+# NRI socket to listen on.
+nri_listen = "/var/run/nri/nri.sock"
+
+# NRI plugin directory to use.
+nri_plugin_dir = "/opt/nri/plugins"
```
klihub, line 85: II think it would be better to either use crio itself to enable NRI (so it injects any defaults as comments as well)

```
CRIO_CONF=/etc/crio/crio.conf
cp $CRIO_CONF $CRIO_CONF.orig
crio --enable-nri config > $CRIO_CONF
```

or to use a drop-in fragment. I *think* something like this should work:

```
cat > /etc/crio/crio.conf.d/10-enable-nri.conf <<EOF
[crio.nri]
enable_nri = true
EOF
```

The problem with the above instructions is that they can easily break something in a future setup if the CRI-O NRI configuration change (especially any defaults in there).

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,227 @@
...
+| prefer-isolated-cpus.cri-resource-manager.intel.com | prefer-isolated-cpus.nri-resmgr.intel.com |
+| prefer-shared-cpus.cri-resource-manager.intel.com   | prefer-shared-cpus.nri-resmgr.intel.com   |
+| memory-type.cri-resource-manager.intel.com          | memory-type.nri-resmgr.intel.com          |
+| cold-start.cri-resource-manager.intel.com           | cold-start.nri-resmgr.intel.com           |
+| prefer-reserved-cpus.cri-resource-manager.intel.com | prefer-reserved-cpus.nri-resmgr.intel.com |
+| rdtclass.cri-resource-manager.intel.com             | rdtclass.nri-resmgr.intel.com             |
+| blockioclass.cri-resource-manager.intel.com         | blockioclass.nri-resmgr.intel.com         |
+| toptierlimit.cri-resource-manager.intel.com         | toptierlimit.nri-resmgr.intel.com         |
+| topologyhints.cri-resource-manager.intel.com        | topologyhints.nri-resmgr.intel.com        |
+| balloon.balloons.cri-resource-manager.intel.com     | balloon.balloons.nri-resmgr.intel.com     |
```
klihub, line 227: In commit bdbb668186a1ab244aca166564bea1c5bfd7c147 @jukkar updated these from `nri-resource-policy.intel.com` to `resource-policy.nri.io`. We need to update this table accordingly.

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,212 @@
...
+
+- Then deploy the plugin simlarly to the earlier step.
+
+# Migrating existing configuration 
+
+- The ConfigMap used by the ported policies/infra has a different name/naming scheme than the original one used in CRI-RM, ex:
+  - configMapName:
+    ```diff
+    - configmap-name: cri-resmgr-config
+    + configmap-name: nri-resmgr-config
```
klihub, line 143: AFAICT, this has been changed to `nri-resource-policy-config` in the [same commit](https://github.com/containers/nri-plugins/commit/bdbb668186a1ab244aca166564bea1c5bfd7c147) as the annotations.

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,212 @@
...
+# Get the Kubelet args
+systemctl cat kubelet <- Look for: EnvironmentFile=/.../kubeadm-flags.env
+
+vim /.../kubeadm-flags.env
+  KUBELET_KUBEADM_ARGS="--container-runtime-endpoint=unix:///var/run/containerd/containerd.sock --pod-infra-container-image=registry.k8s.io/pause:3.9"
+
+vim /etc/sysconfig/kubelet 
+  KUBELET_EXTRA_ARGS= --container-runtime-endpoint=/var/run/containerd/containerd.sock <- Remember this aswell
+
+systemctl restart containerd
```
klihub, line 26: nit: this should be done after the containerd configuration has been updated.

docs/migration-to-NRI.md
klihub: Shouldn't be a problem. The commented lines cri-o outputs for the `config` subcommand are its defaults. So you don't need to uncomment them unless you want to change the configuration to diverge from the defaults.

klihub: But if you do uncomment them, then you are not running any more with the defaults. You are running with values which are explicitly set to the same as the current default values. If the defaults change, then in the former case you will still run with the updated defaults, but in the latter case you will keep running with your forced values which used to be the defaults but are not any more. So there is a difference and (usually) therefore if the defaults (at any point in time) are fine for you, you shouldn't uncomment them in the config.

luukasmakila: Thanks for clearing that up!

klihub (dismissed):

klihub:
Okay, looks like one more thing needs to be taken care of. You need to point to this migration documentation from the existing docs, otherwise our doc builder will complain. That is the reason of the test build failure.

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,214 @@
...
+
+Replacing the runtime endpoint on a node that was setup using Kubeadm:
+```console 
+# Get the Kubelet args
+systemctl cat kubelet <- Look for: EnvironmentFile=/.../kubeadm-flags.env
+
+vim /.../kubeadm-flags.env
+  KUBELET_KUBEADM_ARGS="--container-runtime-endpoint=unix:///var/run/crio/crio.sock --pod-infra-container-image=registry.k8s.io/pause:3.9"
+
+vim /etc/sysconfig/kubelet
```
klihub, line 61: nit: extra trailing space...

docs/migration-to-NRI.md
```diff
@@ -0,0 +1,214 @@
...
+  ```
+
+- Remember to change the image name & pull policy in the plugins .yaml file to match your registyr and image, ex:
+  ```console
+  vim build/images/nri-resmgr-topology-aware-deployment.yaml
+  ```
+
+- Then deploy the plugin simlarly to the earlier step.
+
+# Migrating existing configuration
```
klihub, line 139: nit: extra trailing space...

docs/migration-to-NRI.md
jukkar: > or to use a drop-in fragment. I _think_ something like this should work:
> 
> ```
> cat > /etc/crio/crio.conf.d/10-enable-nri.conf <<EOF
> [crio.nri]
> enable_nri = true
> EOF
> ```
> 
> The problem with the above instructions is that they can easily break something in a future setup if the CRI-O NRI configuration change (especially any defaults in there).

BTW, I used this variant in the nri-plugins repo e2e tests, and cri-o worked just fine with nri resource plugin.

klihub:
@luukasmakila If you try to build, serve, and look at the resulting documentation, there are few remaining problems. You can check them out by running `make site-serve` in your working branch/tree then pointing your browser at the [locally served copy]( http://127.0.0.1:8081/) from your tree. I am really sorry for none of us being awake enough earlier to point to this most important and helpful piece of information for you. PTAL.

The problems are
- Sphinx (mysteriously) does not like your in-doc forward ref to the real-life migration chapter => add workaround
- the doc is considered dangling as it is not reachable from the ToC/index => link it in to ToC/index
- the current header structure is not deep enough and clobbers/contaminates the ToC/index => deepen it

Here is a [sample git branch](https://github.com/klihub/cri-resource-manager/tree/docs/migration-fixes) with sample set of commits solving these.

klihub (approved):
LGTM.

jukkar (approved):
LGTM

PR #990, opened 2023-03-27
[WIP] Add security-scanning with SNYK

oxxenix:
Adding SNYK and CodeQL security scanning jobs.

### DIFF
--- a/.github/workflows/security-scanning.yaml
+++ b/.github/workflows/security-scanning.yaml
@@ -1,0 +1,72 @@
+name: Security Scanning
+
+on:
+  push:
+    branches: [ master ]
+    tags: [ '*' ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+    environment:
+      name: dev
+      url: https://github.com
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+
+    - name: Set Up Snyk
+      uses: snyk/actions/setup@master
+
+    - name: Run Snyk to check for vulnerabilities
+      continue-on-error: true
+      uses: snyk/actions/golang@master
+      env:
+        SNYK_TOKEN: ${{ env.SNYK_TOKEN }}
+      with:
+        args: --sarif-file-output=snyk.sarif
+
+    - name: Upload result to GitHub Code Scanning
+      if: ${{ github.event_name != 'pull_request' }}
+      uses: github/codeql-action/upload-sarif@v2
+      with:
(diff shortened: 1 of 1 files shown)


kad:
I've added needed secret to the settings.

.github/workflows/security-scanning.yaml
```diff
@@ -0,0 +1,71 @@
...
+      uses: snyk/actions/setup@master
+
+    - name: Run Snyk to check for vulnerabilities
+      continue-on-error: true
+      uses: snyk/actions/golang@master
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      with:
+        json: true
+        args: --sarif-file-output=snyk.json
```
klihub, line 31: I think using an absolute path here (perhaps `/tmp/snyk.json`) or then `sarif_file: /github/workspace/snyk.json` later might work around the `upload-sarif` failures.

But maybe something like this would work and be easier to read, too:

```
...

env:
    SARIF_FILE_PATH: /tmp/snyk.json

jobs:
...

    - name: Run Snyk to check for vulnerabilities
      continue-on-error: true
      uses: snyk/actions/golang@master
      env:
        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
      with:
        json: true
        args: --sarif-file-output=${{ env.SARIF_FILE_PATH }}

    - name: Upload result to GitHub Code Scanning
      uses: github/codeql-action/upload-sarif@v2
      with:
       sarif_file: ${{ env.SARIF_FILE_PATH }}
...

```

klihub (commented):
The original problem of `snyk.json` not being found still seems to be there.

I think it might come from the snyk run action being `/usr/bin/docker run --name snyksnykgolang_69232c --label 6c0442 * --workdir /github/workspace *...` and because of that `snyk.json` being created as `/github/workspace/snyk.json` while the upload action looking for it as `snyk.json` with another working directory (maybe a default `/home/runner`).

.github/workflows/security-scanning.yaml
```diff
@@ -0,0 +1,74 @@
...
+      uses: actions/checkout@master
+
+    - name: Set Up Snyk
+      uses: snyk/actions/setup@master
+
+    - name: Run Snyk to check for vulnerabilities
+      continue-on-error: true
+      uses: snyk/actions/golang@master
+      env:
+        SNYK_TOKEN: ${{ env.SNYK_TOKEN }}
```
klihub, line 31: nit: is this still needed ? I think we have now this set up globally.

PR #995, opened 2023-04-04
topology-aware: fix logging of cpuset changes.

klihub:
Fix logging changes to allowed and reserved cpuset when configuration is updated. The cpusets were logged the wrong way around.


jukkar (approved):

marquiz (approved):
LGTM

PR #999, opened 2023-04-17
dockerfiles: (re)name suse Dockerfile as expected.

klihub:
Building native distro packages with 'make packages' expects the distro-specific Dockerfile to have the '-opensuse-leap-$VERSION' suffix  when building on opensuse, since this is how the distro names itself in /etc/os-release. (Re)name the Dockerfile to be as expected.


marquiz:
Sorry, does this fix some issue? Seems to work fine for me, e.g.`make cross-rpm.suse-15.4` generated an rpm as expected

marquiz (dismissed):
LGTM

klihub:
> > Sorry, does this fix some issue? Seems to work fine for me, e.g.`make cross-rpm.suse-15.4` generated an rpm as expected
> 
> Yes. Clone the repo on any machine running opensuse and try 'make packages'. It fails without this commit and succeeds with this one present in the tree.

And the reason is that the distro calls itself opensuse in `/etc/os-release`.

jukkar (approved):

klihub:
Hmm... but you are right that there might be a better/easier fix for that. And I think this might break some (packaging) tests without further changes.

klihub:
> Hmm... but you are right that there might be a better/easier fix for that. And I think this might break some (packaging) tests without further changes.

Updated commits with 2 related fixes/adjustments, one in the docs and another in a packaging test case.

marquiz (approved):
LGTM

PR #1000, opened 2023-04-17
[release-0.8] dockerfiles: (re)name suse Dockerfile as expected.

klihub:
[cherry-picked #999/a54f3165]

Building native distro packages with 'make packages' expects the distro-specific Dockerfile to have the '-opensuse-leap-$VERSION' suffix  when building on opensuse, since this is how the distro names itself in /etc/os-release. (Re)name the Dockerfile to be as expected.


marquiz (approved):
LGTM (hold until #999 is merged)

marquiz (approved):
LGTM

PR #1002, opened 2023-04-19
Make findContainerDir support docker runtime

changzhi1990:


klihub:
@changzhi1990 We deliberately decided to not support docker/cri-dockerd (it has been deprecated for ages and suppor removed from kubelet quite a while ago). Why do you want to add it back here ?

PR #1006, opened 2023-04-24
Enable fixing balloon instances to hardcoded CPUs and memories for benchmarking

askervin:
These patches add support for `BalloonInstances` configuration. It enables setting constraints and overrides on individual balloon instances regarding CPU allocation and memory pinning.

This feature is useful for benchmarking well-known hardware under different loads. However, using this feature makes the balloons policy configuration totally inapplicable to heterogenous clusters and very prone to configuration errors. These issues are likely to manifest themselves by not being able to run pods that have been scheduled on a node. Therefore this feature is not available by default, but only if `cri-resmgr` is built with `benchmarking` build tag (or `make BENCHMARKING=1`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,15 @@ RANDOM_ID := "$(shell head -c20 /dev/urandom | od -An -tx1 | tr -d ' \n')"
 
 ifdef STATIC
     STATIC_LDFLAGS:=-extldflags=-static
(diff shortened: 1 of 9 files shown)


askervin:
Thanks @jukkar, good point. I added both "why" and "why not" reasoning on the commit that introduces `BalloonInstances` to the configuration.

askervin:
Let's close this PR for now.

PR #1022, opened 2023-08-11
github: take golang version from go.mod

marquiz:
Fix the verify workflow to actually install the version of Go specified
in go.mod. Previously, because of the yaml parsing peculiarities go v1.2 was
installed.

Also changes the order of steps so that go.mod is available.

> **NOTE:** This expects that the practice from now on is to keep the Go version
> in go.mod and Makefile in sync.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -7,14 +7,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go
-      uses: actions/setup-go@v1
-      with:
-        go-version: 1.20
-      id: go
-
     - name: Check out code into the Go module directory
       uses: actions/checkout@v1
+
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version-file: go.mod
+      id: go
 
     - name: Install golangci-lint
       run: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.51.2


.github/workflows/verify.yml
```diff
@@ -8,9 +8,9 @@ jobs:
     steps:
 
     - name: Set up Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v4
       with:
-        go-version: 1.20
+        go-version: "1.20"
```
klihub, line 13: I'm, fine with this. One alterntive I've seen (in ttrpc) is something like:
```
env:
  GO_VERSION: 1.20.x
...
     - uses: actions/setup-go@v3
        with:
          go-version: ${{ env.GO_VERSION }}
...
```

marquiz: I don't have strong opinions. But I probably prefer to keep things simple, like now, as we only use the go version in one place.

mythi: One more alternative is to configure setup-go to read from go.mod

marquiz: Hmm, ok,, so it seems, didn't know about that. The only caveat is that the version in go.mod is a different thing than the version we're building with. But in our case (as we don't expect cri-resmgr to be used as a library/module dependency for any other project) we can probably have a practice to keep go.mod in sync with the golang version we use to build our binaries

marquiz:
Changed the PR to do what @mythi suggested i.e. take Go version from the go.mod file. I think this is the least error-prone method for us, keeping the golang version in sync everywhere

.github/workflows/verify.yml
klihub: > I don't have strong opinions. But I probably prefer to keep things simple, like now, as we only use the go version in one place.

Sorry, my comment was unclear because the example took the version from the environment. But what I meant was that using 1.20.x is another way to avoid the (lack of) quoting problem. Anyway, I think taking it directly from go.mod is probably even better.

klihub (approved):
LGTM

klihub (approved):

PR #1025, opened 2023-08-11
docs: use ADD in the dockerfile to fetch go tarball

marquiz:
Speeds up subsequent builds as the tarball will be cached.


klihub (approved):

marquiz:
rebased...

PR #1026, opened 2023-08-11
Makefile: prepare binary packages as release assets

marquiz:
When cross-building binary packages (rpm, deb, tarball), create a separate packages/release-assets directory where all binary packages are copied with a distintive distro-specific name. Make cutting releases way easier by removing the need for manual renaming (because without the distro-specific naming package names would clash). Also opens up the possibility to automatic uploading of release assets.


klihub (approved):
LGTM.

PR #1027, opened 2023-08-11
cross-build: get golang binaries instead of compiling from source

marquiz:
Speed up builds considerably by reducing the download times (the golang tarball will be cached by docker) and skipping the build of golang itself.


klihub (approved):
LGTM.

marquiz:
Rebased and fixed a typo in the commit message

PR #1034, opened 2023-08-14
fixes: fix a few innocent codeql alert.

klihub:
Fix the following rather innocent two codeql alerts:
- cache: escape '.' in topology path regexp
- cgroupstats: node to int-sized integer (as opposed to int64) before casting to int


marquiz (approved):
LGTM

askervin (approved):
LGTM

PR #1037, opened 2023-08-15
github: add workflows for image building

marquiz:
Add two new workflows, "Release" and "Publish devel images". The first one is intended for building and publishing all release artefacts - we now add only container images. The second one is for image-publishing only from the master branch.

The patch also adds a common re-usable image building job.


klihub (approved):

marquiz:
We could give this a try

PR #1038, opened 2023-08-16
Makefile: drop unwanted update-workflows target

marquiz:
We definitely don't want 'make build' to be mangling our code tree. Plus, this is useless as we take the golang version from go.mod nowadays.


klihub (approved):

marquiz:
Small change to the commit message

PR #1049, opened 2023-08-21
e2e: clean up vm as the last step in the static-pools test suite

askervin:
Without this patch static-pools test suite leaves webhook and cmk node taint behind which will cause next tests on the same vm fail.


klihub (approved):

marquiz (approved):
LGTM

PR #1050, opened 2023-08-21
e2e: fix topology-aware/n4c16/test09-container-exit test

askervin:
"sleep" in the latest busybox does not create a child process named "sleep". Change process matching from sleep to matching pod0c0 parent/child instead.

### DIFF
--- a/test/e2e/policies.test-suite/topology-aware/n4c16/test09-container-exit/code.var.sh
+++ b/test/e2e/policies.test-suite/topology-aware/n4c16/test09-container-exit/code.var.sh
@@ -8,7 +8,13 @@ pyexec 'assert "pod0c0" in allocations'
 
 out '### Crash and restart pod0c0'
 vm-command "kubectl get pods pod0"
-vm-command "kill -KILL \$(pgrep -f pod0c0)"
+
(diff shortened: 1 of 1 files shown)


test/e2e/policies.test-suite/topology-aware/n4c16/test09-container-exit/code.var.sh
```diff
@@ -8,7 +8,13 @@ pyexec 'assert "pod0c0" in allocations'
 
 out '### Crash and restart pod0c0'
 vm-command "kubectl get pods pod0"
-vm-command "kill -KILL \$(pgrep -f pod0c0)"
+
+vm-command "set -x; [[ -n \"\$(pgrep -f pod0c0)\" ]] && [[ \"\$(pgrep -f pod0c0 --oldest)\" != \"\$(pgrep -f pod0c0 --newest)\" ]]" || {
```
klihub, line 12: Is that `set -x` needed/wanted or just a debugging leftover ?

askervin: `set -x` is there by purpose. Similarly to "kubectl get pods pod0" above, it is not strictly necessary but helps figuring out pids in the system from the test output, should the test fail again in the future.

marquiz (approved):
lgtm

PR #1058, opened 2023-08-22
e2e: use bridge cni plugin by default

marquiz:
Bridge networking is stupid and simple and works perfectly in our one node cluster setup. More complicated CNI setup brings no value in terms of testing cri-resmgr. Using bridge speeds up tests considerably (by eliminating all clilium/weavenet download and initialization time) plus should virtually eliminate all test flakyness related to CNI plugin failures.


klihub (approved):

askervin (approved):
Cool! Thanks @marquiz!

PR #1059, opened 2023-08-23
e2e: add --wait to kubectl delete of pods and namespaces

marquiz:
Add --wait and --ignore-missing (to make sure that we really wait for the objects that does exist) to prevent naming collisions caused by long termination times of pods. Also add --now where it is missing so that stuff gets deleted quickly.


marquiz:
ping @klihub @askervin

klihub (approved):

marquiz:
I simplified this a bit, not adding `--ignore-missing` if `--all` is specified

askervin (approved):
LGTM

klihub (approved):

PR #1060, opened 2023-08-23
e2e: check default serviceaccount in cluster readiness check

marquiz:
In some cases there seems to be a race that the node is in ready condition but pod creation still fails because of default service account (in the default namespace) is missing.


marquiz:
ping @klihub @askervin

klihub (approved):

askervin (approved):
LGTM

PR #1062, opened 2023-08-23
e2e: fix "ambiguous allowed resources" caused by race

askervin:
If cri-resmgr changes CPU pinning at the same time when allowed resources of container's processes are read from /proc/PID/status, it may happen that different processes in the same cgroup show different CPU/memory masks. This patch fixes the race by retrying reading the resources few times. If the problem persists, it is most likely caused by multiple matching containers running in the vm, possibly created manually for debugging or by previously failed/aborted tests.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -222,16 +222,7 @@ get-py-allowed() {
         }
         echo -e "$COMMAND_OUTPUT" > "$topology_dump_file"
     fi
-    # Fetch data and update allowed* variables from the virtual machine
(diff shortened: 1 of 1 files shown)


klihub (approved):

test/e2e/run.sh
```diff
@@ -297,6 +288,29 @@ cpu_ids = lambda i: set_ids(i, '[cpu]')
 "
 }
 
+get-res-allowed() {
+    local res_allowed_file="$1"
+    local retries=5
+    while (( retries > 0 )); do
```
marquiz, line 294: Thanks @askervin! Looks good. One suggestion: could you add a comment above the retry-loop here rationalizing why retries are needed here (ie.g. something similar than the commit message)

PR #1064, opened 2023-08-23
e2e: drop vm-force-restart()

marquiz:
Drop the vm-force-restart() function and just use vm-reboot() instead.
Improve the vm-reboot() to do a forced restart (via Qemu command) if the
normal reboot fails - something that the old vm-force-restart() tried to
do. However, the old vm-force-restart() was broken as we always
unconditionally forced a qemu system resstart independent of whether the
normal shutdown succeeded or not. This led to races in rare cases where
the qemu command was executed precisely when the container was in not
running state. Not to mention that this practice never resulted in
graceful restarts.

Tha patch increases the sleep after reboot command to 10 seconds to
match what vm-force-restart() had, apparently to avoid inadvertent ssh
logins to a shutting down system (before ssh daemon is killed and actual
reboot happens).

### DIFF
--- a/demo/lib/vm.bash
+++ b/demo/lib/vm.bash
@@ -616,25 +616,10 @@ vm-reboot() { # script API
     # Reboots the virtual machine and waits that the ssh server starts
     # responding again.
     vm-command "reboot"
-    sleep 5
-    host-wait-vm-ssh-server
-}
-
-vm-force-restart() { # script API
-    # Usage: vm-force-restart
-    #
-    # Give the virtual machine a chance to shut itself down, then
-    # forcibly restart it using govm stop/start. Wait for the ssh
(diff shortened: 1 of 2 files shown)


demo/lib/vm.bash
```diff
@@ -630,9 +630,11 @@ vm-force-restart() { # script API
     # and fall back to vm-reboot instead.
     if vm-is-govm; then
         vm-command "shutdown -h now"
```
klihub, line 632: Shouldn't this be 'shutdown -r now' ? Otherwise I'd expect the ssh wait test to always fail and the reboot happen by reset.

klihub (commented):
One minor question related to '-h' vs. '-r' shutdown option.

klihub (approved):

askervin (approved):
LGTM

PR #1065, opened 2023-08-24
e2e: ignore terminated processes on fetching allowed resources

askervin:
Fetching allowed resources (needed by "verify" and "report allowed") caused test failures when a process found by pgrep got terminated before reading its /proc/pid/files. This change omits pids whose files could not be read without errors.


marquiz (approved):
LGTM

PR #1066, opened 2023-08-24
e2e: don't use pyexec in tests directly

marquiz:


marquiz:
Another one that helps fighting flakyness if/when we implement retry for the verify() function
ping @askervin @klihub

klihub (approved):

PR #1063, opened 2023-08-23
e2e: fix flakiness in dynamic-pools test03-rebalancing

askervin:
Optimize verifications, make first verifications quickly without long stabilizing periods, retry if needed.

PR #1067, opened 2023-08-24
e2e: re-try verify command

marquiz:
Add a simple and stupid unparameterized 5 second re-try loop to the verify() command. This should mitigate test flakyness on slow (or heavily loaded) systems.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -765,6 +765,17 @@ verify() { # script API
     #          nodes and that pod0c0 is allowed to run on 4 CPUs:
     #   verify 'set.intersection(nodes["pod0c0"], nodes["pod1c0"]) == set()' \
     #          'len(cpus["pod0c0"]) == 4'
+    local retries=5
+    while (( retries > 0 )); do
+        ( __verify "$@" ) && return 0
+        retries=$(( retries - 1 ))
+        echo "re-trying verify $@, $retries tries left..."
+        sleep 1s
+    done
+    error "failed to verify $@"
+}
+
+__verify() {
     get-py-allowed
     get-py-cache
     for py_assertion in "$@"; do


marquiz:
ping @askervin @klihub 

This PR sort of generalizes (and mostly replaces) #1063, as I was seeing similar issues on other tests as well on some machines.

klihub (approved):

askervin (commented):
@marquiz, do you think we could take https://github.com/intel/cri-resource-manager/pull/1069 instead?

I'd prefer to include explicit `--retry X` on verifications where we hope situation to get better soon, and we are fine trying it out many times.

Another thing is executing `on_verify_fail` hook, that should probably be executed only in the case where we have decided to fail the whole verify instead of each retry.

Both of these matters are debatable...

test/e2e/run.sh
```diff
@@ -765,6 +765,17 @@ verify() { # script API
     #          nodes and that pod0c0 is allowed to run on 4 CPUs:
     #   verify 'set.intersection(nodes["pod0c0"], nodes["pod1c0"]) == set()' \
     #          'len(cpus["pod0c0"]) == 4'
+    local retries=5
```
klihub, line 768: #1063 wraps implement an explicit configurable `retry` and wraps some of the `verify` calls in that. Does not make sense to do both, so we need to chose either. I am fine with both approaches. If a retrying verify is often necessary then perhaps this approach is slightly preferred. But maybe we should make it parametrizable with something like this:
```bash
verify() {
    local retries=${retries:-1} poll_delay=${poll_delay:=0s}
    # if we want to default to a retrying verify then we just need to set the defaults up so, eg.
    # local retries=${retries:-5} poll_delay=${poll_delay:=0.2s}
    while (( retries > 0 )); do
        ( __verify "$@" ) && return 0
        retries=$(( retries - 1 ))
        echo "re-trying verify $@, $retries tries left..."
        sleep $poll_delay
    done
    error "failed to verify $@"
}

# Then you can do a non-retrying/default verify like
verify 'len(cpus["pod0c0"]) == 4'
# and a corresponding retrying verify with something like
retries=5 poll_delay=0.2s verify 'len(cpus["pod0c0"]) == 4'
```

marquiz: > If a retrying verify is often necessary then perhaps this approach is slightly preferred

This is always needed. Not just in one test

marquiz:
> @marquiz, do you think we could take #1069 instead?

We can take, although I'm not sure know how much value the --retry parameter brings. I'd say we need to have retry 3 times (at least) for all tests.
 
> I'd prefer to include explicit `--retry X` on verifications where we hope situation to get better soon, and we are fine trying it out many times.

It's very hard to predict where `--retry 0´ would be ok and where we need `--retry 3`. It just so much depends on the random factors of the execution environment at the time. It's just futile battle if we start adding --retry parameter every time an assertion randomly fails somewhere.

Re-trying shouldn't cause any harm. If situation stabilises fast then the test continues immediately (without retries)

marquiz:
I take my words back. Let's do #1069 with greater thatn zero default retry count

test/e2e/run.sh
klihub: @marquiz No, I don't think it is always necessary. For instance for most basic workload placement tests it should not be needed. There you typically schedule a few pods (waiting for them to be running) then test allocations. I think in those cases by the time the shell regains control after the CRI requests returns everything should be done on the runtime side. The asynchronous race should be present only in cases when you react to a policy decision making event/action, for instance based on a logged message in the coldstart case, and then start checking resources...

Anyway, regardless of this I'm fine with always retrying by default.

marquiz: OK, maybe so. Anyway, I don't want to spend time (and I'm not sure anyone else does) on combing all the test assertions to figure out whether `--retry` is possibly needed or not. Just have retry everywhere by default.  It shouldn't do any harm or cause extra delays. Then add `--retry 0` to those cases you're absulutely sure that retries won't be necessary (and shouldn't happen)

PR #1072, opened 2023-10-05
e2e: fix starting a custom containerd version

askervin:
This patch generates new default containerd configuration file when a locally built containerd is installed from containerd_src. This fixes the problem where containerd configuration provided by the Linux distribution does not work with the containerd build defined by the user.


fmuyassarov (approved):

marquiz (approved):
LGTM

PR #1077, opened 2023-11-03
test/e2e: fix incorrect Allocat[ion|or]Priority config key.

klihub:
Fix incorrect config key in some balloons tests. Use more consistent test environment variables wrt. the actual config key.


askervin (commented):
Do you think we could do `s/ALLOCATIONPRIORITY/ALLOCATORPRIORITY/g` with the same commit, too?

klihub:
> Do you think we could do `s/ALLOCATIONPRIORITY/ALLOCATORPRIORITY/g` with the same commit, too?

To have consistent naming in the env vs. config ? Yeah, probably a good idea... I'll update.

askervin (approved):
LGTM

PR #1082, opened 2024-01-05
Bump golangci-lint to v1.55.2

marquiz:
The second commit patches the code to pass muster:

- drop unused function arguments
- rename unused function args (that cannot be dropped) to `_`
- rename variables that would override builtin names (`print`, `close`, `new`)
- remove unnecessary `else`


klihub (approved):
LGTM.

askervin (approved):
Wow. Didn't catch anything suspicious. LGTM.

PR #1083, opened 2024-01-05
pkg/sysfs: drop discovery flags

marquiz:
Simplify the code by dropping unused functionality. Effectively, we always used the default flags.

Depends on #1082

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -41,26 +41,6 @@ const (
 	sysfsNumaNodePath = "devices/system/node"
 )
 
-// DiscoveryFlag controls what hardware details to discover.
-type DiscoveryFlag uint
-
-const (
-	// DiscoverCPUTopology requests discovering CPU topology details.
-	DiscoverCPUTopology DiscoveryFlag = 1 << iota
-	// DiscoverMemTopology requests discovering memory topology details.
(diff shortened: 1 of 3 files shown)


pkg/sysfs/system.go
```diff
@@ -625,15 +581,6 @@ func (sys *system) discoverCPU(path string) error {
 
 	sys.cpus[cpu.id] = cpu
 
-	if (sys.flags & DiscoverCache) != 0 {
-		entries, _ := filepath.Glob(filepath.Join(path, "cache/index[0-9]*"))
-		for _, entry := range entries {
-			if err := sys.discoverCache(entry); err != nil {
```
klihub, line 631: nitpick: If we're eliminating unused code (this way) then we should also remove `discoverCache()`.

klihub: Especially since it is broken (recently fixed in nri-plugins).

marquiz: Ach, thanks for pointing this out. Changed now (i.e. `discoverCache` dropped)

klihub (dismissed):
LGTM.

klihub (approved):
LGTM.

askervin (approved):
LGTM

PR #1085, opened 2024-01-09
pkg/cri: implement container event relaying.

klihub:
This PR implements relaying the stream of events for CRI streaming GetContainerEvents() calls.

### DIFF
--- a/pkg/cri/relay/runtime-service.go
+++ b/pkg/cri/relay/runtime-service.go
@@ -16,9 +16,9 @@ package relay
 
 import (
 	"context"
+	"fmt"
+	"time"
 
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 	criv1 "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	"github.com/intel/cri-resource-manager/pkg/dump"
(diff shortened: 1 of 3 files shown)


pkg/cri/relay/runtime-service.go
```diff
@@ -206,3 +221,80 @@ func (r *relay) RuntimeConfig(ctx context.Context, req *criv1.RuntimeConfigReque
...
+		return nil
+	}
+
+	c, err := r.client.GetContainerEvents(context.Background(), req)
+	if err != nil {
+		return fmt.Errorf("failed to create container event client: %w", err)
+	}
+
+	r.evtClient = c
+	go r.relayEvents()
```
askervin, line 263: I don't know if we need to support multiple `GetContainerEvents()` with different `req`s on the same relay that might be calling two `startEventRelay()`s concurrently.

If we don't, maybe we should not allow overwriting `r.evtClient` and launching the relaying goroutine.

If we do, we might want to handle relaying here a bit differently.

Do we need separate goroutine If we need a separate goroutine `r.relayEvents()` to receive events from each client? Then it would be safer to call `go r.relayEvents(c)` and try to manage without `r.evtClient` that get overwritten on line 262.

askervin (commented):
Raising one concern that might be just because of not understanding the big picture...

pkg/cri/relay/runtime-service.go
marquiz: But this shouldn't happen because the first one added `r.evtClient` and the second req bails out before starting the goroutine. Or did I read this right?

askervin: Yes, of course. Thanks @marquiz. I'm dazzled by my blindness.

askervin (approved):
LGTM

PR #1097, opened 2024-08-22
github: only run periodic scan weekly

marquiz:


marquiz:
ping @klihub @askervin

PR #1101, opened 2024-09-03
Drop support for CentOS

marquiz:
Both CentOS 7 and CentOS 8 have been EOL'd and our package builds fail because of that.


klihub (approved):

askervin (approved):
LGTM

PR #1102, opened 2024-09-03
Bump Golang to v1.22

marquiz:
Version 1.21 has been archived.


askervin (commented):
Could we bump GO_VERSION in demo/lib/distro.bash, too?

marquiz:
> Could we bump GO_VERSION in demo/lib/distro.bash, too?

Woops, sure, that was badly outdated. Now updated

klihub (approved):

askervin (approved):
LGTM, thanks @marquiz!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/intel/cri-resource-manager
 
-go 1.21
+go 1.22
```
mythi, line 3: since Go 1.21 this means the minimum required go version which follows the new versioning. you probably want 1.22.0 here (see the codeql warnings also)

marquiz: Good point! See #1105

PR #1119, opened 2024-09-09
e2e: fix pod deletion in Ubuntu 24.04 by disabling runc apparmor

askervin:
Pods fail to terminate in Ubuntu 24.04 without disabling runc apparmor.

Changing containerd configuration did not work for me.

https://www.reddit.com/r/kubernetes/comments/1dv6z9d/ubuntu_2404_pod_termination_issue/

This does not break Ubuntu 22.04, that works both with and without this change.

### DIFF
--- a/demo/lib/distro.bash
+++ b/demo/lib/distro.bash
@@ -176,6 +176,21 @@ debian-ssh-user() {
     echo debian
 }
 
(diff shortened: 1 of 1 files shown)


klihub (approved):

marquiz (approved):
LGTM

PR #1132, opened 2025-03-13
Bump golang to v1.24

marquiz:
Also contains collaterals for making this wreck buildable again:

1. Bump the upload-artifact github action to v4. Need to update because v3 has been deprecated (and made unavailable) and causes the CI to fail.
2. Update some deps to make the trivy scan pass muster
3. Avoid "non-constant format string" errors in tests
4. Bump golangci-lint to v1.64.7
5. Disable revive checker from golangci-lint


marquiz:
ping @klihub @askervin the CI now passes muster

klihub (approved):

